PR #618, opened 2015-09-16
Update travis badge from umple-ucosp to just umple

aforward:
This will get the travis badge to work against the correct umple repo (renamed from umple-ucosp during our merge from google code)


Nava2:
Simple typo fix. Merging.

PR #632, opened 2015-10-11
Initial guidelines on contributing to umple

aforward:
Here are some guidelines for how best to contribute to the Umple project.  Much of the details and process align with another project that I follow called Elixir (their contribute page is here https://github.com/elixir-lang/elixir/blob/master/CONTRIBUTING.md)

This should help those that are unfamiliar with git to properly setup an environment where pull requests would be easy to do.


TimLethbridge:
This would result in us increasing the amount of documentation in the wiki + the regular read.md (both of which duplicate bits of this)

Personally, I would prefer to shorten this a lot, and move or merge content with new or existing wiki pages, and simply have here pointers and an abbreviated summary.

Detailed comments:
- It says to not use the issue tracker for feature requests . I disagree, and furthermore it says later that they are welcome
- The section ' umple project is split into many smaller subproject ..." duplicates the Readme.txt and should simply point there.
- Personally, it is not clear to me _why_ one should fork rather than directly adding a branch. I can guess why, but this needs to be explained so people 'get it' and remember.
- It would be good to explain why 'Never ever_ merge upstream `master` into your branches'. Telling people not to do something without explaining the bad consequences might result in people not remembering/learning.

aforward:
For feature requests.... it points the developer to use the mailing list first.  I'll remove that.

CONTRIBUTE.md
```diff
@@ -0,0 +1,239 @@
...
+process easy and effective for everyone involved!
+
+## Using the issue tracker
+
+Use the issues tracker for:
+
+* [bug reports](#bug-reports)
+* [submitting pull requests](#pull-requests)
+
+Please **do not** use the issue tracker for personal support requests nor feature requests. Support requests should be sent to:
```
TimLethbridge: Line 13 should not say 'feature requests'. Indeed we want the issue tracker to be used for feature requests.

Nava2:
_Really_ tiny nitpick.. the "unwritten standard quo" is to use `CONTRIBUTING.md` not `CONTRIBUTE.md`. It doesn't really matter, figured I'd mention it, though.

Issue #411, opened 2015-08-26
Allow as an option the display of generated code in UmpleOnline in a series of tabs


Nava2:
Originally reported on Google Code with ID 513

---

Currently, all the generated files are joined together in the output. And you can't
even see the file names.

It would be better to allow, as an option, the ability to see the generated files in
separate tabs.

This should be optional because sometimes it is highly advantageous to be able to search
through all generated output looking for certain strings.

One way to achieve this would be to have the code that generates the single glommed
together file to insert a special line before each file containing the file name, for
example
//%% abcd.java

Then the option to display tabs could simply parse this and split it. Note that when
files are in subdirectories caused by namespace directives, this would have to be accounted
for.

An alternative would be just to look for all the generated files. The problem is that
currently generated files may be left lying around in certain cases from a previous
generate

---

Reported by @umple on 2014-02-13 18:27:40

Nava2:
See also issue 225

---

Reported by @umple on 2014-02-13 18:32:00

Nava2:
See also issue 225

---

Reported by @umple on 2014-09-05 15:06:33
- **Status changed**: `Accepted`

TimLethbridge:
This was fixed long ago

PR #636, opened 2015-10-15
Issue #411 - tabs for generated code in umpleonline

akereliuk:
The feature to output generated codes in separate code blocks for every file was requested. I have added a checkbox to UmpleOnline after the code is generated called "Show Files in Separate Tabs". Clicking this will bring up buttons with the corresponding file names, and clicking on one of these will show only the code that is found in that file. Unchecking this checkbox will revert to showing one code block with all the files glommed together.

Closes Issue #411

### DIFF
--- a/cruise.umple/src/Main_Code.ump
+++ b/cruise.umple/src/Main_Code.ump
@@ -533,8 +533,8 @@ class PlaygroundMain
         {
           answer += "\n\n\n\n";
         }
-	answer += "//%% NEW FILE " + generatedCode.getKey() + " BEGINS HERE %%\n\n";
-        answer += generatedCode.getValue();
+				answer += "//%% NEW FILE " + generatedCode.getKey() + " BEGINS HERE %%\n\n";
+				answer += generatedCode.getValue();
       }
     }
     else if ("-generate".equals(args[0]))
--- a/cruise.umple/src/Main_Code.ump
+++ b/cruise.umple/src/Main_Code.ump
@@ -292,7 +292,7 @@ class UmpleConsoleMain
     
     optparser.acceptsAll(Arrays.asList("version", "v"), "Print out the current Umple version number");
     optparser.acceptsAll(Arrays.asList("help"), "Display the help message");
-    optparser.acceptsAll(Arrays.asList("g", "generate"), "Specify the output language: Java,Cpp,RTCpp,SimpleCpp,Php,Ruby,SQL,Json,Ecore,TextUml,GvStateDiagram,StructureDiagram,GvClassDiagram,GvClassTraitDiagram,GvEntityRelationshipDiagram,Alloy,NuSMV,Yuml,SimpleMetrics,Uigu2").withRequiredArg().ofType(String.class);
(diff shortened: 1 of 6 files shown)


TimLethbridge:
A test is failing in your PR.

You need to run a full build and then check for the word FAIL in the
build. Don't rely on 'BUILD SUCCESSFUL'.

You should also look in the dist/qa/index.php you will see there that is
says that PlaygroundMainTest is failing.

The test in quuestion, on line 93 asserts that the generated output should

## start with

## /_PLEASE DO NOT EDIT THIS CODE_/

## In fact, your output now begins with

//%% NEW FILE One BEGINS HERE %%

## /_PLEASE DO NOT EDIT THIS CODE_/

So you need to modify the test

There are additional issues.
1. The 'show files in separate tabs' checkbox continues to appear when
   generating Java Api Doc, Ecore, and other outputs that don't require it.
2. When a tab is displayed, the first line should be the /*PLEASE DO NOT
   ... line or whatever would have come first before you added the feature.
3. If I show a tab, and then 'generate' again, the new stuff appears and
   the previously displayed tab appears below.
4. I also think that if the 'show as tabs' is selected it should stay
   selected when regenerating or generating some other output. In other
   words it needs to be remembered as an option.
5. I think it might be preferable also to automatically display the
   leftmost tab when switching to this mode.

I am glad to see this works nicely with Java, PhP, Ruby and C++ and indeed
in C++ it is properly showing the .h and .cpp suffixes. I note you took  
the decision to hide the suffixes in the other languages. I suppose that  
is acceptable, but I find it a little disconcerting to not see the 
filename appear as x.java for example in the non-tabbed form. I think we  
leave that as it is for now.

TimLethbridge:
To clarify the above points:

Point 0: The test should pass. Your new combined output now has the extra lines that say  //%% NEW FILE One BEGINS HERE %%
The test needs to expect these.

When displaying _combined_ output without tabs, the whole file appears, which would have 
//%% NEW FILE xxx BEGINS HERE
/_PLEASE DO NOT EDIT ...
body of code for file xxx
//%% NEW FILE yay BEGINS HERE
/_PLEASE DO NOT EDIT ...
body of code for file YAY
etc.

Point 2: When displaying tabs, the //%% NEW FILE lines should be hidden. What appears in each would tab starts with the /*PLEASE DO NOT EDIT lines. The contents of each tab looks identical to what would appear if the zip file was unzipped, so the line numbers match.

akereliuk:
Before I re-make the Pull Request, I just want to make sure of something.
For Point 0, I modified the test such that it looks for checks if the
source code begins with "//%% NEW FILE". Do you also want me to insert the
filename as a wildcard after NEW FILE, and looking for the "/\* PLEASE DO
NOT EDIT" lines in this test as well?

TimLethbridge:
This is much better, but there are still some bugs.
1. The 'download zip' link is wrong when tabs are displayed.

If I do the normal 'generate code' in Java I see a download zip line that allows me to download the zip correctly.

But if I then click 'Show Files as Tabs' the link is then incorrect (it has the wrong tmpnnnnn directory number
1. SQL code doesn't split into tabs as it should 
2. When splitting into tabs, there is a blank line 1 at the start of each file that needs to be removed.
   So in the following, the 001 line should be the one that has the comment

001.
002. /_PLEASE DO NOT EDIT THIS CODE_/

PR #637, opened 2015-10-18
Abstract class

mviz:
For issue: 529.

Added italics and guillemets to indicate abstract classes and methods to Editable Graph and GraphViz diagrams on umple online.

Updated all JSON tests to include the updates to the JSON (to accommodate an isAbstract attribute) and added a new test.


TimLethbridge:
Excellent job. It works nicely. I will merge it. Then test in Umpleonline/test before putting it into production. Don't worry if cruiseControl initially fails to build; we are having some problems with it.

PR #642, opened 2015-11-05
Issue #594 - Using immutable key generates invalid Java code

### LINKED ISSUE #594: Using immutable key generates invalid Java code
Originally reported on Google Code with ID 696

---

1. http://cruise.eecs.uottawa.ca/umpleonline/umple.php?model=150401739102
2. Generate Java code, view hashcode() method, "canSetK" is used however does not exist

**What is the expected output? What do you see instead?**

"canSetK" should not be used.

---

Reported by @Nava2 on 2015-04-01 21:24:13

toulayha:
Closes #594 
The variable canSetK is not generated anymore for immutable keys, in the java code under the hashCode() method. The JavaClass generator now checks if the key is immutable before appending the "canSetKey" string.


toulayha:
No files were changed.

PR #648, opened 2015-11-15
Added test for immutable key #594

toulayha:
- Added actual test for Immutable key #594 
- Added test case ImmutableKey.ump

PR #669, opened 2015-12-10
Added test for immutable key #594

Nava2:
- Added actual test for Immutable key #594
- Added test case ImmutableKey.ump

See #648, sorry @toulayha, I tried to update the branch to kick off builds and accidentally deleted it. It's back here as it was. :+1:


TimLethbridge:
This was superceded

PR #678, opened 2016-01-15
Tab Display

vtrlx:
This is my fix for the issue of tabs not displaying in UmpleOnline when viewing individual files. The problem was that each class file was added to a &lt;div&gt; on the page only when the next class was found, meaning the last class would be dropped as there is no next class after the final one.

The fix involves just re-ordering the logic so that each line in the full file is added to the &lt;div&gt; as it's processed, instead of adding the whole class once the next one is found.

### DIFF
--- a/umpleonline/scripts/umple_action.js
+++ b/umpleonline/scripts/umple_action.js
@@ -2132,30 +2132,29 @@ Action.generateTabsCode = function()
   jQuery('.content').each(function(){
     // If New File Beginning
     if(jQuery(this).text().indexOf("//%%") >= 0){
+      intFileCounter++;
+      if(intFileCounter > 1){ arrCodeFiles[intFileCounter] += "</p>"; }
       strFileName = jQuery(this).text().slice(14);
       strFileName = strFileName.substr(0, strFileName.indexOf(' '));
       arrFileNames[intFileCounter] = strFileName;
-      arrCodeFiles[intFileCounter] = strFileContents;
-      intFileCounter++;
       jQuery('#generatedCodeRow').append("<div id='innerGeneratedCodeRow" + intFileCounter + "'></div>");
-      strFileContents = "<p>URL_SPLIT";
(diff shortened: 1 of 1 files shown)


TimLethbridge:
Can you please confirm that you have generated every possible type of generated output in cases where there is a) Just 1 class, and b) two or more classes. There were lots of little cases before; I would prefer to get your confirmation that you have done all this testing before doing a test, or having someone else do it. The student who worked on this before tended not do test all the output types (there are many). Also some output types don't generated on a 'per class' basis at all, so there have been problems with those. And some output types generate an extra class or two to support the compilation (I think cpp does that in some cases).

vtrlx:
Alrighty. I've thoughroughly tested the results both in my local UmpleOnline (with the fix) and [UmpleOnline on Cruise](http://cruise.eecs.uottawa.ca/umpleonline/).

Cases where the user can tabulate files are when generating code for the following languages:
- Java
- PHP
- C++
- Ruby
- SQL

For all but C++, one file is generated for each class. On Cruise, the generated code consistently returns no tabs at all when generating a model with a single class. On my local UmpleOnline, the single tab pops up as it should. For models that generate multiple files, Cruise fails to display a tab for the last file generated, but the rest display fine and work as expected.

C++ is a bit of a different case because it generates it's own files on top of the .cpp and .h files for each class in the model. In both cases for single classes and multiple classes, the tab for the last file is not displayed in Cruise's UmpleOnline. This case works normally for both cases on my local instance with the fix.

I've tested to see differences for the other generation targets even though they didn't have any tab display enabled, and found no difference in both single- and multiple-class tests.

Looks good!

TimLethbridge:
I merged it. There is a separate process to test it on the UmpleOnline server at http://try.umple.org/test which I will do after a build runs there, so be patient for it to appear live. Good job!

vahdat-ab:
Can we close this issue Tim?
It's good to close it if it's been fixed :-)

TimLethbridge:
This is a PR. The issue is 650 and it is closed!

Issue #548, opened 2015-08-26
Trying to disable text editor with shortcut opens a new tab


Nava2:
Originally reported on Google Code with ID 650

---

**What steps will reproduce the problem?**
1. Write any umple code so we can see a diagram
2. Press ctrl+t to hide the text editor

**What is the expected output? What do you see instead?**
Expected: text editor hides.
Actually: a new tab opens.

**What version of the product are you using? On what operating system?**
Latest UmpleOnline, Ubuntu 14.04.

**Please provide any additional information below.**
I am using Chrome Version 38.0.2125.111 (64-bit). Ctrl+t is a Chrome short-cut for
opening a new tab, which interferes with Umple's shortcut.

---

Reported by `gabriel.tapuc` on 2014-11-07 17:17:38

TimLethbridge:
Control-t works fine in Mac OS (on Chrome and other browsers. Keep that. 

Detect OS and make it something else that doesn't conflict on Linux and Windows, (e.g. something else with the 't'). Check the other shortcuts too.

Diusrex:
I can take this issue.

vahdat-ab:
great.

TimLethbridge:
Make sure you leave the key bindings on Mac OS in tact. The key is to use key binding combinations that work on other platforms (e.g. control-option-t) might be a good choice in Windows and Linux instead of (control-t). The latter would still work on Mac. This needs testing in multiple browsers on multiple OSes. And all the key bindings need changing. key bindings are set in two separate places ... one for the text pane and one for the graph pane.

Diusrex:
The two key bindings that failed are control-t and control-n. It looks like this issue isn't because of Windows/Linux since it works in Firefox for both Windows and Linux. Instead, it fails in Google Chrome because the two key bindings are reserved.

For control-n, I was able to get it working as control-alt (option)-t in both Windows and Linux. However, while I was able to get ctrl-alt-t working in Windows, in Linux (Ubuntu) it will open a terminal rather than be captured by the javascript.

Nava2:
I feel like working with `ctrl + t` or `cmd + t` is just a bad idea. This shortcut is commonly used in many applications. What about using `cmd + h` or `ctrl + h`. It'd take some research to confirm, but I don't think those key shortcuts are used. It'd take some investigation, but I feel like using `t` is just too common and will run into problems. 

[...]

PR #681, opened 2016-01-15
UmpleOnline shortcuts fixes

Diusrex:
Fix for issue #548. Choose to allow both the alternative and original shortcut because they are more browser dependent than OS dependent.

Tested this on Windows, Ubuntu, and Mac.


TimLethbridge:
Great. I will test this and merge it shortly if it seems to be OK

Issue #697, opened 2016-01-17
Need to sync UmpleToJava template with generated code and sync with UmpleToTemplate


Diusrex:
In commit [328c2664d9a158092772422597a53cff5d24be98](https://github.com/umple/umple/commit/328c2664d9a158092772422597a53cff5d24be98), from November 25, there were many changes to UmpleToJava/templates/association_Get_All.jet and UmpleToJava/templates/association_Set_All.jet. These changes essentially simplified the selection of association get templates, so that only the base variable name will be used.

However, there have not been any changes to the JET generated files UmpleToJava/src/cruise/umple/compiler/java/JavaClassGenerator.java or UmpleToJava/src/cruise/umple/compiler/java/JavaSpecGenerator.java since these changes were applied.

This means that when the JET files are regenerated (which will ONLY happen in Eclipse when the UmpleToJava project is included), there are a large number of changes to the JavaClassGenerator.java required to update it to the current template.

These changes cause the full build to fail when running testbed/src/TestHarness.ump, because when the newly updated JavaClassGenerator is used, there are a lot of errors in umple/testbed/src-gen-umple/cruise/associations/specializations/{Az, Alteration, Leg, Spatula, Wheel}.java, as well as other files, due to getters and setters being assigned the same name. They get the same name due to the previously mentioned simplification of names

TimLethbridge:
Hi Morgan,

It would seem to me that these changes to the two Jet files in question apparently were never applied in the compiler (because they couldn’t have been if there was no generated code saved)

I suggest obtaining the versions as they were just before this, and seeing if that solves the problem.

Then, and only then, see if the changes to the jet files made make any sense. A quick glance suggests to me that this was an accidental save of experimental work, since it makes no sense to so radically simplify. The generator templates were had various forms for good reason.

> On Jan 17, 2016, at 14:26, Morgan Redshaw notifications@github.com wrote:
> 
> In commit 328c2664d9a158092772422597a53cff5d24be98 https://github.com/umple/umple/commit/328c2664d9a158092772422597a53cff5d24be98, from November 25, there were many changes to UmpleToJava/templates/association_Get_All.jet and UmpleToJava/templates/association_Set_All.jet. These changes essentially simplified the selection of association get templates, so that only the base variable name will be used.
> 
[...]

PR #702, opened 2016-01-25
Sync Java JET Template with generated code

Diusrex:
The JET Templates should now generate code that is exactly the same as what was already existing in the repository.

Relates to issue #697, although not a complete solution.

Until that issue is fixed, will not be able to run the codegen.xml since will overwrite updated Java JET templates


TimLethbridge:
I have verified that the students in the fall all worked on issues unrelated to Java generation, such as analysis, website, php generation. So I am pretty sure we are safe here now.

Issue #244, opened 2015-08-26
Attempt at multiple inheritance not detected as an error


Nava2:
Originally reported on Google Code with ID 345
Owned by @ope

---

**What steps will reproduce the problem?**

class Z{}
class B{}  
class C{
 isA Z;
 isA B;
}

Will result in C being a subclass of Z. No error is raised about B

---

Reported by @umple on 2012-10-25 14:59:58

Nava2:
Reported by @umple on 2013-05-01 01:49:13
- **Owner changed**: @ope

Nava2:
This issue has been resolved. The attached is the patch for fixing it.

---

Reported by @ope on 2013-06-25 15:06:29
- **cc added:** +@umple
  - _Attachment: [patch-Adesina-345-Attempt at multiple class inheritance.diff](https://storage.googleapis.com/google-code-attachments/umple/issue-345/comment-2/patch-Adesina-345-Attempt at multiple class inheritance.diff)_

Nava2:
Reported by @ope on 2013-06-25 15:23:17
- **Status changed**: `Done`

Issue #242, opened 2015-08-26
Eclipse highlighting of words with u followed by s is wrong


Nava2:
Originally reported on Google Code with ID 343

---

**What steps will reproduce the problem?**

In Eclipse, any word
that has the sequential characters 'u' and 's' has that highlighted up
to the nearest occurrence of the character 'e' or a newline character. 
This means that the code:
class Customer{} has "ustome" highlighted as though it were the keyword
"use".

This was reported by 
Guylain Greer guylaingreer@gmail.com

---

Reported by @umple on 2012-10-25 14:09:45

PR #706, opened 2016-01-28
Add two more cache directories to `.travis.yml`

Nava2:
I added the two following directories:
- `$HOME/.ivy2/` - This is where Ivy downloads it's dependencies to
- `$HOME/.m2/` - This is where Maven downloads, while I don't think it's used, it's not worse for wear to have it

I also removed the `-Donline=true` marker from the `first-build` command. I would rather see this fail to detect than overriding it.

Additionally, I cleaned up the output on OSX by removing the `brew update` output (it's not useful to begin with). 

You can see the improved output from caching by comparing the `first-build` steps from: [Build #242.1](https://travis-ci.org/umple/umple/jobs/105486000) to [Build #244.1](https://travis-ci.org/umple/umple/jobs/105488194)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ language: java
 cache:
   directories:
     - $HOME/.ant/lib/
+    - $HOME/.ivy2/
+    - $HOME/.m2/
     - $TRAVIS_BUILD_DIR/dist/libs
 
 addons:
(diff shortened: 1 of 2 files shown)


TimLethbridge:
I am happy to merge this ... I just would like a few lines of explanation of the benefits

Nava2:
Sure! I added the directories that Ivy downloads to. This means we won't be
downloading a lot of the dependencies unless it is necessary. This should
cut down around 45s per build (higher for slower connections).

On OSX, we were installing a useless package, so I removed it.

This only applies to Travis, though.

PR #721, opened 2016-02-09
Reduce memory requirements for building Umple on Travis Linux closes #720

### LINKED ISSUE #720: Umple, Travis, and the Memory problem in general
Currently, when we build on Travis-CI's Linux platform, it fails. See [#257](https://travis-ci.org/umple/umple/builds/107889342), [#255](https://travis-ci.org/umple/umple/builds/107675157), etc.. 

The reason why it fails (talked about [here](https://github.com/travis-ci/travis-ci/issues/5582)) is due to the memory Umple consumes is too large. Memory problems will arise on **any** memory constrained machine, not just Travis. The solution of "add more heap space" when we already consume 1.2+GB of RAM is not reasonable for a compilation process.

I'm not an expert by any means on memory profiling, so I would like any and all input about possible improvement points and reductions: 
- [ ] Investigate changing the GC cycle to run more often (this will likely slow down compilation)
- [ ] Use string caching instead of the "load at class time" scenario we currently have
  - Load from files stored as Java resources (packaged into the jar) instead of static strings
  - Use [Guava's Cache](https://code.google.com/p/guava-libraries/wiki/CachesExplained) or hand-roll one for our needs when loading/storing strings (with Ivy we can pull this in simply)
- [ ] Improve the memory footprint of Umple's template language 
  - This is a big consideration given #708 going forward
  - Also an opportunity!
- [ ] Cache models during compilation: 
  - Once a model is used, if we can drop any references from memory, it would be beneficial
    - I think the model verification stage needs all the memory at once, but we might be able to hack some parts together 
  - Cause "model rendering" to be incremental
  - Use a flat-file cache to write object meta-model data to then pull from it as required and recreate the objects when generating code and are needed keeping "weak" objects in memory that hold the graph of dependencies

These are my initial thoughts, I would really like input and help in implementing this. I don't think I have time to tackle this problem myself, but I'll make some time going forward to help out.

TimLethbridge:
On Travis Umple's resetUmpleSelf will now be limited to 500MB this causes a little more garbage collection (1% total time) when compiling Master.ump, but prevents travis from failing out on Linux. When doing a normal full build the umpleSelf ant task is used instead, which doesn't have this limit. 

Also the limit is applied to the documenter when building the manual and to template.test. These were failing out on travis on Linux.

### DIFF
--- a/build/_template.xml
+++ b/build/_template.xml
@@ -182,6 +182,7 @@
     <deps-get-path conf="test" pathid="test.ivy.classpath" />
 
     <junit fork="yes" forkmode="perBatch" haltonfailure="${haltonfailure}" showoutput="${showJunitOutput}" printsummary="${showJunitSummary}" >
+      <jvmarg value="-Xmx500m"/>    
       <!-- <jmvarg value="-XstartOnFirstThread"/> -->
       <formatter type="xml" unless="${showJunitOutput}"/>
       <formatter usefile="false" type="brief" if="${showJunitOutput}"/>
--- a/build/build.umple.xml
+++ b/build/build.umple.xml
@@ -72,12 +72,15 @@
 
     <java jar="${umple.stable.jar}" fork="true" failonerror="true">
       <arg value="cruise.umple/src/Master.ump"/>
+      <jvmarg value="-Xmx500m"/>
     </java>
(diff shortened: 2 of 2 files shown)


TimLethbridge:
This PR seems to have passed Travis satisfactorily.

PR #728, opened 2016-02-11
Removed coupling from Position class

bitwizeshift:
Removed coupling from the parser's `Position` class by moving some Umple-specific methods to a new utility module called `ParserUtil`. Also updated all references to the original methods to instead call the new methods.


vahdat-ab:
I'm reviewing the code. I'll commit it soon.

vahdat-ab:
First, I don't know why your branch is failing in my machine. It passed in two servers. Would you mind double check in your machine again?

Second, we could get rid of one of the methods "getRelativePath" by using UmpleClassifier as a superclass for both UmpleTrait and UmpleClass. However, there are other modifications needed.
forget about this because we will deal with this later.

aforward:
I would prefer to see this a mixin to support the Umple language design.  So you have a base Parser class, which could be refactored out, and then we mixin the desired functions for umple's needs.

I don't like going against OO, taking `p.xxx(a, b)` and turning it into `PUtil.xxx(p, a, b)`. We could create a new parser.ump file that has the mixed in additional methods.

TimLethbridge:
I am inclined to agree with Andrew. Best practice is to avoid static methods if at all possible. 

I think using a mixin file is indeed the way to go for special purpose methods.

bitwizeshift:
@vahdat-ab 
Not sure why it wasn't working for you, I'm not having any problems. Did it perhaps cache some of the generated .class files from a previous build?

> Second, we could get rid of one of the methods "getRelativePath" by using UmpleClassifier as a superclass for both UmpleTrait and UmpleClass

I was originally looking into that with `UmpleElement` instead of `UmpleClassifier` -- but either would work. The only problem with this is neither of them have the `getSourceModel()` method. I don't really know whether or not all `UmpleClassifier`s _need_ a source model, since [UmpleInterface](http://cruise.eecs.uottawa.ca/umple/umple-compiler-javadoc/cruise/umple/compiler/UmpleInterface.html) doesn't currently have that method.

@aforward & @TimLethbridge 

My primary motivation for making the methods static is that the methods -- in my mind -- don't really fit properly in either the `UmpleClass`,`UmpleTrait` or the `Position`. In my opinion, It would be strange for the `Position` to have to have knowledge about an `UmpleClass` or `UmpleTrait`; the parser should just deal with parsing data. Similarly it seems strange for `UmpleClass` or `UmpleTrait` to have to know about `Position` elements. That's why I figured an aggregate utility method would remove that issue.

If doing this as a mixin, do you suggest that the capability be mixed into the `UmpleClass`/`UmpleTrait` instead, basically switching the call from `<position>.getRelativePath( UmpleModel model, String language )` to `<class or trait>.getRelativePath( Position position, String language )`?

vahdat-ab:
I tried to move sourceModel to UmpleClassifier and then I just changed two places in the code in which I had to give a proper argument for UmpleInterface constructor. You can do it. Afterwards, you face with test cases. You can define a method like below and get rid of those errors.

```
public UmpleInterface(String name){
super(name,null);
}
```

I left it to you so you may even move more properties to UmpleClassifier (We can do it later because we don't want to mix stuff up in the level).
In my point of view, all subclasses of Umpleclassifier needs to have access to the model because in some points the need to process other elements of the model.

@aforward & @TimLethbridge 
Regarding getRelativePath, I agree with Matthew because the class Position is going to be a general element and it shouldn't be polluted with other classes. We can implement this in UmpleClassifier or other superclasses.

@bitwizeshift 
If we cannot find a common superclass for other elements we can put the method inside a trait and reuse it in different classes.

bitwizeshift:
@vahdat-ab 
I will start refactoring the sourceModel into the superclass, and look into moving the `getRelativePath` method into there as well.

I'll issue a new pull request once the changes have been made and tested. In the meantime, I'll leave this one open in case anyone wishes to offer further input.

vahdat-ab:
@bitwizeshift The failure in my machine is not related to your code. Yours looks like passing. Recently, a modification in build process caused test failures in Windows machines. 
I wonder if you build in Windows and see those failures? (Update the repository to the latest data)?

bitwizeshift:
@vahdat-ab 
I'm no longer working on/building Umple on Windows due to the really long build-times it produces. Instead I have everything set up with a remote connection to my Linux server (running Ubuntu 14.04.3 LTS) that I use for my work, and I just remote-desktop/ssh/map network drives to it when I work on it.

With regards to refactoring: I've modified `UmpleClassifier` to now contain the source model, and updated the `getRelativePath` method to take the superclass instead of the subclasses. This leaves me with one question: What should the ultimate decision be with where we throw the `getRelativePath` method?

I understand the reasoning behind trying to avoid public static methods, but at the same point I can't quite understand why the method would exist in any subclasses of the `UmpleClassifier`.

The `UmpleClass` and `UmpleTrait` represent the concept of a class and trait respectively in Umple, and thus it doesn't feel proper for it to require knowledge about its position in the file that was parsed by the parser. 
Similarly the whole reason for this refactor was to remove the dependency between the `Position` and the `UmpleClass`/`UmpleTrait`, because it doesn't make any sense for something spit out by the parser to rely on them.

vahdat-ab:
@aforward @TimLethbridge 

I support the idea that not to have static methods, but the solution defined by mixins looks not be more modular. If we define `getRelativePath` as a mixin, finally it will be part of the class parser and again a dependency on Umple analyzer (Umple project).
I think the Umple analyzer should have the parser as an input and then define it's own APIs on top of it.

I also don't think that `getRelativePath` should be part of `UmpleClass` and `UmpleTrait` because it's completely different concern and it is not those classes' responsibilities. Furthermore, `getRelativePath` can have a general application (for example, for other Umple elements) so it's better to be separated. 

I would like to get your final thoughts because Matthew is waiting for it so as to continue his work.

aforward:
My concern is the code smell of moving p.doX(a,b) into PUtil.doX(p, a, b), which in my view goes absolutely against OOP.

The bigger concern about whether doX should be there in the first place is separate and we shouldn't introduce one code to remove another.

aforward:
One code smell for another --- sorry on my phone putting kids to bed

vahdat-ab:
I do agree that using static methods is the source of code smell (mostly for testing purposes). However, it's not always the case. There are some cases in which even using static methods may be useful. For example,  using static methods as a utility method or using static methods in place of constructors.

Generally, the proper approach is to figure out the method `getRelativePath` belongs exactly to which classes so we can assign it to them. If we assign it to the class parser (generally to the parser project) we cannot consider that one as an independent module because we will force any future users to have those related classes (for example, UmpleTrait and UmpleClass). On the other hand, it is not also related to UmpleTrait and UmpleClass responsibilities. 
This is a functionality that we would like to have in Umple and is unique for us.

@TimLethbridge @aforward 
I wonder what the best approach is for these cases?
Would you mind clearly specifying what the proper solution is regarding `getRelativePath`?

bitwizeshift:
I pretty much agree entirely with what Vahdat just said.

The reason I suggested a static method is so that it would be treated as an extension to the existing API without needing to override or overload anything -- while at the same time minimizing coupling. Though this creates something that is essentially the same as 'free-functions' in non-OO programming terms, it acts functionality like an extension to the class -- adding a feature not present in the original API.

Mixins would only complicate the descendent classes by re-introducing the coupling into them, which is why I was trying to shy away from that.

The only other alternative I could suggest is to produce a wrapper-class, say `UmplePath` that takes an `UmpleClassifier`, `Position`, and a language string in its construction, and contains a method `getRelativePath()`. This would make the syntax become something like `new UmplePath( classifier, position, "Java" ).getRelativePath()`. It would essentially be a dummy-wrapper around the two classes that provides the functionality in an OO way. This, to me, feels more complicated and verbose than necessary, but is the only true OO approach I can think of to solve this issue

aforward:
Sorry, I disagree.

First note that my comment about the code smell was about a specific use of static methods, not all static methods have their place.

Second, you haven't reduced coupling.  You have at function that needs three things to work in both scenarios, whether it's in the util class or not, the same number for things are still interacting with each other.

Finally, the mixin capability allowe you to extend an existing API without the need to override, overload, OR created helper classes with the functions you desire.

Why not move the function into UmpleClass? Swapping the args, so uClass.getRelativePath(p, "Java").  And this could be done in a separate .ump file with a mixin:

aforward:
Darn autocorrect on my phone, and the inability to edit a post... Let me know if you can follow the above despite the atrocious grammar.

vahdat-ab:
I somehow understood from your earlier post that you're talking about mixing on the Parser, which in that case I don't think we remove the coupling (I think I misunderstood).

If you're talking about a mixin on the class `UmpleClass` in terms of `getRelativePath`, this is what indicated in my first post by putting `getRelativePath` in UmpleClassifier to avoid duplication. This is much logical and practical because `getRelativePath` is going to be used by UmpleTrait as well.
If you're more on a mixin for UmpleClass (or UmpleClassifier), so Matthew can start and implement it.

Note: Mixin we have in Umple is not following the general concept of Mixins in CS. We use mixins to just gather elements of a class defined in several locations. Therefore, it exactly means adding `getRelativePath` to a class.

bitwizeshift:
Alright, thanks everyone -- I will get right on the new changes.

aforward:
@vahdat-ab ---> I think the Umple `mixin` is pretty close to the CS one, no?

From, https://en.wikipedia.org/wiki/Mixin

```
In object-oriented programming languages, a mixin is a class that contains methods for use by other classes without having to be the parent class of those other classes. How those other classes gain access to the mixin's methods depends on the language. Mixins are sometimes described as being "included" rather than "inherited".
```

Here the language `mechanism` is to copy and paste into the object code the functions, but the semantic of making other `functions` available directly within your class.  The feature that isn't available is using those functions in more than one class

```
use Person_extension.ump

class Person
{

}
```

vahdat-ab:
What we have in Umple is not what is called Mixin in CS. For example,
class A{
void test1(){}
}
class A{
void test2(){}
}

In UmpleModel we gather them as one class like this:
class A{
void test1(){}
void test2(){}
}

There is no way to use test2() in another class because it's part of class
A. If it was part of a Mixin, you would be able to use easily in other
classes. Reusable is the main advantage of mixins (even though there are
issues with Mixins as well).

In Umple we have traits that are more advanced than Mixins (CS definition).
For example,
class A{
isA T;
void test1(){}
}
trait T{
void test2(){}
}

The result is

class A{
void test1(){}
void test2(){}
}

While you can also reuse test2() in another class, for example
class B{
isA T;
}

bitwizeshift:
I've changed everything over to be inherited from an Umple trait, but this has introduced a new problem. 

The original implementation of `getRelativePath` could accept null instead of an `UmpleClass` or `UmpleTrait`  since they were passed as parameters. This was by design because state machines will often return null when `sm.getUmpleClass()` is called. 

This is not able to easily be handled with the new changes, since now in the event of a null, a call to `getRelativePath(...)` will throw null pointer exceptions. At first I thought it could be as easy as wrapping a small "`if( sm.getUmpleClass() != null )`" around the calls -- but it turns out there are almost 10 or so calls to this method, which makes this an inelegant solution.

I'm not trying to fixate on a solution path, but short of a refactor of state machines I can't think of a way to solve this that isn't either static, a wrapper class (as my [`UmplePath` suggestion](#issuecomment-185006577)), or using some method from a third-party class. 

Does anyone have any suggestions for this that is considered good OO design? Should I just do a bunch of null-checks as mentioned above?

vahdat-ab:
I wonder why do you use traits? What's the issue with just putting it in `UmpleClassifier`? 
If you try to have a reuse method so as to some other elements from different hierarchy level in the future can reuse it, it makes sense to have a trait. Otherwise, it's better to have in UmpleClassifier.
Regarding the `null` value, I think all state machines must return a valid class name. If this happens maybe it's because of an attempt from Omar in which he wanted to have reusable state machines. 
I wonder if you can let me have you modification?

Nava2:
If you're worried about speed, a conditional check will be optimized away
at runtime if it always evaluates to false/true. Particularly if it's known
that getUmpleClass() always returns null. Especially when the check has no
side effects.

I agree with you that its inelegant in that you're checking for it, but
it's really not a huge concern. There's something to be said about asking
why the statemachine code is the way it is, however that screams "out of
scope" to me.

bitwizeshift:
I forgot to mention that it also crashes JUnit tests due to the NullPointerException

@vahdat-ab 
I put it as a trait because I was told I should be using mixins, and you were just mentioning how traits are functionally similar to the definition of a mixin. Currently I have `UmpleClassifier` inherit from the trait, which means it is still equivalent to being in `UmpleClassifier` -- and I can move the code directly into there if it is preferable. At any rate, this isn't the cause of the problem. 

My modification is equivalent to just moving `getRelativePath( ... )` to `UmpleClassifier`, and changing all calls from `ParserUtil.getRelativePath( <classifier>, <position>, <lang string> )` to `<classifier>.getRelativePath( <position>, <lang string> )`

@Nava2 
I agree with the out-of-scope comment. I'm not sure _why_ it returns null, I just know that the method originally had the following piece of code:

```
if (parent == null)
{ //No parent class? This might happen with state machines
  return Paths.get(filename).getFileName().toString();
}
```

I had _hoped_ this was just written preemptively in case someone passed a null. It turns out that in practice this is now causing a NullPointerException during the compileJava step of the ant build, since `parent` is now the instance of the class calling this method.

The only reason I was looking to avoid wrapping if statements, is that it will now mean that most calls will turn into:

```
String filename = p.getFilename();
String path = null;
if( sm.getUmpleClass() == null ){
  if( filename == null ){
    path = "";
  }else{
    path = Paths.get(filename).getFileName().toString();
  }
}else{
  path = sm.getUmpleClass().getRelativePath( ... );
}
```

Which is a particularly large amount of code to do for each necessary call to `sm.getUmpleClass()`. Some of this stuff can be refactored so that it's only done once per state machine token -- but it still puts more work on the caller of the function and may be more difficult to maintain.

aforward:
As for, "There is no way to use test2() in another class because it's part of classA", I would consider that a lacking feature in Umple, as opposed to a requirement for a Mixin, in the sense that Umple use of Mixin isn't really that far outside what's in wikipedia, especially with the explicit note that the language gets to decide on the implementation.

vahdat-ab:
In the wikipedia it says "a mixin is a class that contains methods for use by other classes". The definition says reuse without worries. Anyhow, I think it's getting out of the topic for this issue :-). We could discuss it in anothe issue. BTW, there is no such lacking feature in Umple anymore because we have now full functional traits.

bitwizeshift:
So I don't think anybody directly acknowledged my suggestion for a non-static class that calculates the relative path between two objects. This would solve the NullPointerException issue, and has reuse implications as well in that it could be used to find the relative path between two `UmpleFile`s, or perhaps other `UmpleClassifier`s, etc.

It would have a low amount of coupling between the offending classes in the same sense that any utility class would have, but at least it would contain an added level of indirection instead of having the method embedded in one of the two classes. In terms of coupling, it's not much worse than having the `UmpleClassifier` rely on the `Position`.

It would essentially be a utility class, but it would at least contain a useful implementation. It could calculate the path in the constructor, and could be reused in places where `getRelativePath` has many subsequent calls since we can store an instance of the class. 

Any thoughts?

vahdat-ab:
My overall thought:
Yesterday, I dove in the code and wanted to explore the idea of having `getRelativePath` in `UmpleClassifier`. I had to do lots of minor changes to make it  pass all test cases. It opened some issues we had in our code:
- `UmpleClass` and `UmpleTrait` both have a source model (`UmpleModel`) while `UmpleInterface` doesn't have. Logically, it should have because each interface belongs to a Umple model. That's the reason putting the `UmpleModel` in `UmpleClassifier` makes sense.
- There is a direct association between `UmpleClass / UmpleTrait` and `StateMachine`. In the current implementation, `UmpleClass / UmpleTrait` of state machines are not set at all. Therefore, it must be done. Exception: If we want to have general reuse state machines (an old idea). I think now with traits we may not need it. State machines can be defined in traits and are reused several times.
  -We need `getRelativePath` for other elements in the model so it's not a concern just for two or three elements in the system. In the project `UmpleToRTCpp` it's been cleared indicated that the solution is not general.

Now, I can somehow say that putting  `getRelativePath` in `UmpleClassifier` can work, **but** this is not the correct implementation at all. The functionality `getRelativePath` is completely a utility and we may add other languages and elements, therefore, it should be separated.
Another reason for separation is clear because this is more a code generation requirement and so there is no need to be part of `Position` or `UmpleClassifier` in my point of view.

It looks like we have more challenge in analysis phase regarding the architecture so I think it's better to talk about them in another issue (when Matthew is done with this part).

Andrew proposed using mixins to add the `getRelativePath` to `Position`. With the current implementation it means directly adding it to the 'Position'. Therefore, anybody who uses Parser need to know about UmpleTrait, UmpleClass etc. Furthermore, I pointed out that this is a functionality between two elements so better to be separated. 
We may think that we put the mixin on `Position` in Umpl project and compile it when we use it in our own project. This has more potential for future students to come and extend parser's classes through the same approach. Therefore, we won't be able to keep parser growing independently because nobody will see those minins when compiling the parser project separately. Another reason is that in the mixin we have to make hard code path for using 'use` statement.    

In my point of view, Matthew's first (with static) and second (wrapper) make more sense. If the first one is considered as a code smell (which in my opinion it's not) we can use the second one which is more OO.

Generally, I would say it's better to consent soon on one of the methods and let Matthew explore more issues :-)

aforward:
Isn't this fun.

To clarify my position, I don't want to move something to a static helper of the form `p.do(x)` to `PHelper.do(p, x)`.  Mixins was a possible way to keep the parser separate, yet putting the code back in for Umple's needs.

As this discussion has been dragging on for a while, it was time for me to do more than simply look at surface of the issue, and instead dive a bit deeper to look at both the implementation and the usage to see if there's a better place to do things at a more fundamental level

## Where is it being Used?

The feature is being used to augment the generated code to point back to the original code, here are they are..

```
p.getRelativePath((UmpleClass)(inject.getUmpleClassifier()), ((UmpleClass)inject.getUmpleClassifier()).getSourceModel().getDefaultGenerate() )
exception.getPosition().getRelativePath(aClass, "Java"));
exception.getPosition().getRelativePath(aInterface, "Java"));
token.getPosition().getRelativePath(aClass, "Java");
token.getPosition().getRelativePath(aTrait, "Java");
exception.getPosition().getRelativePath(aTrait, "Java"));
thePosition.getRelativePath((UmpleClass)null,"Java")
thePosition.getRelativePath((UmpleTrait)null,"Java")
```

So not that bad.  The major trouble spots are the last two where we don't have "anything" to hang off of (See below).  A end goal on changes here (that would need the internals changed could be as follows) could look like

```
inject.getUmpleClassifier().getRelativePath(((UmpleClass)inject.getUmpleClassifier()).getSourceModel().getDefaultGenerate())
aClass.getRelativePath("Java");
aInterface.getRelativePath("Java");
aClass.getRelativePath("Java");
aTrait.getRelativePath("Java");
```

Notice that the code is coming from the Umple side of things and Position is completely gone.

Not that for the last two cases (where the classifier was null), we need to look a bit more at the code (and failing tests as we make changes), because I don't see why the `null` is buying us, as we have `aClass` available

```
    // Add special position comment at the end if @outputumplesource had been 
    // detected earlier in a comment
    if(outputUmpleSource == true) {  
      aClass.addComment(new Comment("@umplesource " + thePosition.getRelativePath((UmpleClass)null,"Java")+" "+thePosition.getLineNumber()));
    }
```

So we could 

```
aClass.getRelativePath("Java");
```

## How we we using Path / UmpleClass  and language

At present, the method needs `Position`, `UmpleClass` and `language`.  To reduce the coupling, we need to drop something there, not just move the code around. 

Let's start by looking at the implementation, here are all the things we reference

```
  Position.filename // For things outside of a "Class" like a StateMachine
  parent.getPackageName() // needed for Java's namespace file hierarchy
  parent.getSourceModel() // not sure what this is, but we use it to get the UmpleFile()
                          // as well as get all the targets
                          // to see if we are getting the provided "language"
```

Two things have have popped out.
1. We are jumping all over the place to get a `filename`, which is sometimes from the `Position` and mostly from `UmpleFile`
2. We could refactor the code to locate a `Generated` language, as a separate PR to help clean up this function (as easier is easy to change)

I am thinking more analysis should be done to refactor the guts of the function in the following ways,

a. The first parameter should be fixed to never be null,  there should be an ultimate `parent` for ever element which would have the necessary interface / methods for `getPackageName()` and `getSourceModel()`

b. By fixing the first parameter, we shouldn't ever have to use the positions `filename` (which is currently all we are using it for)

c. By fixing the first parameter to never be null, we have IMO found the perfect place for said method.  We no longer need the position's `filename`, and we move the function into this Class / Interface (implementation) and now the signature becomes `getRelativePath(String language)`

For c. it might be as easy as dropping the `null` and seeing how things land.  My guess is that the code was written that way for a reason, but still we should have lots of failing tests to help guid the refactor

# Next Steps

Here is what I propose

## Consolidate getRelativePath

We should remove the code clone separate from "fixing" it out of Position.

At present we have two `getRelativePath` functions that are complete dups of each other.  So a first PR should consolidates them into one. So @vahdat-ab, let's push a PR that moves the one function into `UmpleClassifier`.  Small steps to make the code better.

## Drop the ability to support NULL for parent

We have two edge cases in the code where we are explicitly shoving in Null.  Let's see if we can do better.  Doing so, we refactor out a bunch of references to `'Position.filename`

## Move getRelativePath into UmpleClassifier (or similar)

Now that we have removed the null case, we can move the code into `UmpleClassifier` as follows

```
aClassifier.getRelativePath(Position p, String language)
```

## Completely drop reference to Position

Finally, the current position is only used for the filename, which should be more than available from `UmpleClassifier` (or similar).  Drop all references and then we end up with... tada

```
aClassifier.getRelativePath(String language)
```

vahdat-ab:
I think all steps have been done in the branch [fix_3](https://github.com/umple/Umple/tree/fix_3) when I was working on except the "Completely drop reference to Position".
"Consolidate getRelativePath"
it's been done.

"Drop the ability to support NULL for parent"
These two locations also were confusing to me so Matthew please see what exactly they do.
I commented them because there is no part in `getRelativePath` to deal with `null` input.

"Move getRelativePath into UmpleClassifier (or similar)"
It's been done and all test cases and code generator templated were fixed.

"Completely drop reference to Position"
need to be done.

Issue #710, opened 2016-01-29
Better support for Aspect-orientation in Umple


vahdat-ab:
Currently, we support aspect-oriented features in Umple mostly for methods and events generated automatically by Umple. We would like to extend this to user-defined methods and also support parameters for both user-defined and umple-generated methods/events.
Another extension is to have such a feature for states so we can add code before and after entry/do/exit parts of states.
Finally, we would like to explore feasiblity of "around" mechanism of AOP for Umple.

vahdat-ab:
In order to get some understanding about some implementation of AOP, you can checkout [AspectJ](https://eclipse.org/aspectj/) or check the [wikipedia ](https://en.wikipedia.org/wiki/Aspect-oriented_programming)about different AOP implementations

vahdat-ab:
@TimLethbridge 
Currently, we have the following syntax for AOP:

```
class X  
{  
  a;  
  whenWasASet;  

  after setA {  
     //code  
  }  
  after getA {  
     //code
  }    
} 
```

In which we don't specify the parameters for the joint points (for example, setA). This can make issues. If we have two methods with the same name and different parameters, this approach won't work. 
What do you thin about this?
The following makes more sense

```
class X  
{  
  a;  
  whenWasASet;  

  after setA(String) {  
     //code  
  }
  after getA() {  
     //code
  }    
}  
```

This will allow us more flexibility.
I think putting parentheses is also critical to make it obvious for developers and the compiler. In the future, we may go with other modeling elements so it's better to indicate that the join points are methods by expressing parentheses.

TimLethbridge:
Sure. But keep the original so as not to break existing code.

> On Feb 8, 2016, at 17:17, Vahdat Abdelzad notifications@github.com wrote:
> 
> @TimLethbridge https://github.com/TimLethbridge 
> Currently, we have the following syntax for AOP:
> 
> class X  
> {  
>   a;  
>   whenWasASet;  
> 
>   after setA {  
>      //code  
>   }  
>   after getA {  
>      //code
>   }  
> } 
> In which we don't specify the parameters for the joint points (for example, setA). This can make issues. If we have two methods with the same name and different parameters, this approach won't work. 
> What do you thin about this?
> The following makes more sense
> 
> class X  
> {  
>   a;  
>   whenWasASet;  
> 
>   after setA(String) {  
>      //code  
>   }
>   after getA() {  
>      //code
>   }  
> }  
> This will allow us more flexibility.
[...]

PR #731, opened 2016-02-16
implemented code injections for user-defined methods

Shikib:
This pull request contains the implementation of functionality that would allow users to perform before/after injections on custom defined methods.

For example if you have:

```
class A {
  void foo(int x) {
    if(x < 0) {
      return;
    }
    System.out.println("foo");
  }

  after foo {
     System.out.println("bar");
  }  
```

This is something that previously did not perform the injection but will after this pull request. Right now what I'm doing is:

For before code injections:
- Placing the injection at the start of the method

For after code injections:
- Placing the injection before every return statement
- Placing the injection at the end of the method body if there are no return statement
- Placing the injection at the end of the method body if there are return statements but the last line isn't a return (consider example above for this).

The majority of the code in this PR is to deal with the indentation of the injected code, the actual implementation part is fairly simple.


TimLethbridge:
Does it work properly when a return statement is on the same line as other stuff or is in a conditional without braces. This may be bad style, but it should work?
   if (a=5) return(b);

TimLethbridge:
Oh no!

You have edited generated code.

The JavaClassGenerator.java
is coming from Jet

Shikib:
I moved the implementation to the .jet file. 

I don't think my current implementation handles this code:

```
if (a==5) return b;
```

The problem isn't that the return is on the same line though, it can handle the following reasonable well (i.e., the formatting is a bit off, but it's valid code nonetheless).

```
if (a==5) { return b; }
```

The problem is that my implementation expects that if a return is in a code block, we can simply insert the lines of code directly above the return statement and all of the code will remain in that block. This however isn't necessarily true.

```
if (a == 5)
  return b;
```

If code (eg. a System.out.println(b)) was to be injected into there, my current implementation would generate:

```
if (a==5) 
  System.out.println(b);
  return b;
```

Which is incorrect because the return is outside of the if statement (at least in Java/cpp -- I think ruby/php don't have this issue). 

What would be an acceptable way of handling this? I can't think of any approach that would allow me to detect whether a code block has braces surrounding it (because if we knew it didn't have braces, we'd inject the braces for the code block along with our code injection), without relying on external factors like indentation (which doesn't seem like a great approach as Umple doesn't enforce any indentation rules).

TimLethbridge:
See my comments on the original issue regarding another option.

Shikib:
@TimLethbridge: I fixed the issues that were identified earlier.

At this point in time, the following *.ump file:

```
class A {

  void setK() {
    if(test) {
      return b;
    }

    if(test)
      return b;

    if(test) {
      // )
      return b;
    }

    if(test)
      // {
      return b;

    if(test)
      /*
           {
      */
      return b;

    if(test) {
      /*
           )
      */
      return b;
    }

    if(test)return b;

    if(test)     return b;

    if(test) { return b; }

    if(test) {    return b; }

  }

  before setK {
    // foo
  }

  after setK {
    // bar
  }
}
```

will generate the following .java file:

```
/*This code was generated using the UMPLE @UMPLE_VERSION@ modeling language!*/



// line 1 "helloworld.ump"
public class A
{

  //------------------------
  // MEMBER VARIABLES
  //------------------------

  //------------------------
  // CONSTRUCTOR
  //------------------------

  public A()
  {}

  //------------------------
  // INTERFACE
  //------------------------

  public void delete()
  {}

  // line 4 "helloworld.ump"
  public void setK(){
    // line 43 "helloworld.ump"
    // foo
    if(test) {
      // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
    // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
      // )
      // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
      // {
     // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
      /*
           {
      */
     // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
      /*
           )
      */
      // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
    // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
    // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
      // line 47 "helloworld.ump"      
      // bar
      return b;
    }

    if(test) {
      // line 47 "helloworld.ump"      
      // bar
      return b;
    }
    // line 47 "helloworld.ump"
    // bar
  }

}
```

Proper functionality of code + indentation is preserved.

Shikib:
I fixed all of the previously identified issues with the code injection into user defined methods. The following .ump file, which encompasses all the various cases that I predicted would cause issues, generates the .java file that I would expect it to:

Umple file:

```
class A {

  void setK() {
    if(test) {
      return b;
    }

    if(test)
      return b;

    if(test) {
      // )
      return b;
    }

    if(test)
      // {
      return b;

    if(test)
      /*
           {
      */
      return b;

    if(test) {
      /*
           )
      */
      return b;
    }

    if(test)return b;

    if(test)     return b;

    if(test) { return b; }

    if(test) {    return b; }

  }

  before setK {
    // foo
  }

  after setK {
    // bar
  }
}

```

Generated Java file:

```
/*This code was generated using the UMPLE @UMPLE_VERSION@ modeling language!*/



// line 1 "helloworld.ump"
public class A
{

  //------------------------
  // MEMBER VARIABLES
  //------------------------

  //------------------------
  // CONSTRUCTOR
  //------------------------

  public A()
  {}

  //------------------------
  // INTERFACE
  //------------------------

  public void delete()
  {}

  // line 4 "helloworld.ump"
  public void setK(){
    // line 43 "helloworld.ump"
    // foo
    if(test) {
      // line 47 "helloworld.ump"
      // bar
      return b;
    }

    if(test) {
      // line 47 "helloworld.ump"
      // bar
      return b;
    }

    if(test) {
      // )      
      // line 47 "helloworld.ump"
      // bar
      return b;
    }

    if(test) {
      // {      
      // line 47 "helloworld.ump"
      // bar
      return b;
    }
    if(test) {
      /*
           {
      */
      // line 47 "helloworld.ump"
      // bar
      return b;
    }

    if(test) {
      /*
           )
      */
      // line 47 "helloworld.ump"
      // bar
      return b;
    }

    if(test) {
      // line 47 "helloworld.ump"
      // bar
      return b;
    }

    if(test) {
      // line 47 "helloworld.ump"
      // bar
      return b;
    }

    if(test) {
      // line 47 "helloworld.ump"
      // bar
      return b;
    }

    if(test) {
      // line 47 "helloworld.ump"
      // bar
      return b;
    }
    // line 47 "helloworld.ump"
    // bar
  }

}
```

TimLethbridge:
Looks promising! But can you please add an actual automated test.

Not clear why appveyor build gives different results from Travis! Some strange generated java seems to be hanging around in Appveyor.

TimLethbridge:
I see the builds are passing! But I don't see any automated tests yet. These are needed

TimLethbridge:
Any progress on getting some tests into place for this @Shikib

Shikib:
I do have some tests written locally. The reason I haven't pushed them yet is because the code injection functionality isn't currently active. The reason for this, which I briefly mentioned in our last meeting, is that until now -- a number of Umple developers have used code injections with the assumption that the code injections wouldn't inject into custom defined methods.

For example, something that I've seen a few times throughout the codebase is the following scenario. Say the method `setFoo(aFoo)` is automatically generated. The developer would then go and define the following method:

```
void setFoo(String aFooName, int aFooInt) {
  ...
}

after setFoo {
  System.out.println(aFoo.toString());
}
```

Because I don't think that the code injection changes should break any existing code (i.e., I don't think it's a good idea to go an rename every instance of such a method in the codebase), I decided to go with requiring a different syntax for code injections into custom methods. 

So for example, if someone wanted to inject code into all methods named setFoo (including custom defined ones), they'd do something like:

```
after setFoo~ {
  a = test;
}
```

I think we should discuss what the syntax is (I don't necessarily think a tilde is the best option), but right now I'm just working on changing how the code injection tokens and code injections objects are created to allow this to work. This ties in heavily to the next thing I'm working on, code injections with parameters specified, so a lot of the work I'm doing for this will be transferable to that.

So no functionality will exist, until I add in the new syntax parsing for custom code injections. When I do that, I will modify the tests (to use the new syntax) and then update the pull request.

vahdat-ab:
chek the issue #710 for more comments.

TimLethbridge:
This looks excellent. Good work. @vahdat-ab do you have any comments? If you want I can merge this after Morgan's work goes through.

vahdat-ab:
@TimLethbridge  I'm checking the code. Inform you soon.

vahdat-ab:
It looks it is working, I need more explanaition though. However, it's safe to go head.

Issue #740, opened 2016-02-21
Changes to UmpleTL break some tests


TimLethbridge:
I am @ignoring tests in the following files to make the build pass

NuSMVTemplateTest.java
UmpleTLTest.java

TimLethbridge:
The core issue here is that UmpleTL behaviour changed slightly in response to issue #732. This caused some tests to fail, but that wasn't detected due to the build not bootstrapping properly of the self-built jar (which is being addressed in pr #741 ). @opeyemiAdesina please adjust your generation as needed and re-enable your tests, @ahmedvc, please fix your tests too and re-enable them.

PR #743 addresses this to the extent that the build should work (I hope)

PR #741, opened 2016-02-21
Fix `umpleself` to use the current jar

### LINKED ISSUE #739: `umpleSelf` does not use "active" jar
Currently, `build.umple.xml:umpleSelf` uses the stable jar, not the currently built one. This caused a PR to "pass" when in reality the tests failed.

Nava2:
I think an overzealous refactoring happened which caused `resetUmpleSelf` to be equivalent to `umpleSelf`.

I can't test directly, but I'll let the CI handle that.

The tests fail, but I'm unaware if this is caused by my change or the issues in #740. 

Closes #739.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,16 +24,16 @@ before_install:
 
 env:
   global:
-    - ANT_FIRST='ant -f build.umple.xml -Dmyenv=travis'
-    - ANT_BUILD='ant -f build.umple.xml -Dmyenv=travis -Dfirst.build=false'
-    - ANT_TESTBED='ant -f build.testbed.xml -Dmyenv=travis -Dfirst.build=false'
+    - ANT_FIRST='ant -f build.umple.xml'
+    - ANT_BUILD='ant -f build.umple.xml -Dfirst.build=false'
+    - ANT_TESTBED='ant -f build.testbed.xml -Dfirst.build=false'
 
 script:
   - php --version ; ruby -v
   - cd build/
-  - ant -Dmyenv=travis bootstrap
+  - ant bootstrap
   # We resolve all of the dependencies before, this way the command will be unlikely to time out
-  - ant -Dmyenv=travis deps-resolve-all
+  - ant deps-resolve-all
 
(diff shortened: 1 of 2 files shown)


TimLethbridge:
Thanks. OK. I almost have the build working again. I will get that done first.

TimLethbridge:
I am going to test this on my local environment. There was a problem before with 
 <property name="dist.umple.jar"       value="umple-${umple.version}.jar"
because directly after a 'release' this changes, and people can't build. Also people doing local testing need a filename that never changes That is why I set up a symbolic link to make umple.jar point to "umple-${umple.version}.jar". My very strong recommendation would be that the bootstrap always use simply "umple.jar". Not sure, however, if the symbolic linking I set up works in Windows.

TimLethbridge:
Just to be clear regarding my last comment ... I had added the following a while back at line 339 in packageMainJar in build.umple.xml
   <echo message="Making symbolic link so it can be run as umple.jar" />
    <symlink link="${dist.dir}/umple.jar" resource="${dist.dir}/${dist.umple.jar}" overwrite="true" failonerror="false" />
The documentation shows this does not work on Windows
https://ant.apache.org/manual/Tasks/symlink.html
It is very nice (almost essential) in unix for normal development.
In Windows we would have to detect a failure of the above and do a copy instead. We could do a copy in Unix but it seems that would waste space and possibly lead to errors..

Nava2:
The build actually reads from the "version.txt" file and reads the jars appropriately. That symlink does exist on Unix, but it's not necessary as the `${dist.dir}/${dist.umple.jar}` variable will always have the right path (there is also `${dist.dir}/${dist.umple.prev.jar}`). So, I see what you're saying, but I don't think it's an issue. 

I just rebased this branch against `master`, so the build should pass all checks now.

TimLethbridge:
This broke the cc build. Reason: It tries to build using the 'upcoming'  jar which doesn't exist yet. This is why I had tried to insist on using a jar called 'umple.jar'

PR #743, opened 2016-02-21
Fixing Ecore+Use, commenting out others

TimLethbridge:
This temporarily addresses issue #740 by fixing the generators for Use and Ecore

It does not fix the tests for UmpleTL itself and for NUsSMV. @ahmedvc and @opeyemiAdesina will have to separately fix them. The failing tests there are just commented out for now.


TimLethbridge:
I am going to merge this immediately since it is late at night and the build is already broken.

Issue #748, opened 2016-02-21
Fix cc build by having umpleSelf build using umple.jar not umple.version.jar


TimLethbridge:
cc doesn't build properly since it updates the current version number that it is about to build. Then it tries to build using it! We need to have umpleSelf build using a generically named umple.jar in the build directory as we always used to

TimLethbridge:
One solution would be to have a variable set in _cc.xml , _local.xml and so on that defines which selfbuild jar to use. For the most part it would be "${dist.dir}/${dist.umple.jar}", but for cc it would be "${umple.stable.jar}""

TimLethbridge:
I will try this first thing in the morning. No more work tonight.

TimLethbridge:
I am going to fix this now

TimLethbridge:
Fixed by #749

PR #749, opened 2016-02-22
Fixing umplSelf for cc

TimLethbridge:
Changing build.umple.org to make sure cc uses the correct jar. Also parallelizing.

### DIFF
--- a/build/build.umple.xml
+++ b/build/build.umple.xml
@@ -47,41 +47,55 @@
     <antcall target="template.init" />
   </target>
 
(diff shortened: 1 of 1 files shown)


TimLethbridge:
Fixes #748 (hopefully). Notice that this also does the parallel part of #742 because I was modifying the same lines, but it does not do the other stuff in #742, so #742 will need to be modified to do the other changes

Nava2:
I'm very confused as to why this happens. 

In the build logs, it still states that the symlink is being made and I don't think there are any conditionals around CC in that process. Does CC support symlinking? I'm unaware what OS it's running.

PR #764, opened 2016-02-29
Add CORS support to UmpleOnline's compiler

vtrlx:
Only one change here - the compiler will return a header allowing any origin to access /scripts/compiler.php in UmpleOnline. This means that others can write applications (that aren't UmpleOnline) that will interface with the compiler.

More specifically, any instance of Orion will be able to access the compiler using the plugin I'm developing.

### DIFF
--- a/umpleonline/scripts/compiler.php
+++ b/umpleonline/scripts/compiler.php
@@ -23,6 +23,9 @@
 
 require_once("compiler_config.php");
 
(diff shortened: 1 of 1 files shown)


TimLethbridge:
This changes UmpleOnline so I will need to also run a script on the server after merging.

TimLethbridge:
Try it now. Changes are live on the server

vtrlx:
100% working. Thank you.

Issue #708, opened 2016-01-29
Moving from Jet to our own code generator mechanism


vahdat-ab:
JET has been a major technology for model-to-text transfomration, but its development has been stopped because of some diffuculties that developers have to deal with (maybe). Furthermore, we are getting so many issues reagding keeping its integration into Umple.
We would like to get ride of JET and transform all of the JET templates to our own code generation language.

vahdat-ab:
Comments From @TimLethbridge 
- Jet generates files that do the generation for each language; these files call java(Umple) methods to fine-tune generation.
- There is a Nebua framework developed by Umple team that uses its own templates to generate code along with 'generation points' for decision making. We don't have good documentations for this framework yet. We have separately used our inbuild templating for some other generators e.g. state tables. We know that works well.
- I think it ought to be possible to simply replace a Jet template with a version of the same template simply translated into our language. It would be best yet if this translation can be automated (wholy or partly) to prevent the need to do this (semi-)automatically; but Jet creates files in specific formats, and the issue of how Umple's output would be able to substitute for this would need to be addressed.

Diusrex:
Experimented quite a bit with JET and the built in Umple Template.

To clarify the terminology so I don't confuse myself or anyone else, since there are two different stages of generated code:
The **templates** will be transformed into **generated Java code**, which is stored in the repository. Then, this generated Java code is used by the rest of Umple to generate **source code** in various languages. So: 
**template(s)** refers to the **JET/Umple templates**
**generated code** refers to the **Java code generated from the templates**
**source code** refers to the code that is generated by **Umple** not related to templates, and it is made by compiling code written in Umple
(I don't refer to code written by hand in umple, so don't have a term for it)

If someone has suggestions about better names, please say so, I think this will help when describing everything.

### General Overview

It is definitely possible to use the Umple templating to generate the same general structure of classes and calls used in JET. However, without significant parsing of the dependencies between and inside each template (to get the parameters) or changes to Umple's templating, the solution will be semi-automatic due to the function calls issue I discuss lower.

##### Little introduction:

If you are unfamiliar with the JET templates, see [this existing template](https://github.com/umple/umple/blob/master/UmpleToPhp/templates/association_SetOptionalOneToOptionalOne.jet).

I then transformed it to be an [Umple template](http://pastebin.com/9Ca8Xynh).
[...]

PR #776, opened 2016-03-14
Removed Java tab and improved Php template

Diusrex:
## Description

This is a small pull request to set up the repository for starting to allow using the php Umple template instead of JET.

Had to change one template to allow Umple and JET to produce the same output for php.

Outlined in Issue #708, but does not fix much of the spacing issues in Java, to make it easier for everyone still working on Java templates. Further, similar, changes will need to be performed on Java before Umple templates for Java will be able to pass.
## Tests

Changed one Php test, to not expect a tab, and removed unnecessary spaces between two functions.
Also changed all Java tests that included `String outputString = "";` in their `toString()` function to not use a tab.


TimLethbridge:
Looks good to me

 @vahdat-ab if it is OK by you I will let you merge it

vahdat-ab:
great. Thank @Diusrex

PR #780, opened 2016-03-17
Added php Umple templates

Diusrex:
Does not compile them by default, or use them in the build. But I had provided commented out code to build decide to build using them.

For this, see `build/build.umple.xml`, line 99 as example to recompile the php Umple templates, and line 154 for how to use the php Umple templates in the build instead of JET.


TimLethbridge:
I am going to merge this. It looks great. I will also test building with it.

TimLethbridge:
I confirm the build works fine with the generated php files (I changed the commenting in the build file). I would be confident to change the build file now. We should put the generation from UmpleTL at the start of  codegen around line 149.

PR #781, opened 2016-03-17
Php compiles from Umple templates rather than JET by default

Diusrex:
Changed the build file to use Umple templates by default for php. As well, was able to remove all of the generated code for php templates, since they will always be built.

Please note that, when building the Umple templates, it defaults to using the stable version of umple since it should always be available.

There may be a better way to set this up, but I was able to get it to work both with a new, cloned version of the repo as well as with a repo that already existed.


TimLethbridge:
OK Very good. I will merge this one so we can push this forwards.

PR #784, opened 2016-03-19
Ruby compiles from Umple templates rather than JET by default

Diusrex:
Changed the build file to use Umple templates by default for ruby. As well, was able to remove all of the generated code for umple ruby templates, since they will always be built.

Like the switch for php, it defaults to using the stable version of umple since it should always be available.


TimLethbridge:
This looks good. Ignore my overnight comment issue #708. I will merge.

PR #788, opened 2016-03-21
disabling regions with and-crossing transitions

opeyemiAdesina:
In this pull request, we added and-cross transitions to disable orthogonal regions.


opeyemiAdesina:
In particular, we added and-cross transitions to the list of transitions to disable orthogonal regions.

PR #791, opened 2016-03-21
Added Umple templates for Java, build defaulting to JET

Diusrex:
This pull request contains the final Umple templates for Java.

As requested, the build will default to using JET, which will result in 19 failed tests, 16 in Ruby and 3 in Java. Most are due to JET including extra spaces, although some are due to it not adding enough blank lines.

### DIFF
--- a/UmpleToJava/.gitignore
+++ b/UmpleToJava/.gitignore
@@ -1,0 +1,2 @@
+UmpleToJava.java
+src-gen-UmpleTL/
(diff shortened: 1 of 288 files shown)


TimLethbridge:
It has failed on Running cruise.umple.implementation.AssociationBothSidesSortedTest in both CI environments

TimLethbridge:
@Diusrex It is failing on the BothSidesSorted because you included the fixes for the tests that are only valid if UmpleTL is active! But UmpleTL is not active. We have to modify the PR to exclude those three changed files for now, and put them back in when we activate UmpleTl

TimLethbridge:
Also, I can't figure out why you have reverted the building of Ruby, PhP and Sql in codegen

Right now I am testing this PR with java templates activated

TimLethbridge:
Ignore my last comment. 

Note that I have just merged #792 so there could be slight conflicts with this. I think what I am going to do is just merge this one ... have it fail, and then blast things through with another PR that will solve the problem so I can do the release. I have tested this with UmpleTL activated and it is fine.

PR #773, opened 2016-03-13
Fixed incorrect reference comments for umple classes and function locations

bitwizeshift:
This PR fixes incorrectly generated line-reference comments as described in issue #768. In addition to the mentioned issue, it also corrects the line comments before class definitions as in the following example:

**test.ump**

```
use test1.ump;
class A{
    // ...
}
```

**test1.ump**

```
class A{
    // ...
}
```

now produces

```
// line 1 "test1.ump"
// line 2 "test.ump"
public class A
{
 // ...
}
```

instead of 

```
// line 1 "test1.ump"
// line 2 "test1.ump" <-- incorrect name
public class A
{
  // ...
}
```

I'm not sure why git thinks the entire jet-generated files were changed when there are only a few lines changed overall.

Closes #768

Issue #768, opened 2016-03-11
Umple reference comment is not working correctly


vahdat-ab:
When we define a method inside a Umple file, in the generated code we have a reference to Umple file. Looks like, someone has broken it.
Check out the following example

//file: test.ump

```
use test1.ump;
class A{
    void test(){
        //test
    }
}
```

//file: test1.ump

```
use test2.ump;
class A{
    void test1(){
        //test1
    }
}
```

//file: test2.ump

```
class A{
    void test2(){
        //test 2
    }
}
```

**Generated code:**

```
  // line 4 "test2.ump"
  public void test2(){
    //test 2
  }

  // line 4 "test2.ump"
  public void test1(){
    //test1
  }

  // line 49 "test2.ump"
  public void test(){
    //test
  }
```

vahdat-ab:
@bitwizeshift This may be related to your refactoring

bitwizeshift:
Hmm, I'll look into this when I get the chance.

I think this may be related to the `getRelativePath` method.

bitwizeshift:
I just went back to a commit made prior to the pull-request regarding the `getRelativePath` method, and it does appear that this is to blame for the issue. I'm surprised there aren't any tests that would have caught this problem initially.

I'll have to look into this further.

vahdat-ab:
@bitwizeshift  Thanks.

bitwizeshift:
I've found the problem; looks like I forgot to set the source filename in a couple locations in the jet files when I was refactoring the `getRelativePath`.
I'll make the changes and have a PR in a few hours.

PR #798, opened 2016-03-23
Fixed incorrectly generating reference comments and added @umplesource to interfaces

bitwizeshift:
This PR fixes incorrectly generating line statements. This was originally mentioned in issue #768 and was thought to be resolved in PR #773, though originally the fix didn't cover every case.

This also adds generation support for `@umplesource` tags in umple Interfaces, as this feature was previously absent. 

The changes remove some modifications added in a previous PR, and reverts `<UmpleClassifier>.getRelativePath(...)` to rely on a filename -- as the previous implementation was the cause of the issue.


vahdat-ab:
Thanks. For all of my cases it works.

PR #801, opened 2016-03-29
Implemented ability to specify parameters for custom code injections

Shikib:
## Description

This is a working version of the functionality to allow code injections into user defined methods with parameters specified. For example:

```
class A {
  void foo(String a) {
    [...]
  }

  void foo(int a) {
    [..]
  }
}
```

You can now specify the parameters in certain ways to target those methods specifically:

If you want to inject into the first method and only the first method:

```
before custom foo(String) {
  [...]
}
```

Or if you want to inject only into the second method:

```
before custom foo(int) {
  [...]
}
```

Or if you want to inject into both:

```
before custom foo(...) {
  [...]
}
```

This also injects into both (if parameters unspecified -- it is presumed that you don't care about them):

```
before custom foo {
  [...]
}
```

You can also just use multiple code injections on the same line and specify parameters for all of them:

```
before custom foo(int),foo(String) {
  [...]
}
```
## Tests

Cleaned up some of the tests that needed to be changed (for minor and justifiable reasons) due to my changes to the grammar. Added in new test files to test this functionality -- covering all of the examples that I mentioned above.

### DIFF
--- a/cruise.umple/src/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/UmpleInternalParser_CodeClass.ump
@@ -2616,6 +2616,38 @@ class UmpleInternalParser
     return value;
   }
 
+  private String getOperationParameters(Token aToken)
+  {
+    StringBuilder operationParameters = new StringBuilder();  
+    String delimeter = "";
+    for(Token subToken : aToken.getSubTokens())
+    {
+      if(subToken.is("parameterType"))
+      {
+        operationParameters.append(delimeter+subToken.getValue());
+        delimeter = ",";
+      } 
+      else if(subToken.is("parameterTypes")) 
+      {
+        operationParameters.append(delimeter+getOperationParameters(subToken));
+      }
+    }
+    return operationParameters.toString();
+  }
+
+  private ArrayList<String> getOperationsParameters(Token aToken)
+  {
+    ArrayList<String> operationsParameters = new ArrayList<String>();
+    for(Token subToken : aToken.getSubTokens())
+    {
+      if(subToken.is("injectionOperation"))
+      {
+        operationsParameters.add(getOperationParameters(subToken));
+      }
+    }
+    return operationsParameters;
+  }
+
   private String getOperationSource(Token aToken)
   {
     StringBuilder operationSource = new StringBuilder();
@@ -2640,6 +2672,11 @@ class UmpleInternalParser
       {
         operationName.append(comma+subToken.getValue());
         comma = ",";
+      } 
+      else if(subToken.is("injectionOperation")) 
+      {
+        operationName.append(comma+getOperationName(subToken));
+        comma = ",";
       }
     }
     return operationName.toString();
@@ -2659,6 +2696,11 @@ class UmpleInternalParser
 
     CodeInjection injection = new CodeInjection(type,operationName,"",uClassifier);
     injection.setOperationSource(operationSource);
+
+    ArrayList<String> operationsParameters = getOperationsParameters(injectToken);
+    for(String operationParameters : operationsParameters) {
+      injection.addParameter(operationParameters);      
+    }
 
     makeCodeInject(injectToken,injection,cb,uClassifier);
     injection.setSnippet(cb);
--- a/cruise.umple/src/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/UmpleInternalParser_CodeClass.ump
@@ -2640,7 +2640,7 @@ class UmpleInternalParser
     return operationParameters.toString();
   }
 
-  private ArrayList<String> getOperationsParameters(Token aToken)
+  private List<String> getOperationsParameters(Token aToken)
   {
     ArrayList<String> operationsParameters = new ArrayList<String>();
     for(Token subToken : aToken.getSubTokens())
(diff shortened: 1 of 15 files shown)


TimLethbridge:
It is very, very important to never commit changes to  cruise.umple/src/UmpleVersion.ump
Please get rid of this change from the PR.

I also find it odd that you have committed changes to AssociationSortedWithNameSpace_Mentor.java.txt and AssociationSortedWithNameSpace_Student.java.txt
We just changed these before the last release to make tests pass ... are you sure this is properly rebased?

Shikib:
Yes, I have rebased properly. The changes to the two test files entailed nothing more than removing spaces on a line. The tests wouldn't pass because of those spaces. Even reverting my changes to those files now, leads to failures in the tests.

I've removed the UmpleVersion file commit. Sorry about that.

cruise.umple/src/UmpleInternalParser_CodeClass.ump
```diff
@@ -2616,6 +2616,43 @@ class UmpleInternalParser
...
+    }
+    
+    if(aToken.is("injectionOperation")) {
+      return "...";
+    }
+
+    return operationParameters.toString();
+  }
+
+  private ArrayList<String> getOperationsParameters(Token aToken)
```
Nava2: This should be `List<String>` not `ArrayList`.

cruise.umple/src/Umple_Code.ump
```diff
@@ -1481,20 +1482,26 @@ class UmpleClass
...
-/*
         boolean isParameterMatch = true;
-        int i = 0;
-        for(String parameterType : code.getParameters().split(",")) {
-          if(!parameterType.equals(parameters.get(i).getType())) {
-            isParameterMatch = false;
-          }
-          i++;
+
+        if("...".equals(code.getParameters()[opInd])) {
```
Nava2: `code.getParameters()[opInd]` should really be stored in a variable. It's accessed many times.

Nava2:
Sorry, I nitpicked a bit. I needed a short break from my studies. :smile: 

It looks good, otherwise, though.

Shikib:
Thanks Kevin. I fixed/committed the two issues you brought up.

vahdat-ab:
Thanks for the PR.
- The names in Umple are case sensitive so please make sure your solution supports it. Current version doesn't differentiate.
- Furthermore, consider the following example:

```
class A {
    void testData(String i){
        //hello
    }
    void TestData(String i, String j){
        //hello To
        if (x==1) return;
        if (x==0) getData();
        return;
    }
    after custom testData {
        //Injected code
        This is the code.
    }
}
```

It injects code to TestData(String i, String j) while it must respect to names and also parameters. 
- The inject code doesn't have a proper format. It might be because of our new generator. Have a look at it.
- We talked about events of state machines, but as I said we need to talk about it in the issue section.
- If there is no matched joint point, please show a warning regarding that. for example, in the following command related to the above example, there is no method wit `Integer` parameter.

```
    after custom testData(Integer), testData(String,String) {
        //Injected code
        This is the code.
    }
```

It looks like you have a warning for this cases (without parameters). You need to apply it to methods with parameters as well.
- Consider the following example

```
class A{
  data;
  setData(String i,String j){
  }
  after setData(String,String) {
    //this is injected code
  }
}
```

The output is not clear. If it's going to be applied to just auto-generated code why we accept parameters. If you aim at both we need to apply the inject code to both of them. I'm confused about the grammar and output. @TimLethbridge may want to comment on this for more clarification as well.
- I have a question. If you use `custom` keyword, will you inject code for auto-generated ones?

```
class A{
  data;
  setData(String i,String j){
  }
  after custom setData{
    //hey this is a injected code.
  }
}
```

Now, `after` says, any method with the name `setData` without paying attention to parameters. Please clarify this to me. @TimLethbridge may want to comment on this as well.

Shikib:
Thanks for the comments Vahdat! You raised 3 different issues in your comment: (1) injections are not case sensitive, (2) there is no warning for parameter-based injections, (3) how the injection targeting works.

(1) Regarding your example, it seems like code injections were never case sensitive. I didn't change anything with regards to that in this PR. Consider the following example:

```
class A {
  foo;

  after SETFoo {
    // inject
  } 
}
```

This will inject into the setFoo method regardless of the fact that the case of the method name in the injection doesn't match. Based on a preliminary look through the code -- it seems that this is an issue with `StringFormatter.toUnderscore` being called on both the operation name (in the injection) and the method name. I have a rough idea of how to deal with this issue, but I think something like this should be a separate change.

So if we do want to change how Umple handles code injections to make them case sensitive (since they aren't right now), that's a separate issue and a separate PR.

(2) Warnings not being shown when a method name match exists but a parameter match doesn't. 

Sorry, I forgot about the warnings. I can update the PR with this ASAP. (Probably around 10pm PST today).

(3) Confusion about grammar for injections.

Right now the parameters are completely disregarded for auto-generated methods. Specifying them won't change anything about where the injection occurs. I wasn't sure whether this is something we wanted done and I wanted to make sure that it was before I hunted down every call to getApplicableCodeInjections.

> If you use custom keyword, will you inject code for auto-generated ones?

No you won't. In this situation, you'd only inject into user defined methods.

Shikib:
Ok. I fixed the issue with the warnings. It should now display a warning for all situations in which an injection will not take place for injections that specify parameters.

vahdat-ab:
Thanks. I'll create two issues regarding this PR that need separate PR.

PR #802, opened 2016-03-29
Removed all dependencies on Umple from the Parser

bitwizeshift:
## Short Description

This PR removes all dependencies on Umple from the Parser. It extracts a few of the features into Handler classes, `LinkedFileHandler` and `AnalyzerGeneratorHandler` that handle the file linkage and analyzer generator respectively. UmpleFiles are no longer used in the parser at all, with corresponding changes to the appropriate tests and files that make use of the parser.
## Long Description
- Removed `StringFormatter` references in the `GrammarAnalyzer.ump` files. Most of this was able to be handled with simple string concatenation anyway.
- Removed `UmpleCompilerException` from the parser. This was basically just a simple extension of `RuntimeException` that didn't use any of the extensions anyway.
- Moves the main `Analyzer` class to `cruise.umple.analysis`. This is the base class for all analyzers that will be constructed in `GrammarAnalyzer`.
- Changes all calls from `UmpleFile` to `File`. Linked files are now handled as an array of `File` objects supplied in the `parse(...)` method. `UmpleModel` can now be passed the main `File` and the list of linked `File`s in its construction. For the cases that `UmpleFile`s are still passed, a temporary solution is used in `UmpleModel` that splits `UmpleFile`'s `getLinkedFiles()` and translates it into a list of `File` objects. This was done because a large amount of features or tests in Umple rely on the original constructor, and changing it would result in a large amount of changes. That said, this is a step toward deprecating the use of `UmpleFile`.
- Action tokens are now assignable in `RuleBasedParser`. When you assign an action token (which expects the token name and the `ParserAction` to perform) it will cascade the call to the `GrammarAnalyzer` if one is constructed. If one is not constructed, it will automatically populate the `GrammarAnalyzer` with the actions when it is.
- `ParserDataPackage` now contains information about the `GrammarAnalyzer` being used. This is important for being able to create actions that produce `RuleBasedParserThread`s, such as when you encounter a 'use' statement.
- Construction of analyzers, and linkage of files have been extracted out into the two handler classes, `AnalyzerGeneratorHandler` and `LinkedFileHandler` respectively. 
  - `AnalyzerGeneratorHandler` requires an implementation of `generateFromName(String name)` that generates and returns an `Analyzer` object based on the string name of the token. In Umple this generates analyzers using reflection from the `cruise.umple.analysis.` namespace.
  - `LikedFileHandler` requires an implementation of `onFileLink( String input, File[] linkedFiles )` that handles linkage of files and returns a string modified from input. In Umple, this generates a series of "use" statements prior to parsing so that the parser will add the files in-place.
- Implementations of the handler classes are given in `UmpleInternalParser_CodeParserHandlers.ump`. These are just concrete implementations that will be used in the parser, and were placed here since this is where the `RuleBasedParser` gets instantiated to begin with.
## Tests

Some tests have changed, but they were just to reflect the change from `UmpleFile` to `File` in the parser. They are:
- `FilterTest`
- `ModelConstraintTest`
- `UmpleParserFilterTest`
- `UmpleParserStateMachineTest`
- `UmpleParserTest`
- `UmpleParserTracerTest`
- `UmpleUSEParserTest`
- `TemplateTest`

along with the tests, the harness files `TestParseValidation` and `TestResource` were also modified.


vahdat-ab:
@bitwizeshift thanks. Please let's have the above comments in the related issue section so we can keep track of them.

cruise.umple/src/GrammarParsing.ump
```diff
@@ -19,9 +104,19 @@ class RuleBasedParser
   
   public static HashMap<String,ChoiceRule> choicerules = new HashMap<String,ChoiceRule>();
   private static HashMap<String,String> todeclare = new HashMap<String,String>();
+  
+  // List of actions for the parser to perform. This has to be static because of choicerules and todeclare,
+  // but hopefully someday this will change.
+  private static HashMap<String,ParserAction> actionedTokens = new HashMap<String,ParserAction>();
```
Nava2: This should be `Map<String, ParserAction>`, not `HashMap`.

cruise.umple/src/GrammarParsing_Code.ump
```diff
@@ -1,4 +1,102 @@
...
+          List<String> variables = new ArrayList<String>();
+          for(Field field:analyzer.getClass().getDeclaredFields())
+          {
+            Class<? extends Analyzer> clazz = this.getClass();
+            while(clazz!=null&&!"Analyzer".equals(clazz.getSimpleName()))
+            {
+              for(Field mfield:clazz.getDeclaredFields())
+              {
+                if(field.getName().equals(mfield.getName()))
+       	        {
```
Nava2: Spacing

cruise.umple/src/GrammarParsing_Code.ump
```diff
@@ -1,4 +1,102 @@
...
+            {
+              analyzer.addChild(child);
+            }
+	    for(int i=0;i<classes.size();++i)
+	    {
+	      Object object = this.getClass().getMethod("get"+variables.get(i),new Class[]{}).invoke(this);
+	      analyzer.getClass().getMethod("set"+variables.get(i),new Class[]{classes.get(i)}).invoke(analyzer,object);
+	    }
+	  } catch (SecurityException | InstantiationException | IllegalAccessException | IllegalArgumentException | InvocationTargetException | NoSuchMethodException e) {
+	    e.printStackTrace();
```
Nava2: Do you really just want to print the stack trace? If these errors happen it's usually catastrophic failure. Rethrow the checked exceptions in a runtime exception, i.e. [`IllegalStateException`](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalStateException.html).

cruise.umple/src/GrammarParsing_Code.ump
```diff
@@ -1,4 +1,102 @@
...
+  public void prepare(Token token){}
+  
+  /*
+   * Analyze is called to analyze the particular token
+   */
+  public void analyze(Token token){}
+
+  public String toString()
+  {
+    StringBuilder children = new StringBuilder();
```
Nava2: Unused variable.

cruise.umple/src/GrammarParsing_Code.ump
Nava2: Also, the list of exceptions is really long, perhaps it should be split across multiple lines (not clauses, just lines).

cruise.umple/src/GrammarParsing_Code.ump
```diff
@@ -198,9 +313,41 @@ class RuleBasedParser
...
+   * 
+   * @param file the file to parse
+   * 
+   * @return the result of this parsing as a ParseResult
+   */
+  public ParseResult parse(File file){
+    return parse(file, new File[0]);
+  }
+  
   /*
```
Nava2: If this code isn't being used, could it be removed?

cruise.umple/src/GrammarParsing_Code.ump
```diff
@@ -379,9 +527,38 @@ class RuleBasedParser
...
+  
+  // retroactively change the linked file handler if an analyzer is already constructed
+  void setLinkedFileHandler(LinkedFileHandler aLinkedFileHandler){
+    linkedFileHandler = aLinkedFileHandler;
+    if( analyzer != null ){
+      analyzer.setLinkedFileHandler(aLinkedFileHandler);
+    }
+  }
+  
+  LinkedFileHandler getLinkedFileHandler(){
```
Nava2, line 527: Won't this be generated by the field declaration?

cruise.umple/src/GrammarParsing_Code.ump
```diff
@@ -557,22 +720,32 @@ class GrammarAnalyzer {
...
-    setUmpleFile(file);
-    input = file.getPath()+File.separator+file.getFileName();
-    setData(new ParserDataPackage(input));
+    setFile( file );
+    input = file.getAbsolutePath();
+    setData(new ParserDataPackage(input,this));
     getData().setKeys(keys);
     getData().init(null);
-	getData().AddExtraFiles(file);	
+	  getData().linkFiles( getLinkedFileHandler(), linkedFiles );
```
Nava2: spacing

cruise.umple/src/GrammarParsing_Code.ump
```diff
@@ -603,7 +776,7 @@ class GrammarAnalyzer {
     setFailedPosition(parser.getFailedPosition());
     setParseResult(getData().getParseResult());
     if(mustSetupAnalyzers)
-	{
+	  {
```
Nava2: spacing

cruise.umple/src/Parser_Code.ump
```diff
@@ -890,6 +890,20 @@ class Parser
...
+  private static <T extends Closeable> void closeAsRequired(T reader){
+    if (reader != null)
+    {
+      try
+      {
+        reader.close();
+      }
+      catch (IOException e)
+      {
+        e.printStackTrace();
```
Nava2: Same comment about exceptions as before. Throw `IllegalStateException`.

cruise.umple/src/Parser_Code.ump
```diff
@@ -1711,6 +1725,26 @@ class ParseResult
...
+  private static String sanitizeForJson(String input){
+    String ret = "";
+    for (int i=0; i<input.length(); i++)
+    {
+      switch(input.charAt(i))
+      {
+        case '\\':            
+          ret += "\\\\";
+          break;
+          case  '"':
```
Nava2: Spacing

cruise.umple/src/Parser_Code.ump
```diff
@@ -1785,6 +1860,20 @@ class ErrorTypeSingleton
     parse(new TextParser(filename, input));
   }
 
+  private static <T extends Closeable> void closeAsRequired(T reader){
```
Nava2: This is being repeated around, why are we not having this in a method. I thought I saw it in `SampleFileWriter` (a horrible name, I know).

cruise.umple/src/Parser_Code.ump
```diff
@@ -1841,7 +1930,7 @@ class ErrorTypeSingleton
       int severityInt;
 
       if(code == null || severity == null || url == null || formatString == null)
-        throw new UmpleCompilerException("Error parsing error definitions (line " + p.getLineNumber()+")",null);
+        throw new RuntimeException("Error parsing error definitions (line " + p.getLineNumber()+")");
```
Nava2: I'd make this a better exception than RE, use `IllegalStateException` instead. Since that is what it is.

cruise.umple/src/Parser_Code.ump
```diff
@@ -1849,7 +1938,7 @@ class ErrorTypeSingleton
       }
       catch(NumberFormatException e)
       {
-        throw new UmpleCompilerException("Error code must be an integer (line " + p.getLineNumber()+")",null);
+        throw new RuntimeException("Error code must be an integer (line " + p.getLineNumber()+")");
```
Nava2: Prev comment

cruise.umple/src/Parser_Code.ump
```diff
@@ -1858,11 +1947,11 @@ class ErrorTypeSingleton
       }
       catch(NumberFormatException e)
       {
-        throw new UmpleCompilerException("Severity must be an integer between [1,5] (line " + p.getLineNumber()+")",null);
+        throw new RuntimeException("Severity must be an integer between [1,5] (line " + p.getLineNumber()+")");
```
Nava2: Prev comment

cruise.umple/src/Parser_Code.ump
```diff
@@ -1858,11 +1947,11 @@ class ErrorTypeSingleton
...
       }
 
       if(severityInt < 1 || severityInt > 5)
-        throw new UmpleCompilerException("Severity must be an integer between [1,5] (line " + p.getLineNumber()+")",null);
+        throw new RuntimeException("Severity must be an integer between [1,5] (line " + p.getLineNumber()+")");
```
Nava2: Prev comment

cruise.umple/src/UmpleInternalParser_CodeParserHandlers.ump
```diff
@@ -0,0 +1,79 @@
...
+  
+  public Analyzer generateFromName( String name ){
+    try{
+      String className = "cruise.umple.analysis."+name.substring(0,1).toUpperCase()+name.substring(1)+"Analyzer";
+      Class cls = Class.forName(className);
+      Analyzer a = (Analyzer) cls.newInstance();
+      a.setName(name);
+      return a;
+    }
+    catch(ClassNotFoundException e){}
```
Nava2: Do something smarter with these exceptions..

cruise.umple/src/Umple_Code.ump
```diff
@@ -209,9 +229,10 @@ class UmpleModel
     boolean failed = false;
     String input;
     RuleBasedParser parser = new RuleBasedParser();
-    UmpleParser analyzer = new UmpleInternalParser("UmpleInternalParser",this,parser);
-    ParseResult result = parser.parse(umpleFile);
     
+    UmpleParser analyzer = new UmpleInternalParser("UmpleInternalParser",this,parser);
+    //ParseResult result = parser.parse(umpleFile);
```
Nava2: Remove this line

cruise.umple/test/cruise/umple/compiler/UmpleParserTracerTest.java
```diff
@@ -1144,7 +1146,7 @@ private void assertParse(String inCode, String expectedOutput)
 
 	private void assertParse(String filename, String expectedOutput, boolean expected)
 	{
-		UmpleFile file = new UmpleFile(pathToInput,filename);
+    File file = new File(pathToInput,filename);
```
Nava2: spacing

cruise.umple/test/cruise/umple/test/harness/resource/TestParseValidation.java
```diff
@@ -69,7 +69,8 @@ public void assertParsing() {
 		initUmpleModel();
 		RuleBasedParser ruleParser = new RuleBasedParser();
 		UmpleParser parser = new UmpleInternalParser(parserName, umpleModel, ruleParser);
-		ParseResult result = ruleParser.parse(umpleModel.getUmpleFile());
+    File file = new File(umpleModel.getUmpleFile().getPath() + File.separator + umpleModel.getUmpleFile().getFileName());
```
Nava2: spacing

cruise.umple/test/cruise/umple/test/harness/resource/TestResource.java
```diff
@@ -97,7 +97,8 @@ private void initTestResources(){
 	private void parseModel() {
 		RuleBasedParser ruleParser = new RuleBasedParser();
 		UmpleParser parser = new UmpleInternalParser(parserName, umpleModel, ruleParser);
-		ParseResult result = ruleParser.parse(umpleModel.getUmpleFile());
+    File file = new File(umpleModel.getUmpleFile().getPath() + File.separator + umpleModel.getUmpleFile().getFileName());
```
Nava2: spacing

Nava2:
I definitely wrote some comments against code that you just moved around, but while you're refactoring, you should try to fix the code you're touching. 

It does look good though. :+1:

TimLethbridge:
Thanks for all your comments Kevin! It's super to have someone so diligent helping the team.
And thanks

cruise.umple/src/GrammarParsing.ump
bitwizeshift: I only used `HashMap` as the previous design used that as well. (choicerules and todeclare)

cruise.umple/src/GrammarParsing_Code.ump
bitwizeshift: This is the exact same Analyzer that has been used in Umple for a long time, it's location and namespace have simply been moved

TimLethbridge:
My suggestion is for Matthew to do the easy ones, and raise the others (that predated his work) if necessary as a separate PR. We can leave it for Matthew to tell us how far he is willing to go ...

cruise.umple/src/GrammarParsing_Code.ump
bitwizeshift, line 527: Due to the `choiceRules` fields being static, the handlers have to be static in order to work with the initialization around Umple. I added these as member methods for consistency with the current `getAnalyzer()` method.

Short of doing a whole rewrite of the parser, which would require a massive restructure to the tests and other parts of Umple, there was no way around dealing with the static variables.

Nava2, line 527: > Due to the choiceRules fields being static, the handlers have to be static in order to work with the initialization around Umple. I added these as member methods for consistency with the current getAnalyzer() method.

I trust your judgement!

> Short of doing a whole rewrite of the parser, which would require a massive restructure to the tests and other parts of Umple, there was no way around dealing with the static variables.

I guess you know your next project. :wink: But in reality, that make sense, I understand.

cruise.umple/src/GrammarParsing_Code.ump
Nava2: That's okay, as @TimLethbridge suggested, I do think the exceptions should be cleaned up now. It's more code smell than anything -- but if something went wrong, it'd be bad.

bitwizeshift:
@Nava2 I'm just going through all the comments now, and adding my own where appropriate. You raise some good points, but a lot of the way I've modeled my own additions were for consistency with the previous design. Other things were simply just code moved around. I'll make sure to fix what I can! As for the spacing, some of that came from tabs being inserted instead of spaces in the original code (I never noticed it until now)

@TimLethbridge I'm definitely up for fixing what I can, I'm just really swamped with work from it getting to the end of the semester.

cruise.umple/src/GrammarParsing.ump
Nava2: Oh, use a `HashMap`, but store it as a `Map` since the use of the `Map` doesn't care that it's a hash, or link, or anything for that matter. It's an implementation detail that bleeds into poor variable types. 

I can't remember the name of the rule, but you always want to _use_ the most specific and necessary type but _store_ it as the highest general type that matches. It was in Effective Java as some number that I can't remember off hand.

Nava2:
> @Nava2 I'm just going through all the comments now, and adding my own where appropriate. You raise some good points, but a lot of the way I've modeled my own additions were for consistency with the previous design. Other things were simply just code moved around. I'll make sure to fix what I can! As for the spacing, some of that came from tabs being inserted instead of spaces in the original code (I never noticed it until now)

Sounds good to me. Just noting what I saw. It's better to do some fixes as you refactor. Every little bit helps. :+1:

cruise.umple/src/Parser_Code.ump
bitwizeshift: Yeah I'm not 100% sure on how to handle this one.

It was originally in the `SampleFileWriter`, and I wasn't really sure of any other way of handling this. `SampleFileWriter` is used heavily in the tests for the core compiler, and itself exists in a different namespace in Umple. Given that I was trying to remove all ties from the parser, this was one thing I had to extract.

The only reason it's repeated twice is because I can't think of a good reason to add a parent class or trait with this capability to both classes that need them since `Parser` and `ErrorTypeSingleton` don't really share much in common. 

Normally I'm opposed to duplication. That said, given that the method's functionality is relatively simple, that it's only a few lines long, and because it lacks any real ties between the two objects, I figured that this wouldn't really be that huge of a problem.

cruise.umple/test/cruise/umple/compiler/UmpleParserTracerTest.java
bitwizeshift: This one looks like it's actually because the tests were written with tabs instead of 2 space indentations.
Should I just change mine to tabs? Otherwise it's changing the entire file to use spaces, which would be consistent with the rest of Umple, but would attribute every line changed to me.

cruise.umple/test/cruise/umple/test/harness/resource/TestParseValidation.java
bitwizeshift: Same as previous comment

bitwizeshift:
@Nava2 
I just pushed an update that fixes most of your suggestions. 

For the test-cases, I changed my spaces to tabs for now -- because otherwise changing the entire file over will result in it attributing all changes to me (and may make it harder to discern the changes I made in this PR). 

Exceptions are now `IllegalStateException` where necessary. Stack traces are no longer printed, exceptions are thrown instead. `HashMap` was changed to `Map` in the `RuleBasedParser`.

TimLethbridge:
OK good choices. As for the tabs. We need to do a large-scale change of all tabs to spaces to be consistent. We can do that as a separate issue some time.

Nava2:
Great job @bitwizeshift. :+1:

cruise.umple/src/GrammarParsing_Code.ump
```diff
@@ -36,41 +36,37 @@ class Analyzer
...
-	  } catch (SecurityException | InstantiationException | IllegalAccessException | IllegalArgumentException | InvocationTargetException | NoSuchMethodException e) {
-	    e.printStackTrace();
-	  }
+            for(int i=0;i<classes.size();++i)
+            {
+              Object object = this.getClass().getMethod("get"+variables.get(i),new Class[]{}).invoke(this);
+              analyzer.getClass().getMethod("set"+variables.get(i),new Class[]{classes.get(i)}).invoke(analyzer,object);
+            }
+          } catch (SecurityException | InstantiationException | IllegalAccessException | IllegalArgumentException | InvocationTargetException | NoSuchMethodException e) {
+           throw new IllegalStateException("Error instantiating method: " + e.getMessage());
```
Nava2: There is a ctor to just propagate the exception: `IllegalStateException(message, exception)`. This preserves the exception stack and uses the underlying one as a "reason."

Nava2: I really don't mean to be a pedant, but I think these should be changed to use that version.

bitwizeshift: Nothing wrong with being pedantic if it provides a better solution! I would prefer criticism if it contributes to my learning and/or helps to work toward something better :)

I wasn't even aware there was a propagation constructor for exceptions in Java since I've never had a reason to use them, so I'm happy to learn about that. I will update the PR in a few minutes with the changes.

Nava2: :grinning: This is why I made the suggestions. Now you know for later, too.

Issue #804, opened 2016-03-31
not having parameters for joint points that are generated automatically


vahdat-ab:
consider the following example:

```
class A{
  data;
  setData(String i,String j){
  }
  after setData(String,String) {
    //this is injected code
  }
}
```

because we don't use `custom` keyword, we aim at the auto-generated method`setData`for the attribute data. We dont' care about the parameters. Therefore, it should be removed from the grammar.
This will help people to understand if they want to go with maunal methods, they need to use the keyword `custom`

the preferred one:

```
class A{
  data;
  setData(String i,String j){
  }
  after setData {
    //this is injected code
  }
}
```

vahdat-ab:
@Shikib

Shikib:
Ok to make sure I'm understanding properly before I deal with this issue, what you mean is that we shouldn't allow people to specify parameters unless they're targeting custom methods?

Right now, it's possible to just omit the parameters for injections into generated methods -- but including parameters doesn't really do anything. So I should make it impossible to include parameters for anything except custom code injections?

TimLethbridge:
Remember there can be parameters on generated event methods

vahdat-ab:
- I'm just following what you said. The parameters are not considered when we don't use the keyword `custom`. Therefore, there is no need to have them right? They make sense when we have custom methods. If this is correct, so we should prevent developers from writing it. 
  @TimLethbridge  since we know what we generate we don't need to deal with their parameters (based on what understood from other issues).

vahdat-ab:
There is another issue with injected code for auto-generated methods:
consider the following code:

```
class A{
    0..1 -> * B;
     after addB{
    //this is injected code 1
    }
    after removeB {
      //this is injected code 2
    }
   after addBAt {
      //this is injected code 3
   }
   after addOrMoveBAt{
    //this is injected code 4
   }
}
class B{

}
```

Injections 1 and 2 work, but 3 and 4 don't work. This is exactly the reason why I was saying we should  avoid differentiating custom and auto methods. Umple is growing and catching up with them is hard plus we can easily miss stuff (anyhow, one general solution would be a guarantee for the whole life).

Shikib:
Regarding your code example. I briefly looked into the issue just now. It doesn't seem to have anything to do with custom methods or the injection code being incorrect. The issue seems to be that there is no call to `getApplicableCodeInjections` for the addAtMethod or the addOrMoveAt generated methods. 

[...]

PR #805, opened 2016-04-04
displays warning when parameters are specified for generated injections

Shikib:
## Description

Implemented functionality to display warning when parameters were specified. This deals with issue #804.
## Tests

I added in a single test case for this functionality. It tests both that (1) there is in fact a warning printed for the case where there are parameters specified for generated injections and that (2) there is **not** a warning for the case where there are parameters specified for custom injections.

### DIFF
--- a/cruise.umple/src/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/UmpleInternalParser_CodeClass.ump
@@ -2801,6 +2801,11 @@ class UmpleInternalParser
     ArrayList<String> todoExtendedClasses = new ArrayList<String>();
     ArrayList<String> todoInheritedInterfaces = new ArrayList<String>();
 
+    String operationSource = getOperationSource(injectToken);
+    if("".equals(operationSource)) {
+      operationSource = "generated";
+    }
+
     todoExtendedClasses.add(uClass.getName());
     while(!todoExtendedClasses.isEmpty())
     {
@@ -2847,8 +2852,14 @@ class UmpleInternalParser
       }
     }
 
-    for(String operation : operationName.split(","))
+    String[] allOperations = getOperationName(injectToken).split(",");
+    List<String> allParameters = getOperationsParameters(injectToken);
+
+    for(int opInd = 0; opInd < allOperations.length; opInd++)
     {
+      String operation = allOperations[opInd];
+      String currentParameters = allParameters.get(opInd);
+
       Boolean matches = false;
 
       for(String methodName : methodNames) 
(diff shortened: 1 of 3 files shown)


Shikib:
@vahdat-ab I fixed the issue you told me about.

cruise.umple/src/UmpleInternalParser_CodeClass.ump
```diff
@@ -2739,65 +2738,33 @@ class UmpleInternalParser
...
-    return false;
-  } 
 
-  private boolean checkCodeInjectionValidity(Token injectToken, UmpleClass uClass) {
-    ArrayList<String> methodNames = new ArrayList<String>();
-    String operationName = getOperationName(injectToken);
+    return isParameterMatch;
+  }
 
+  ArrayList<String> getAllMethodNames(UmpleClass uClass) {
```
Nava2, line 2766: This should return `List<String>`. Always prefer Interfaces to concrete classes.

cruise.umple/src/UmpleInternalParser_CodeClass.ump
```diff
@@ -2847,10 +2814,58 @@ class UmpleInternalParser
...
+    if("".equals(operationSource)) {
+      operationSource = "generated";
+    }
+
+    for(int opInd = 0; opInd < allOperations.length; opInd++)
+    {
+      String operation = allOperations[opInd];
+      String currentParameters = allParameters.get(opInd);
+
+      Boolean matches = false;
```
Nava2, line 2866: This should be `bool` not `Boolean`. The `Object` forces a box and unboxing everytime you use it as a primitive, like line 2880.

cruise.umple/test/cruise/umple/compiler/UmpleClassTest.java
```diff
@@ -824,6 +824,28 @@ public void getCodeInjectionUnfoundAttributesMethodError_complex()
...
+    boolean generated = false;
+    String code = "class A{a; void foo(int k) { foo(k-1); } after custom foo(int) { return true; } after generated setA(int) { return true; }}";
+    try {
+      UmpleModel model = getModel(code);
+      model.run();
+      ParseResult result = model.getLastResult();
+      List<ErrorMessage> errors = result.getErrorMessages();
+
+      Assert.assertEquals(errors.size(), 2);
+    } catch (Exception e) {
```
Nava2, line 839: This should catch a lower exception type since you want to be specific. Otherwise, it'll catch everything and may pass unknowingly.

Issue #803, opened 2016-03-31
join points in AOP must be case sensitive


vahdat-ab:
Consider the following example:

```
class A {
  foo;

  after SETFoo {
    // inject
  } 
}
```

We expect to see a warning because there is no automatic-generated method called `SETFoo`. However, Umple add injected code into the method `setFoo`.
Umple doesn't pay attention to lowercase and uppercase letters when it comes to AOP joint points. However, completely we have case sensitive concept in Umple. This must be fixed.

vahdat-ab:
@Shikib

Shikib:
I thought this would be a fairly simple change, and it mostly was. I just deleted some code that formatted injection operations/method names when determining relevant code injections for a method.

However, it leads to some tests failing. It seems like it was (at least at one point) intended functionality for Umple to support different naming conventions.

One such test is: `cruise.umple.implementation.CodeInjectionTest.SupportUnderscoreNamingToo`. An example of something that is expected to work in this test case is:

```
class Student
{
  lazy immutable id;
  name;
  defaulted type = "None";
  String[] roles;

  before getDefault_type { print "start getDefault_type"; }
  after getDefault_type { print "end getDefault_type"; }
  after get_roles { print "end get_roles"; }
  before numberOf_roles { print "start numberOf_roles"; }
  after numberOf_roles { print "end numberOf_roles"; }
  before indexOf_role { print "start indexOf_role"; }
  after indexOf_role { print "end indexOf_role"; }

}

```

All of the injections above, are expected (by whoever wrote the test) to inject appropriately. The capitalization of the operation name doesn't match the capitalization of the generated name though `indexOfRole` vs `indexOf_role`. 

So do we no longer want to support different naming conventions like snake case and pascal case and whatever else? What if there's a naming convention where people capitalize everything after the first word (i.e., `setFoo` --> `setFOO`)?

Or should case sensitivity only apply to custom code injections? This would probably make more sense because I think prescribing one naming convention (camel case) onto users of Umple isn't really a good idea -- but it's fair to expect them to maintain consistency between methods they name (custom methods) and injections that they're targeting at said methods. 

@vahdat-ab

TimLethbridge:
There should be consistency between the methods named and the injections targetted. And it seems perhaps only important in custom code.

I guess the reason for the above case is due to ruby where underscores are found.

Shikib:
Ok, so case sensitivity should definitely apply to custom injections. I currently have this working in my local repo.

[...]

PR #812, opened 2016-04-10
Changed code injections to be case sensitive

Shikib:
## Description

Made the change as described in #803.

As I discussed in our last meeting, doing this for Ruby code is super problematic. I've gotten around the issues though. Right now for a target generator of Ruby, if you specify the injection name in the Ruby format or in the typical Umple camel case format, the injection will occur.

However, in general (true for everything), having incorrect case will not allow the injection to occur.
## Tests

I'm not really adding any new functionality -- simply removing existing functionality. I made changes to the test that were required due to my changes. I also removed a number of tests that were intended to specifically test the feature that I changed. Before I submitted this pull request, I made sure that the tests all passed.


Shikib:
I'm not sure why the build is failing on windows. It seems like the first thing that's being done is a gem update, which fails. This is before any of my code is actually built...

Nava2:
So, the problem is actually that it fails to install `ant`. So that's problematic. @TimLethbridge, could you restart the build? It could just be a fluke.

vahdat-ab:
I think there is a problem with appveyor. It shows failure to my PR as well.

Nava2:
@Shikib check #813 for more information about this. It's a problem with Apache Foundation's servers.

vahdat-ab:
Please check the following example:

```
class A{
    x;
    y;
    void setXx(){
        return;
    }
    void setxx(){
        return;
    }

    before set*{
        //This is the extra code.
    }
}
```

the output is thsi

```
 // line 16 "../main.ump"
  // line 16 "../main.ump"
  public void setXx(){
    setXx_Original();
    // line 22 "../main.ump"
    //This is the extra code.
  }

  public void setXx_Original(){

    // line 22 "../main.ump"
    //This is the extra code.
    return;
  }

  // line 19 "../main.ump"
  // line 19 "../main.ump"
  public void setxx(){
    setxx_Original();
    // line 22 "../main.ump"
    //This is the extra code.
  }

  public void setxx_Original(){

    // line 22 "../main.ump"
    //This is the extra code.
    return;
  }
```

It might not rfelated to this PR because looks like you mixed up the name of methods.
What's the application *_Original()? I don't remember we talked about this kind of solution.

Shikib:
@vahdat-ab 

I don't think either of the two issues presented in that example are due to this pull request.

The fact that there is injection into the the custom defined setX methods seems to be because of code that is being called for "Preconditions". I think this code was written a long time ago and might have been forgotten about. Check line 32, 33 of this file: https://github.com/umple/umple/blame/master/UmpleToJava/UmpleTLTemplates/class_MethodDeclaration.ump.

The *_Original issue was also not something I did. Again, I think this is code that somebody else wrote a long time ago that has now been forgotten about.  Look at line 134,138,149 of the file I linked above to see the code that I think is causing this.

I'm not sure what the Precondition/Postcondition code was supposed to do, so I'm not sure whether this is an intended feature or a bug.

I looked at this page to see who wrote the code: https://github.com/umple/umple/blame/7067af06362336472b20317bf1fd17c2d7ba1cb7/UmpleToJava/templates/class_MethodDeclaration.jet#L135. It seems to be someone with the username "Geoffrey Guest."

I think if you revert back to the state of the codebase 4 months ago, this exact code would be produced as well.

vahdat-ab:
@TimLethbridge I'm not sure about the cases related to precondition and postcondition. What I know is that conditions always must be inside `[ ]`. 
Tim, do you think this is because Geoffrey was working on AOP and left the work incomplete?

PR #815, opened 2016-04-11
Namespace issue

vahdat-ab:
I have noticed several inconsistencies in the use of `namespace' specially with mixins. Unfortunately, other people have developed their own code based on these wrong rules. In this PR, I have tried to resolve some of them. However, still there might be some cases that need to be concerted. 
The rules are these.
if there are several class inside a Umple files, each one can have its own namespace. However, if others don't have it will look for the one defined before. If it's a mixin class, the developer  must make sure they both have the same namespaces. Otherwise, the user will get a warning and the Umple will accept the first definition of the namespace. 
namespaces must be defined before classes.

I had to fix many test cases to make this works. I also had to modify a part of Umplificator in which a class had two namespaces. 
I think we should spend more time on developing real-cases studies and make sure the user-manual is concrete and talks about the all wrong or correct cases.

### DIFF
--- a/cruise.umple/src/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/UmpleInternalParser_CodeClass.ump
@@ -4058,17 +4058,17 @@ class UmpleInternalParser
     String currentNameSpace="";
     boolean twice = false;
     boolean resetClass = false;
-    int index=-1;
+    //int index=-1;
     for(int i = 0; i<inToken.getParentToken().getSubTokens().size();i++){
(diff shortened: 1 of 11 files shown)


vahdat-ab:
@TimLethbridge Would you mind checking this PR to make sure the namespaces are working as you expect. The implementation of namespaces inside Umple uses global variables which makes everything complicated and hard to debug. 
@mgarzon You may want to check the Umplificator because I changed one small part in order to make it complied. Don't have any idea about if it affects its correctness or not

PR #838, opened 2016-06-07
Fix for Issue833

amidzak:
This pull fixes the issue of double implements keywords for queued and pooled state machines when the containing class implements an interface.

### DIFF
--- a/UmpleToJava/UmpleTLTemplates/JavaClassGenerator.ump
+++ b/UmpleToJava/UmpleTLTemplates/JavaClassGenerator.ump
@@ -98,17 +98,27 @@ public <<# if (uClass.getIsAbstract()) { append(realSb, "{0} ", "abstract"); } #
 for (StateMachine smq : uClass.getStateMachines())
   {
     if (smq.isQueued())
-    {
-      append(realSb," implements Runnable");
+    { 
+      if (uClass.hasParentInterface() == false){
+        append(realSb," implements Runnable");
+      }
+      else{
+        append(realSb,", Runnable");
+      } 
       break;
     }
     else if(smq.isPooled())
     {
-      append(realSb," implements Runnable");
+      if (uClass.hasParentInterface() == false){
+        append(realSb," implements Runnable");
+      }
+      else{
+        append(realSb,", Runnable");
+      } 
       break;
     }
   }
-
+  
 #>>
 {
   <<#getMemberCode(realSb, model,uClass,gClass,gen,isFirst);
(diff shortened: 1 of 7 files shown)


UmpleToJava/UmpleTLTemplates/JavaClassGenerator.ump
```diff
@@ -98,17 +98,27 @@ public <<# if (uClass.getIsAbstract()) { append(realSb, "{0} ", "abstract"); } #
 for (StateMachine smq : uClass.getStateMachines())
   {
     if (smq.isQueued())
-    {
-      append(realSb," implements Runnable");
+    { 
+      if (uClass.hasParentInterface() == false){
```
Nava2, line 102: should probably just be `if (!uClass.hasParentInterface())`

Nava2, line 102: Is there ever a case that the model has more than a single interface to implement? or it always going to just be `Runnable`?

amidzak, line 102: @Nava2  The problem here is the implementation of code generation for queued and pooled state machines. The implemented class needs to implement Runnable. For other interfaces which are in the model, there is another method that lists all of them.

vahdat-ab, line 102: I just remember I have seen the case in which in the code generator we had two `implements` keywords which would make we cannot compile the java code. Now, I cannot get the example from the top of my head, but if now you're dealing with this case, you may want to check it as well.

vahdat-ab, line 102: @amindzak Please check the issue #840.
You may be able to fix it easily in this part of the code.

amidzak, line 102: @vahdat-ab issue #840 is the same with issue 833. This pr fixes that. When I build on my computer it doesn't fail. Do you know what would be the problem? it doesn't pass AppVeyor.

vahdat-ab, line 102: @amidzak sorry about that. I completely forgot we talked about it before. 
The issue is not related to your PR. You are good to go.

vahdat-ab, line 102: I wonder if you can merge it or not?
If not, I checked your PR, It doesn't show that wrong behavior and I can merge it.

Issue #800, opened 2016-03-23
Graphviz doesn't show properly actions related to specific languages


vahdat-ab:
consider the following example:

```
class A{
  Boolean result = false;
  status {
    s0{
      go-> /Java{result = true;} s1;
    }
    s1{
       go-> /{result = false;}s1;
    }
  }
}
```

In the graphviz, it shows `null` for the action related to the language Java.

![gv](https://cloud.githubusercontent.com/assets/3684908/13993785/39ddedca-f0f8-11e5-97ee-70a6554b112f.png)

PR #861, opened 2016-06-27
 this commit fixes the issue #800.

vahdat-ab:


vahdat-ab:
This fixes the issue#800.

Issue #841, opened 2016-06-08
Singleton class should not be inherited from


amidzak:
In Umple, It is allowed for a class to be inherited from a singleton class. Since the singleton class has a private constructor, and Java does not allow a class to extend a class with private constructor, the generated code is wrong.
`class X{
   singleton;
}
class Y{
  isA X;
}
`
even if we define the class "Y" it is also wrong.
`class X{
   singleton;
}
class Y{
singleton;
  isA X;
}
`
The error is the java code "extends":
`public class Y extends X{`

There are two solutions:
1: Not allowing inheritance of singleton classes by showing an error,
2: Making the constructor of a singleton class "protected" instead of "private" and also forcing the inherited classes to also be singleton. This means children inherit the singleton feature of the parent.

TimLethbridge:
It makes total sense that there should never be allowed to be a subclass of a singleton class. Good catch

PR #866, opened 2016-06-27
Fix for issue #841. Singleton class should not have any subclasses.

amidzak:
This pull request make Umple to generate error when a singleton class has subclasses.
This is done by checking the superclass of every class. if a class has a singleton superclass then it is an error. The example below is one of the illegal cases:
`
        class Airplane
        {
                singleton;
        }
        class F16
        {
                isA Airplane;
       }
`

This Closes issue #841 .


TimLethbridge:
Travis OSX is failing, but it is not to do with this PR. So I am merging it.

PR #888, opened 2016-09-24
Committing fixes for #853

### LINKED ISSUE #853: different behaviour for activities and actions
consider the following example:

```
class A{
    sm{
      s0{
        entry / {s0_entr1();}
        entry / {s0_entry2();}
        e0 -> s1;
        exit /{s0_exit1();}
        exit /{s0_exit2();}
      }
      s1{
        entry / {s1_entry1();}
        entry / {s1_entry1();}
        e0 -> s1;
        exit /{s1_exit1();}
        exit /{s1_exit1();}
      }
  }
}
```

This is the part of the generated code for entries and exits. 

```
 private void exitSm()
  {
    switch(sm)
    {
      case s0:
        // line 8 "model.ump"
        s0_exit1();
        // line 9 "model.ump"
        s0_exit2();
        break;
      case s1:
        // line 15 "model.ump"
        s1_exit1();
        break;
    }
  }

  private void setSm(Sm aSm)
  {
    sm = aSm;

    // entry actions and do activities
    switch(sm)
    {
      case s0:
        // line 5 "model.ump"
        s0_entr1();
        // line 6 "model.ump"
        s0_entry2();
        break;
      case s1:
        // line 12 "model.ump"
        s1_entry1();
        break;
    }
  }
```

As we can see we have both `s0_entry1()` and  `s0_entry2()` in the generated code. The same thing is correct for exit actions `s0_exit1()` and `s0_exit2()`. However, in the state `s1` we have the same structure but this time the actions for entries and exits are exactly the same. However, in the generated code we have just one `s1_entry1()' and`s1_exit1()`.
Looks like we detect duplication and remove it. However, it violates the semantics of having several definitions for entries and actions inside a state. I think we should repeat them in order to have a consistent semantics.

marc22alain:
## Description

Please see my comments in #853 for an explanation of rationale and changes.
Briefly, a one line change to StateMachine.ump allows duplicate entry and exit actions in a single state. 
## Tests

A new test was created, based on the example in the raised issue. Code generated for PHP and Java reflects the fixed issue.

Closes #853.

### DIFF
--- a/cruise.umple/src/StateMachine.ump
+++ b/cruise.umple/src/StateMachine.ump
@@ -181,7 +181,7 @@ class Action
   	codeblock.setCode(lang,code);
   }
   
-  key { actionType, actionCode } 
+  key { actionType, actionCode, position } 
 }
 
 class Event
--- a/cruise.umple/test/cruise/umple/statemachine/implementation/StateMachineTest.java
+++ b/cruise.umple/test/cruise/umple/statemachine/implementation/StateMachineTest.java
@@ -240,6 +240,12 @@ public class StateMachineTest extends StateMachineTemplateTest
   }
 
   @Test
+  public void entryExitActionDuplicates()
+  {
+  assertUmpleTemplateFor("entryExitActionDuplicates.ump",languagePath + "/entryExitActionDuplicates." + languagePath + ".txt", "Duplicate");    
+  }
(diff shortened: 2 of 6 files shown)


vahdat-ab:
Thanks. Great work.

PR #890, opened 2016-09-24
Fixed #834

### LINKED ISSUE #834: Duplicate methods generated when a class implements an interface and has association
The example below has an interface that defines one of the methods that are generated by umple when having an association

``` java
interface X{
        boolean addY(Y ay);
}

class XImp{
        isA X;
        0..1 -> * Y; 
}

class Y{
}
```

When code is generated in java, PHP, Or Ruby (and probably other languages), the generated class has two implementation of `add_y(Y a_y)` . In Ruby it is `addy(Y a_y)`.
The generated code for XImp in Java is below:

``` java

public class XImp implements X
{

  //------------------------
  // MEMBER VARIABLES
  //------------------------

  //XImp Associations
  private List<Y> ies;

  //------------------------
  // CONSTRUCTOR
  //------------------------

  public XImp()
  {
    ies = new ArrayList<Y>();
  }

  //------------------------
  // INTERFACE
  //------------------------

  public Y getY(int index)
  {
    Y aY = ies.get(index);
    return aY;
  }

  public List<Y> getIes()
  {
    List<Y> newIes = Collections.unmodifiableList(ies);
    return newIes;
  }

  public int numberOfIes()
  {
    int number = ies.size();
    return number;
  }

  public boolean hasIes()
  {
    boolean has = ies.size() > 0;
    return has;
  }

  public int indexOfY(Y aY)
  {
    int index = ies.indexOf(aY);
    return index;
  }

  public static int minimumNumberOfIes()
  {
    return 0;
  }

  public boolean addY(Y aY)
  {
    boolean wasAdded = false;
    if (ies.contains(aY)) { return false; }
    ies.add(aY);
    wasAdded = true;
    return wasAdded;
  }

  public boolean removeY(Y aY)
  {
    boolean wasRemoved = false;
    if (ies.contains(aY))
    {
      ies.remove(aY);
      wasRemoved = true;
    }
    return wasRemoved;
  }

  public boolean addYAt(Y aY, int index)
  {  
    boolean wasAdded = false;
    if(addY(aY))
    {
      if(index < 0 ) { index = 0; }
      if(index > numberOfIes()) { index = numberOfIes() - 1; }
      ies.remove(aY);
      ies.add(index, aY);
      wasAdded = true;
    }
    return wasAdded;
  }

  public boolean addOrMoveYAt(Y aY, int index)
  {
    boolean wasAdded = false;
    if(ies.contains(aY))
    {
      if(index < 0 ) { index = 0; }
      if(index > numberOfIes()) { index = numberOfIes() - 1; }
      ies.remove(aY);
      ies.add(index, aY);
      wasAdded = true;
    } 
    else 
    {
      wasAdded = addYAt(aY, index);
    }
    return wasAdded;
  }

  public void delete()
  {
    ies.clear();
  }

  @Override
  public boolean addY(Y aY){
          return false;
  }

}
```

edmundluong:
## Description

Please see #834 for more information on the issue and the rationale behind the solution. 
## Tests

Added a test case based on the example provided in the issue.

### DIFF
--- a/UmpleToJava/UmpleTLTemplates/class_MethodDeclaration.ump
+++ b/UmpleToJava/UmpleTLTemplates/class_MethodDeclaration.ump
@@ -8,6 +8,8 @@ class UmpleToJava {
         
         if(!aMethod.getExistsInLanguage("Java"))
           continue;
+        if(aMethod.existsAlreadyInClass(uClass))
+          continue;
         
           
         Position p = aMethod.getCodePosition();
--- a/cruise.umple/src/Umple_Code.ump
+++ b/cruise.umple/src/Umple_Code.ump
@@ -778,7 +778,38 @@ class Method
 	  }
 	  this. comments = another.getComments();
   }
-
+  public boolean existsAlreadyInClass(final UmpleClassifier umpleClass) {
+	if (umpleClass != null && umpleClass.getMethods() != null) {
+		for (final Method method : umpleClass.getMethods()) {
+			// Ensure the method names match
+			if (!method.getName().equals(this.name)) {
+				continue;
+			}
+			
+			// Check if parameters match
+			boolean isSame = true;
+			if (method.getMethodParameters() != null && this.methodParameters != null) {
+				for (final MethodParameter parameter : this.methodParameters) {
+					boolean parameterExists = false;
+					for (final MethodParameter otherParameter : method.getMethodParameters()) {
+						if (parameter.getName().equals(otherParameter.getName()) && parameter.getType().equals(otherParameter.getType())) {
+							parameterExists = true;
+							break;
+						}
+					}
+					
+					if (!parameterExists) {
+						isSame = false;
+					}
+				}
+			}
+			
+			return isSame;
+		}
+	}  
+  
+  	return false;
+  }
   public boolean getExistsInLanguage(String lang)
   {
     if (!hasMethodBody()){
--- a/cruise.umple/test/cruise/umple/implementation/AssociationInterfaceTest.java
+++ b/cruise.umple/test/cruise/umple/implementation/AssociationInterfaceTest.java
@@ -1,0 +1,12 @@
+package cruise.umple.implementation;
+
+import org.junit.Test;
+
+public class AssociationInterfaceTest extends TemplateTest {
+	
+	@Test
+	public void shouldNotGenerateDuplicateMethods() {
+	    assertUmpleTemplateFor("AssociationInterfaceTest.ump", languagePath + "/AssociationInterfaceTest."+ languagePath +".txt", "Person");
+	}
+
+}
--- a/cruise.umple/test/cruise/umple/implementation/AssociationInterfaceTest.ump
+++ b/cruise.umple/test/cruise/umple/implementation/AssociationInterfaceTest.ump
@@ -1,0 +1,11 @@
+interface Student {
+	boolean addCourse(Course course);
+}
+
+class Person {
+	isA Student;
+    0..1 -> * Course; 
+}
+
+class Course {
+}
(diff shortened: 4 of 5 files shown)


cruise.umple/src/Umple_Code.ump
```diff
@@ -778,7 +778,38 @@ class Method
...
+					boolean parameterExists = false;
+					for (final MethodParameter otherParameter : method.getMethodParameters()) {
+						if (parameter.getName().equals(otherParameter.getName()) && parameter.getType().equals(otherParameter.getType())) {
+							parameterExists = true;
+							break;
+						}
+					}
+					
+					if (!parameterExists) {
+						isSame = false;
```
amidzak, line 802: parameterExists is false if there is no parameter which exists in both methods. It is true if at least one of the parameters are the same.

cruise.umple/src/Umple_Code.ump
```diff
@@ -778,7 +778,38 @@ class Method
...
+					}
+				}
+			}
+			
+			return isSame;
```
amidzak, line 807: This isSame is the isSame for the last parameter, but you want it to be true if all parameters are the same. also for the last parameter isSame is true if there is no parameter that exists in the other method. But you want to return true if methods are the same

UmpleToJava/UmpleTLTemplates/class_MethodDeclaration.ump
```diff
@@ -8,6 +8,8 @@ class UmpleToJava {
         
         if(!aMethod.getExistsInLanguage("Java"))
           continue;
+        if(aMethod.existsAlreadyInClass(uClass))
```
amidzak, line 11: The method only checks for the methods in the class's model, but the issue is with the methods that are generated by the UmpleTL and are not in the model.

edmundluong:
Not sure why the compilation is failing. It seems that generated Java code in umple\cruise.umple\src-gen-umple\cruise\umple\docs\DocumenterMain.java is breaking due to an extra bracket being introduced somewhere.

```
compile:
     [echo] UmpleCore: C:\projects\umple\build\build.umple.xml
     [echo] Resolving configuration: core
[ivy:resolve] :: loading settings :: file = ivysettings.xml
[ivy:retrieve]  conflict on C:\projects\umple\dist\libs\core\asm.jar in [core]: 5.0.1 won
     [echo] Resolving configuration: test
    [javac] Compiling 1335 source files to C:\projects\umple\cruise.umple\bin
    [javac] C:\projects\umple\cruise.umple\src-gen-umple\cruise\umple\docs\DocumenterMain.java:351: error: class, interface, or enum expected
    [javac] }
    [javac] ^
    [javac] 1 error

```

I cannot figure out how this bracket is added, but commenting out the two lines I changed in class_MethodDeclaration.ump (see [here](https://github.com/umple/umple/pull/890/files?diff=split#diff-002be8b4048e75b1a1117b567fe7b203R11)) makes the build pass again.  

Will need to investigate further why this behavior is being observed

UmpleToJava/UmpleTLTemplates/class_MethodDeclaration.ump
edmundluong: Yes, the issue is with the methods that are generated by the UmpleTL, however, isn't there information about those methods on the model and its associations?

I was tracing through the debugger using the example snippet of code in the issue #834, and I was able to find the method on the class which was already generated, so this conditional should stop the code under this line from generating the method stub for the interface in addition to that method that was already previously generated.

cruise.umple/src/Umple_Code.ump
edmundluong: This loop checks for the case whether both methods in question have parameters or not, if they do:
1. Loop through each parameter in MethodA
2. Compare each parameter in MethodA with all parameters in MethodB, if there is a single match, we are good to move onto the next parameter for MethodA
3. Only in the case where the parameter in MethodA was not found at all in MethodB's parameters list will it trigger the conditional that sets the isSame flag to false

Hopefully I did not make a mistake in the logic here, but it seems that this should work fine for determining whether a method is the same as another method by checking the method name, as well as all of its parameters if it has any. 

I went with a "prove me wrong" approach with this loop, so only in the case where a parameter is not found will it return false for isSame

edmundluong:
Closing pull request in favor of revised fix

PR #906, opened 2016-10-28
Js diagram drawing library

marc22alain:
## Description

This pull request is to show a basic implementation of the JointJS library to render UML class models in the umpleonline web site.

A new diagram type is added in order to display the JointJS rendering. A keyboard short-cut (ctrl-j) will also enable the option. Zooming of the view is also possible by simultaneously pressing the control key while scrolling the mouse to swiping a touchpad (touchpad tested with OSX).
## Tests

No behaviour tests have been added to control the new diagram type.


umpleonline/scripts/umple_action.js
```diff
@@ -1571,6 +1590,63 @@ Action.updateUmpleDiagramCallback = function(response)
...
+          var paperWidth = paper.options.width;
+          // scaleFactor is either 1.1 or 0.9
+          var scaleFactor = 1 + (Math.abs(event.deltaY) / (event.deltaY * 10));
+          // console.log(scaleFactor);
+          paper.setDimensions(paperWidth * scaleFactor, paperHeight * scaleFactor)
+          // paper.height = paperHeight * scaleFactor;
+          // paper.width = paperWidth * scaleFactor;
+
+          // console.log(paper);
+          console.log(paper.options);
```
Nava2: Shouldn't leave console logging in a "production" environment, same with commented code, unless its relevant.

umpleonline/scripts/umple_page.js
```diff
@@ -789,6 +797,7 @@ Page.getSelectedExample = function()
         Action.changeDiagramType({type: "GVClass"});
       }
       else {
+        // TODO: determine what behaviour should occur w/r/t JointJS
```
Nava2: Looks like you missed finishing something?

umpleonline/scripts/jjs/jjs_parse.js
```diff
@@ -0,0 +1,117 @@
...
+	},
+
+	makeClasses: function (model) {
+		var i, text, UMLclass, max = model.umpleClasses.length, classes = new Array();
+		var new_class;
+
+		for (i = 0; i < max; i++) {
+			UMLclass = model.umpleClasses[i];
+
+			if (UMLclass.isAbstract == "true") {
```
Nava2: You can just do: `!!UMLclass.isAbstract` instead, but you also shouldn't be using `==` in JS, always use `===` (you use it elsewhere).

umpleonline/scripts/jjs/jjs_parse.js
```diff
@@ -0,0 +1,117 @@
...
+				symbol = "-";
+				break;
+			default:
+				symbol = "-";
+			}
+		return symbol;
+	},
+
+	makeClasses: function (model) {
+		var i, text, UMLclass, max = model.umpleClasses.length, classes = new Array();
```
Nava2: Why are all of your variables defined here? Define them as you use them.

umpleonline/scripts/jjs/jjs_parse.js
```diff
@@ -0,0 +1,117 @@
...
+		var new_class;
+
+		for (i = 0; i < max; i++) {
```
Nava2: Use: `model.umpleClasses.forEach()` instead.

umpleonline/scripts/jjs/jjs_parse.js
```diff
@@ -0,0 +1,117 @@
+var JJsParse = {
+	addToArray: function (item) {
```
Nava2: This function doesn't seem useful. It's exactly equivalent to doing: `[item]`.

umpleonline/scripts/jjs/jjs_parse.js
```diff
@@ -0,0 +1,117 @@
...
+			attributes.push(text);
+		}
+		return attributes;
+	},
+
+	addMethods: function(meths) {
+		// {type: "void", name: "setDoorTimer", parameters: "Integer", visibility: "public", isAbstract: "false"}
+		// console.log(meths);
+		var i, text, method, max = meths.length, methods = new Array();
+		for (i = 0; i < max; i++) {
```
Nava2: Another `forEach()` and more variables should be "more local."

umpleonline/scripts/umple_action.js
```diff
@@ -1571,6 +1590,63 @@ Action.updateUmpleDiagramCallback = function(response)
         // jQuery("div.umpleClass").addClass("unselectable");
       }
     }
+    else if(Page.useJointJSClassDiagram) {
+      // console.log(diagramCode);
+      var model = JSON.parse(diagramCode);
+
+      var umpleCanvas = jQuery("#umpleCanvas");
+      umpleCanvas.html('<div id="myholder"></div>');
```
Nava2: Should probably use a better id than `myholder`, it's just not that accurate to what it is.

umpleonline/scripts/umple_action.js
```diff
@@ -1571,6 +1590,63 @@ Action.updateUmpleDiagramCallback = function(response)
...
+
+          // console.log(paper);
+          console.log(paper.options);
+          // console.log(paper.svg.height());
+          paper.scaleContentToFit({padding: 15});
+        }
+      };
+
+      // using the umpleCanvas as the mouse wheel event target, as it is a stable entity
+      var paperHolder = document.getElementById("umpleCanvas");
```
Nava2, line 1633: Use JQuery here, it's nicer: 

``` js
var $paperHolder = jQuery("#umpleCanvas");
$paperHolder.on('mousewheel', function () { ... })
```

umpleonline/scripts/umple_action.js
```diff
@@ -1571,6 +1590,63 @@ Action.updateUmpleDiagramCallback = function(response)
...
+
+      if (paperHolder.addEventListener) {
+        // IE9, Chrome, Safari, Opera
+        paperHolder.addEventListener("mousewheel", MouseWheelHandler, false);
+        // Firefox
+        paperHolder.addEventListener("DOMMouseScroll", MouseWheelHandler, false);
+      }
+      // IE 6/7/8
```
Nava2, line 1641: Do we actually support IE < 10?

TimLethbridge:
Thanks for all the suggestions kevin. I think Marc can consider these pretty easily

Nava2:
JavaScript is a strange beast, no problem.

umpleonline/scripts/umple_action.js
marc22alain: Done - commented out now.

umpleonline/scripts/umple_page.js
marc22alain: Now I see why I put that off, I couldn't understand it. It looks like an unreachable branch to me.

umpleonline/scripts/jjs/jjs_parse.js
marc22alain: This is so because the JSON arrives with the truth values as type String, and the JSON.Parse() does not change the type to Boolean.

I have changed the `==` to `===`.

umpleonline/scripts/jjs/jjs_parse.js
marc22alain: Ah! Nice pattern, it is new to me.
It's much cleaner now, and I will implement this in the other methods that also apply.

umpleonline/scripts/jjs/jjs_parse.js
marc22alain: Done, deleted.

umpleonline/scripts/jjs/jjs_parse.js
marc22alain: Done.

umpleonline/scripts/umple_action.js
marc22alain: Changed to `jjsPaper`.

umpleonline/scripts/umple_action.js
Nava2: Remove "dead code," though. If it's commented, it's not needed.

umpleonline/scripts/umple_action.js
marc22alain: Your question has prompted me to take a deeper dive into cross-browser compatibility.
The code that I have comes from [here](https://www.sitepoint.com/html5-javascript-mouse-wheel/). It works for me in OSX 10.9.5 with Chrome, Safari, and Firefox.

The `mousewheel` event is not implemented across all browsers in jQuery, so we would end up having to add a different event listener for each browser type anyway.

Finally, I discovered that Firefox captures the control-key modifier to perform a window-level zoom. So I have changed the zoom modifier to the alt key (option key on Mac). I am thinking that we somehow need to inform the user of this.

umpleonline/scripts/umple_action.js
marc22alain: I have not tested in any version of IE.

PR #909, opened 2016-10-29
C++ Code generation support for Mac

ahmedvc:
Support real-time code generatio for Mac
Sync composite-structure and other constraints with the latest model
updates


TimLethbridge:
Appveyor failed, but this was on a time-sensitive test. I suspect it will be OK I just reran appveyor

PR #917, opened 2016-11-14
Distributable RMI primary changes in generated code

amidzak:
This pull request adds distributable(using RMI) and proxy pattern feature into the grammar and generated code. The generated code is still incomplete (Does not work for more than 1 component therefore does not distribute the system), but the changes to grammar are final.

1- Grammar changes:

1.1- a new directive is added: “distribute” can be “on”, “off”, or “forced”
Examples:
distribute on; // means the classes may be distributed. this is the default setting.
distribute off; // means the classes may NOT be distributed
distribute forced; // means all of the classes are distributed

1.2- a new keyword for classes and interfaces called “distributableRMI”
Examples:
class A {distributableRMI;}	// This means the class is distributable using RMI technology
interface IA{distributableRMI;} // Means every class implementing this interface will be distributable and the interface for communications in IA

1.3- a new keyword for classes called “proxyPattern” which means we want a default interface and a proxy file be generated.

2- Changes in the model:
2.1- UmpleModel class has three new boolean attributes:
“distributeOn” is true if distribute is “on” or “forced”.
“distributeForced ” is true if distribute is “forced”.
“distributed” is true only if there is at least one distributed class.

2.2- UmpleClassifier class has a boolean and a string attribute:
“isDistributable” which is true only if the keyword “distributableRMI” is used for the class or interface.
“distributeTechnology” which is RMI by default.( other technologies may added later)

2.3- UmpleClass has three new boolean attribute:
“isDistributed” is true only if the class is distributed for code generation purposes.
“hasProxyPattern” is true only if the class has proxy pattern for code generation purposes. this is always true if the class is distributed.
“needsDefaultInterface” is true if class is distributed or has proxy, unless it has a direct distributable parent interface.

3- Changes in the generated code:
for Classes that are distributable or have proxy pattern, the name of the class is changed to NAME+”Impl”. The generated class is parsed and all of the public methods are used to generate a proxy with the same name of the class and an interface with “I”+NAME.
The proxy delegates all of the public methods of the class and also overrides constructor. The proxy has three constructor to accept component number or component name. It uses the object factory to instantiate the actual class.
The proxy has the same name as the class so that we don’t need to change other classes.

UmpleObjectFactory class and IUmpleObjectFactory interface are generated using new templates.
For classes with proxies; changes to class generator templates are done to replace “this” reference to a reference realSelf which is a reference to the proxy of the class in method calls of generated codes for association. because no class other than the proxy should have access to it.

4- implementation/java/DistributedClassTest.java
Tests for changes in model when directives are used.
Tests for changes in model when keywords are used.
Tests for code generation for UmpleObjectFactory, proxies, distributed classes and interfaces.
5- To add;
reading from configuration file to have more than one components.
RMI connections to have more than 1 components.
add tests for the new features.


TimLethbridge:
The PR succeeded. The issue with Appveyor is not with the PR. So I will merge it.

PR #918, opened 2016-11-16
Fixing disabled shortcuts

marc22alain:
This pull request fixes the bug about the lack of yellow highlighting for the JointJS diagram when a user clicks on its background.
It also addresses the bug about disabled shortcuts when the JointJS diagram is active, or when any diagram is active and no element has focus.

Here is the list shortcuts that have been enabled when no element has focus:
diagram selections:
- ctrl-e
- ctrl-j
- ctrl-g
- ctrl-s
- ctrl-l

view selections:
- ctrl-t, ctrl-alt-shift-t
- ctrl-d
- ctrl-n, ctrl-alt-n

diagram entity selections:
- ctrl-a
- ctrl-m
- ctrl-r
- ctrl-i
- ctrl-k

The following shortcuts have been left unchanged:
class object moves:
- up, shift-up
- down, shift-down
- left, shift-left
- right, shift-right

diagram entity additions:
- g
- a
- c

### DIFF
--- a/umpleonline/scripts/umple_action.js
+++ b/umpleonline/scripts/umple_action.js
@@ -1591,7 +1591,6 @@ Action.updateUmpleDiagramCallback = function(response)
       }
     }
     else if(Page.useJointJSClassDiagram) {
-      console.log(diagramCode);
       var model = JSON.parse(diagramCode);
 
       var umpleCanvas = jQuery("#umpleCanvas");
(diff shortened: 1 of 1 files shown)


umpleonline/scripts/umple_action.js
```diff
@@ -2039,109 +2041,70 @@ Mousetrap.bind(['shift+right'], function(e){
 
 //Functions for changing diagram view - change consistently in umple_page.js
 Mousetrap.bind(['ctrl+e'], function(e){
-  if(jQuery('.focus').length != 0)
```
Nava2, line 2042: What was the point of this before? Just curious.

TimLethbridge: Good question. This needs a thorough manual test. I suspect there might be something working differently. I will do the manual test, but it may be a day or two.

TimLethbridge:
PR seems to work ok in my manual testing so will merge and promote to umpleonline/test

PR #927, opened 2016-12-06
Adding the ability to make state machine diagrams

marc22alain:

## Description

This PR extends the work that implements the JointJS diagramming library, and now enables the display of state machine diagrams.

Scope of work includes:
- a new code generator that produces a JSON description of the project classes and their nested state machines; its new language name is "JsonMixed"
- complier.php is revised to be able to recognize this new language option and make the correct call to umplesync.jar
- umple_action.js is also revised to recognize the new language option and initiate the correct diagram
- a new plug-in to the JointJS library is created to display a more UML standard-compliant state machine diagram: joint.shapes.uml-state-machine.js
- the ability parse this new data and instantiate the diagram is added to jjs_parse.js
- the JS library dependencies are updated to enable auto-layout of the state machine diagram

 Functions that are working:
- the class diagram now shows little red icons in those classes which have state machines; clicking on this icon will trigger the display of that class's state machine
- the state machine diagram includes a 'back' button, that then redraws the class diagram

Unimplemented functions, issues, and bugs:
- the generator does not yet parse 'do' actions
- the generator does not yet parse event parameters
- the 'Show View' selector for Actions does not hide actions
- the diagram parser is not set up to instantiate multiple top-level state machines in a single class
- some transition code is extremely long and all of it is being displayed, so need to provide the user with the means to show/hide it 
- clicking on some parts of a transition will cause it to be deleted; the UI does provides no warning of this result
- pseudo-start states of nested machines are not getting included within their parents' boundaries
- the diagramming code is not yet able to reliably render systems that have more than one level of nesting
- the 'Final' state has not been modified to be able to show actions ('entry' and 'do')
- application of the joint.css rules is not working properly
- states encapsulated within a parent are not separately movable
- transitions with additional vertices are not moving with their parents
- the auto-layout is packing all elements too closely together 

I may be confused about internal transitions vs self-transitions, so may have incorrectly implemented these as internal transitions.

Finally, this PR also includes earlier fixes such as:
- fixing a constant zoom level on the initial display
- adjusting the class colour


vahdat-ab:
@TimLethbridge  I just merge this one and I think it's good to make it available on the test site so we can test and document bugs.

vahdat-ab:
I'll do it when the current build of the previous merge is done

Issue #831, opened 2016-05-18
Intermittent build failures due to timing issues in cruise.statemachine.test.DoActivityTest


opeyemiAdesina:
My effort to commit little changes to the server is unsuccessful. The following is an excerpt from the log. Hope this can  help fix the issue. On my machine, nothing fails but building via travis is always failing.    

```
[junit] Running cruise.statemachine.test.DoActivityTest
[junit] Tests run: 8, Failures: 0, Errors: 1, Time elapsed: 36.854 sec
```

BUILD FAILED
/home/travis/build/umple/umple/build/build.testbed.xml:144: The following error occurred while executing this line:
/home/travis/build/umple/umple/build/build.testbed.xml:158: Tests failed

See also #591 that may be related or may have a suggested solution

TimLethbridge:
have you opened the QA page? It should tell you more. I have seen this test randomly fail in the past, however. WHen you say 'always failing', have you tried rerunning it? How about creating a pull request

TimLethbridge:
I just ran Travis again ... it passed fine now. This is an intermittent (relatively rare). I already ignored a couple of similar cases See #757

TimLethbridge:
This will be challenging to fix given that it seems to occur rarely, but more commonly when the CPU is heavily loaded. Try to load down your machine to reproduce.

TimLethbridge:
This is happening about 1/6 of the time on travis. If somebody wants to try analysing it or maybe running it 100 times with some kind of logging to see if we can detect the cause, it would be helpful.

TimLethbridge:
This issue is happening less often now. And it doesn't affect end-users, so I an downloading its priority a bit

TimLethbridge:
This issue is still a problem

TimLethbridge:
Fixed (we hope) by #2417

Issue #766, opened 2016-03-09
inconsistency of calling of exit and entry in state machine self transitions


vahdat-ab:
In the generated code for Umple state machines, `exit` and `entry` parts of states are dealt in different ways. We know that cyclic transitions can be `internal` or `external`. As long as I know, Umple consider the cyclic transitions as an `external` one. Therefore, it must run `entry` and `exit` parts for every cyclic transiton. However, In the generated code, we can see the `entry` part is executed, but the `exit` part doesn't. Let's have a look at an example:

```
class A{
  boolean result = false;
  sm{
    created{
      exit /{execute_exit_code();}
      entry /{execute_entry_code();}
      init [b==false] -> created;
      init [b==true] -> initialized;
    }
    initialized{
      getback ->created;
    }
  } 
}
```

in the generated code, we have the following code for the event `init`:

```
  public boolean init()
  {
    boolean wasEventProcessed = false;

    Sm aSm = sm;
    switch (aSm)
    {
      case created:
        if (b.equals(false))
        {
          setSm(Sm.created);
          wasEventProcessed = true;
          break;
        }
        if (b.equals(true))
        {
          exitSm();
          setSm(Sm.initialized);
          wasEventProcessed = true;
          break;
        }
        break;
      default:
        // Other states do respond to this event
    }

    return wasEventProcessed;
  }
```

As we can see, the method `exitSm()` is not executed for the cyclic case. However, the method `execute_entry_code()` is executed always through the method `setSm`

```
  private void setSm(Sm aSm)
  {
    sm = aSm;

    // entry actions and do activities
    switch(sm)
    {
      case created:
        // line 7 "model.ump"
        execute_entry_code();
        break;
    }
  }
```

tlaport4:
There's logic to ignore the exit action when it's a self transition. When this logic is removed, many tests fail. Should I be updating the tests to pass in this situation?

Should both the entry and exit code be executed for self transitions? In this example, it would look like this:

```
  public boolean init()
  {
    boolean wasEventProcessed = false;

    Sm aSm = sm;
    switch (aSm)
    {
      case created:
        if (b.equals(false))
        {
          exitSm();
          setSm(Sm.created);
          wasEventProcessed = true;
          break;
        }
        if (b.equals(true))
        {
          exitSm();
          setSm(Sm.initialized);
          wasEventProcessed = true;
          break;
        }
        break;
      default:
        // Other states do respond to this event
    }

    return wasEventProcessed;
  }
```

[...]

PR #932, opened 2016-12-15
Fixes issue 766

tlaport4:

## Description

Issue 766 - added logic to make exit actions be called even on self transitions

## Tests

Testbed - added the test ExitActionSelfTransition. It executes the state machine state self transition and verifies that the exit action was executed.

StateMachineTest 
- updated the entryExitActionDuplicates test to have the exit action for the state machine's self transition
- created the test exitActioNSelfTransition - verifies that an umple class that has an exit action and state self transition creates the correct code


tlaport4:
Can this check be rerun? It looks like it's failing because of this issue: https://github.com/umple/umple/issues/831

TimLethbridge:
Ignore the appveyor failed msg. Seems to be fine when I look at appveyor

vahdat-ab:
check my comments in the issue #766

PR #938, opened 2016-12-20
Issue760

tlaport4:

## Description

Now there is logic to check if guards are equivalent or not. It will do most basic comparison between the guards. For example, two guards like [(x + y) < z] and [z >= (y+x)] would be confirmed as equivalent. However, some equivalent guards will not be deems equivalent. For example: [(2+5) > 5] and [7 > 5] would not be resolved. Also correctly applies negation and deals with parenthesis. For example, [ ((a > b) || (variable && (!(b >= a) && (c < b))))] and [ (((!(b >= c) && (b >= a)) && variable) || (a > b))] would be equivalent. For more examples, refer to the test equivalentGuards.ump 

If the events are the same and the guards are equivalent, the transition will be ignored and a warning will be displayed.

## Tests

Added test equivalentGuards to StateMachineTest.java. Has multiple guards with many equivalent guards that get ignored. Each "group" of transitions in the file are equivalent guards, and only one from each gets through.

**Note:** This template is a guide, remove sections as applicable.


tlaport4:
The failure is that DoActivityTest again

PR #942, opened 2016-12-31
Ensuring latest php is used

TimLethbridge:
Appveyor was failing as it was getting an older php


TimLethbridge:
Updating appveyor version of php

PR #955, opened 2017-01-17
Fix Traffic Lights B UmpleOnline Example

### LINKED ISSUE #593: State machines with capitalized names fail when referenced within other state machines
Originally reported on Google Code with ID 695

---

1. Try the following umple code with two state machines - sm and Other:

``` java
class A 
{
  sm 
  {
    s1 
    {
      entry / {Other = Other.os2;}
    }
  }

  Other
  {
    os1
    {

    }
    os2
    {

    }
  }
}
```

```
Notice that entry into s1 of sm causes other to move into state os2.
```
1. Generate Java code for this Umple model.
2. Compile the Java code.

The Java compilation fails, with the following error:

```
A.java:60: error: cannot find symbol
      Other = Other.os2;
      ^
  symbol:   variable Other
  location: class A
1 error
```

This code does not fail Java compilation if the second state machine is named with
a lowercase name (ie "other") nor does it fail if the first state machine ("sm") does
not reference the second state machine (ie. if the entry action is removed).

This bug is present in the "Traffic Lights B" example on UmpleOnline.

**Please use labels and text to provide additional information.**

---

Reported by @CraigBryan on 2015-03-31 13:55:59

jblang94:
## Description

This PR fixes the Java compilation issue described in [issue 593](https://github.com/umple/umple/issues/593) . The "Traffic Lights B" UmpleOnline example now uses the state machine setter methods to change the state of its state machines. The timer values in the example have also been converted into seconds. I removed the "@@@skipcompile" and "@@@skiprubycompile" directives because both issue 694 and 695 have been fixed.

## Tests

No additional tests were added for this fix. I confirmed that the changes passed using   
"ant -Dmyenv=local -f build.exampletests.xml allUserManualAndExampleTests".

Closes #593

### DIFF
--- a/umpleonline/ump/TrafficLightsB.ump
+++ b/umpleonline/ump/TrafficLightsB.ump
@@ -3,14 +3,14 @@ class IntersectionTimerDriven {
   
   // Traffic lights are green for 25s in one direction
   // and 20s in the other. 
-  Integer directionOneGoTime=25000;
-  Integer directionTwoGoTime=20000;
+  Integer directionOneGoTime = 25;
+  Integer directionTwoGoTime = 20;
 
   // Lights are yellow for 5s
-  Integer yellowTime=5000;
+  Integer yellowTime = 5;
 
   // Pedestrians have 10s extra warning
-  Integer extraPedestrianWarning=10000;  
+  Integer extraPedestrianWarning = 10;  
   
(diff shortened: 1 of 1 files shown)


vahdat-ab:
Thanks @jblang94

PR #968, opened 2017-01-22
Moving the source for simpletest to github instead of sourceforge

TimLethbridge:
Travis was failing on attempting to obtain the simpltest php testing utility from sourceforge. This PR moves the sourcing of that to be from github. Minor renaming turned out to be required since the untarred directory had a different name


TimLethbridge:
Hopefully fixing travis on MacOS X by moving simpletest source to be obtained at github instead of sourceforge

PR #972, opened 2017-01-24
Fixed #970

### LINKED ISSUE #970: Type of association reference list is lower case in UmpleOnline when clicked on "show tabs on different tabs"
Considering the following Umple code:
```java
class D{
}
class C{
        * -- * D;
}
```
The following code is generated
```java
public class C
{
 
  //------------------------
  // MEMBER VARIABLES
  //------------------------
 
  //C Associations
  private List<d> ds;
 
  //------------------------
  // CONSTRUCTOR
  //------------------------
 
  public C()
  {
           ds = new ArrayList<d>();
```
`List<d>` is wrong it should be `List<D>`.
This problem seems to be a occurring only in the online version.

Nava2:
## Description

When using `jQuery#html()` it will format html tags, this wasn't being done in the non-tabbed way, but was being done when using tabs. I replaced it with `.innerHTML`. 

Closes #970. Please verify that this works (it is working locally).

### DIFF
--- a/umpleonline/scripts/umple_page.js
+++ b/umpleonline/scripts/umple_page.js
@@ -867,9 +867,8 @@ Page.showGeneratedCode = function(code,language,tabnumber)
   if(language == "java" || language == "php" || language == "cpp" 
     || language == "ruby" || language == "xml" || language == "sql" || language == "alloy" || language == "nusmv")
   {
-		jQuery("#innerGeneratedCodeRow" + tabnumber).html(
-			formatOnce('<pre class="brush: {1};">{0}</pre>',generatedMarkup,language)
-		)
+		jQuery("#innerGeneratedCodeRow" + tabnumber).innerHTML = 
+		  formatOnce('<pre class="brush: {1};">{0}</pre>', generatedMarkup, language);
     SyntaxHighlighter.highlight("code");
 
 		if(tabnumber == ""){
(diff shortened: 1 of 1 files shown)


TimLethbridge:
No, this does not work when I test it. The generated code does not appear at all!

Nava2:
Damn, I was hoping the local changes would work. I'll open a new PR once I fix this up.

Nava2:
@TimLethbridge is UmpleOnline supported running locally on Windows? I can't get the compiler to spew anything other than "Couldn't read results from the Umple compiler!"

Nava2:
I did fix the error in 994e3b19a48c, I believe. But I am still wondering about running UmpleOnline on Windows.

TimLethbridge:
Well with the latest, the code appears, but the problem is not solved: We will get lowercase conversion.

As for UmpleOnline in Windows. You are the second person this week who reported this issue. On Mac I know the symptoms "Couldn't read results ..." are caused by permission problems (web server not being able to create/write files/directory in the umpleonline/ump directory

TimLethbridge:
My guess is that the lowercasing occurs a little earlier in the process of constructing the html. I have to focus on other deadlines, but will monitor this.

TimLethbridge:
The code that comes from the compiler with all files glommed together has 
```
private List&lt;D&gt; ds;
```
The lt and gt appear to be being stripped before the tabs are generated. Hence the glommed display looks OK, but the tabbed display has the lowercasing happening

vahdat-ab:
@Nava2 I have the same issue in the Windows and tried to fix it but were not successful. If you still want to have it on Windows, please use "Bash on Ubuntu on Windows". It works without any issue.

TimLethbridge:
I have solved the problem I will open a different PR on it.

TimLethbridge:
The fix is done instead in #973

PR #973, opened 2017-01-24
Fixes 970 - getting raw text instead of processed text for tabs

### LINKED ISSUE #970: Type of association reference list is lower case in UmpleOnline when clicked on "show tabs on different tabs"

TimLethbridge:
This pull request replaces that of @Nava2 

The problem was that the tab text was being obtained by querying the DOM, where the less than and greater than html codes had already been rendered. In this fix, the raw code from the compiler is used instead.

This fixes #970

### DIFF
--- a/umpleonline/scripts/umple_action.js
+++ b/umpleonline/scripts/umple_action.js
@@ -2191,7 +2191,7 @@ Action.toggleTabs = function()
   }
 }
 
-Action.generateTabsCode = function()
+Action.generateTabsCode = function(theCode)
 {
   var arrCodeFiles = [];
   var intFileCounter = 0;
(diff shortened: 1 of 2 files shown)


Nava2:
Excellent, thanks for doing this. I was working on it, but I got lost in the somewhat messy setup that was there. 

The JS for UmpleOnline could use some love at some point to modernize it and clean up some of the cruft.

PR #975, opened 2017-01-26
 this fixes the issue with required methods injecting to abstract classes

### LINKED ISSUE #758: Activate testing of UmpleOnline and User Manual examples during the build by compiling C++/Php
Work was being done by @CraigBryan to ensure that all the examples in the user manual and umpleonline are compiled in the build.

This was never activated. We really need to finish that so we can keep improving the robustness

See also #363

vahdat-ab:
This PR fixes the issue related how required methods are propagated to abstract classes. This issue was detected in code sprint in the example named `Geometric system`. I also fixed it.  

Closes #758


vahdat-ab:
@jblang94 I wonder if you can double check if the example now working as you expected?

jblang94:
@vahdat-ab the fix worked! Affected files such as Quadrilateral.java no longer throw compiler errors that indicate that they need to implement the isEqual(CurvedShape object). Could you please remove the @@@skipjavacompile directive from GeometricSystem.ump since the example now works for Java? Great job!

Issue #686, opened 2016-01-16
After/before incorrectly generates warnings when inserting into Umple generated methods for associations.


Shikib:
Consider the following code:

```
class A{
  int a;
}

class B {
  1 -- * A;

  after addA {
    test();
  }
}
```

addA exists as a method because its is generated by Umple for the association between A and B. The code that prints out warnings for unidentified methods doesn't recognize addA because it doesn't attempt to determine any of the methods for an association.

TimLethbridge:
Check if this is still an issue

vahdat-ab:
This works. However, what I noticed is that there are two methods with the name addA `addA(int aA)` and `addA(A aA)`. This injection is applied to the second one. The question is that why we don't have it for the first one? it must be justified.

PR #977, opened 2017-01-27
Fix#682 - added warning for exclusion operation

a3994288:
## Description

The original issue was already handled.

I added warning 1014: 

> 1014: 3, "http://cruise.eecs.uottawa.ca/umple/PageBeingDeveloped.html", Excluded method '{0}' cannot be found. The exclusion was ignored. ;  

This will raise when there is an exclusion method like "!getMyBirth", and getMyBirth method does not exist.

## Tests

All tests were passed and no new test was added.

### DIFF
--- a/cruise.umple/src/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/UmpleInternalParser_CodeClass.ump
@@ -3111,19 +3111,20 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
 
       Boolean matches = false;
       
-      Boolean exclusionMatch = true;
+      Boolean exclusionMatch = false;
 
       for(String methodName : methodNames) 
       { 
         if (uClass.matchOperationMethod(operation, methodName)) {
           matches = true;
+          exclusionMatch = true;
         }
         
         // issue#682
         else if(operation.charAt(0) =='!') {
           matches = true;
-          if (!uClass.matchOperationMethod(operation.substring(1), methodName)) {
-            exclusionMatch = false;
+          if (uClass.matchOperationMethod(operation.substring(1), methodName)) {
+            exclusionMatch = true;
           }
         }
       }
--- a/cruise.umple/src/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/UmpleInternalParser_CodeClass.ump
@@ -3096,7 +3096,9 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
 
   private void checkCodeInjectionValidity(Token injectToken, UmpleClass uClass) {
     ArrayList<String> methodNames = getAllMethodNames(uClass);
+    
     String[] allOperations = getOperationName(injectToken).split(",");
+    
     List<String> allParameters = getOperationsParameters(injectToken);
 
     String operationSource = getOperationSource(injectToken);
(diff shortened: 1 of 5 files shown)


vahdat-ab:
Thanks :-)
You need to write test cases for any issue fixed. 
When the compiler raises the issue, the name of methods comes with the exclamation marks. Just remove it and show the name of the method.
Another issue is that when there is no exclamation mark and the method does not exist, it shows the warning 1012 and 1014.  For example,
```
class Student
{
  before cityOfBirt { 
    if ( DEBUG) { System.out.println("accessing the name"); }
  }
}
```
output:
```
Warning 1012 on line 3 of file "test.ump":
Method 'cityOfBirt' cannot be found. Injection was ignored.
Warning 1014 on line 3 of file "test.ump":
Excluded method 'cityOfBirt' cannot be found. The exclusion was ignored.
```
The error code 1014 doesn't need to be raised. I think you can fix the main issue by investigating better how the warning code 1012 is raised and detected.

vahdat-ab:
Check the method `checkCodeInjectionValidity(Token injectToken, UmpleClass uClass)` in the file `UmpleInternalParser_CodeClass.ump`.
The issue is with this function`getOperationName(injectToken).split(",")`. It returns `!` as a part of method name. If you fix it, the whole issue is fixed.

a3994288:
@vahdat-ab  It should work now. Tests also added ( However, I am not sure if the filename of the tests I used are correct ). Please take a look. Thanks.

vahdat-ab:
I think there is sth wrong with the previous implementation.
Check this out 
```
class Student
{
  String textdata(){
    
  }
  before !textdata() { 
    if ( DEBUG) { System.out.println("accessing the name"); }
  }
}
```
Although the method is there, but it shows the error.
It also shows warning to the following example:
```
class Student
{
  String textdata(){
    
  }
  before textdata() { 
    if ( DEBUG) { System.out.println("accessing the name"); }
  }
}
```
Looks like the way methods are found and compared has been modified.

a3994288:
```
class Student
{
  String textdata( Integer i ){
  }
  
  before textdata( Integer i ){ 
    if ( DEBUG) { System.out.println("accessing the name"); }
  }
}  
```
The code above is working fine when the method has parameter, I will investigate more later.

a3994288:
@vahdat-ab When I am reading this user manual  
[http://cruise.eecs.uottawa.ca/umple/BeforeandAfterStatements.html](url)  
It says before/after statements are not available for methods for now, is that the reason the examples you provided are giving errors?

vahdat-ab:
Yes. It looks like. However, why 

> The code above is working fine when the method has parameter, I will investigate more later.

that above code is working. Is there a straightforward to fix that? if yes, please create another issue for that.

a3994288:
The above code with parameters actually does not work properly.  
For example, if you try this code(added a '!' after 'before' operation)  
```
class Student
{
  String textdata( Integer i ){
  }
  
  before !textdata( Integer i ){ 
    if ( DEBUG) { System.out.println("accessing the name"); }
  }
} 
```
It will NOT raise a error, but if generates Java code, the java code is absolutely wrong.  

Based on my understanding so far, I haven't seen a very straightforward way to fix it (I could be wrong!).  
However, I believe when 'before/after' operations support methods, all of those should be fixed automatically.  

Is there anyone working on developing the feature or you want to me to create an issue for that? Or, you want to consider that as part of issue#682 so I will just keep investigating on it?

a3994288:
@vahdat-ab I have not found an existing issue for this one (issue #686 is somewhat related), can I create an issue for it? However, in the user manual (http://cruise.eecs.uottawa.ca/umple/BeforeandAfterStatements.html), it says 

> The ability to do this with methods will be available soon  

Does that mean someone is working on it?  

BTW, this PR can be merged now if we want a separate issue.

vahdat-ab:
No one is working one it. Please create proper issues.

PR #980, opened 2017-01-28
Fix #964

### LINKED ISSUE #964: Umple's PHP method generation fails to create valid method parameters
## Brief Description

When Umple generates PHP code, it fails to adjust method parameters so that they are valid PHP method parameters. To be more exact, the generated PHP code fails to prepend the parameter name with the "$" symbol. As a result, parsing errors occur when PHP attempts to process the generated code.

This was discovered when attempting to fix the "GeometricSystem.ump" UmpleOnline example for PHP compilation in issue #758.
 
## Minimum Steps to Reproduce

Consider the following Umple code
```
generate Php;

class GeometricObject {

}

class Canvas {
        void paint(GeometricObject g){/*implementation*/}
        void update(GeometricObject g){/*implementation*/}
        void example(boolean b) {}
}
```

## Expected Result
In the generated Canvas.php, for the paint, update, and example, we would expect the following:
```
public function paint(GeometricObject $g)
{
       /*implementation*/
}

public function update(GeometricObject $g)
{
       /*implementation*/
}

public function example($b)
{
       
}
```

## Actual Result

```
public function paint(GeometricObject g)
{
       /*implementation*/
}

public function update(GeometricObject g)
{
       /*implementation*/
}

public function example(boolean b)
{
       
}
```

If a method's parameter is a *primitive* type in the Umple code (ex. void example(boolean b)), the corresponding generated method in the PHP code should *not* include the type (ex. public function example($b)) . Otherwise, we keep the method parameter's type in the generated PHP method.

For both cases, the parameter name must be prepended with the "$" symbol.

Primitive types include strings, booleans, integers, floats, doubles, and booleans.

nwam:
## Description
This fixes issue #964. 

## Tests
Several test templates have been slightly modified such that the method parameters are actual valid PHP method parameters. An additional test has been created to cover every type of method parameter.

Closes #964

### DIFF
--- a/cruise.umple/test/cruise/umple/implementation/php/BasicPostcondition1.php.txt
+++ b/cruise.umple/test/cruise/umple/implementation/php/BasicPostcondition1.php.txt
@@ -1,6 +1,6 @@
 <?php
 /*PLEASE DO NOT EDIT THIS CODE*/
-/*This code was generated using the UMPLE 1.25.0-963d2bd modeling language!*/
+/*This code was generated using the UMPLE @UMPLE_VERSION@ modeling language!*/
 
 class Client
 {
--- a/cruise.umple/test/cruise/umple/implementation/php/BasicPostcondition1.php.txt
+++ b/cruise.umple/test/cruise/umple/implementation/php/BasicPostcondition1.php.txt
@@ -1,6 +1,6 @@
 <?php
 /*PLEASE DO NOT EDIT THIS CODE*/
-/*This code was generated using the UMPLE @UMPLE_VERSION@ modeling language!*/
+/*This code was generated using the UMPLE 1.25.0-963d2bd modeling language!*/
 
 class Client
 {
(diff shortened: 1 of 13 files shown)


cruise.umple/test/cruise/umple/implementation/php/BasicPostcondition1.php.txt
```diff
@@ -1,6 +1,6 @@
 <?php
 /*PLEASE DO NOT EDIT THIS CODE*/
-/*This code was generated using the UMPLE @UMPLE_VERSION@ modeling language!*/
+/*This code was generated using the UMPLE 1.25.0-963d2bd modeling language!*/
```
vahdat-ab: You don't need to specify the version. Please follow other test cases. We never put the version in our test cases. 
This is applied to all others you have changed.

nwam: I generated this code via my build at `/dist/umple-*.jar`. Is there another way in which I should be creating the `*.txt` test files? Or should I just change that comment in the generated files?

vahdat-ab: If you keep the following line, it should works.
This is what you can see in all test cases.
```
/*This code was generated using the UMPLE @UMPLE_VERSION@ modeling language!*/
```
The rest should be adjusted.

vahdat-ab:
Thanks.

PR #982, opened 2017-01-29
Fixing issue 961. Check to make sure a subclass does not have an asso…

### LINKED ISSUE #961: Associations with the same role name but two different classes in superclasses and subclasses should produce an error.
## Brief Description

When a superclass and subclass both have an association with the same role name, but the association is to different classes, an error should be produced.

## Minimum Steps to Reproduce

```java
class A
{
  0..1 -> * C assoc;
}

class B {
  isA A;
  0..1 -> * D assoc;
}

class C
{
}

class D
{
}
```

## Expected Result

Expected results: ERROR

## Actual Result

Invalid code is generated. Included below is a small subset to illustrate the issue:
```java
public class A
{

  //------------------------
  // MEMBER VARIABLES
  //------------------------

  //A Associations
  private List<C> assoc;

  //------------------------
  // CONSTRUCTOR
  //------------------------

  public A()
  {
    assoc = new ArrayList<C>();
  }

 //------------------------
  // INTERFACE
  //------------------------

  public C getAssoc(int index)
  {
    C aAssoc = assoc.get(index);
    return aAssoc;
  }
}

public class B extends A
{

  //------------------------
  // MEMBER VARIABLES
  //------------------------

  //B Associations
  private List<D> assoc;

  //------------------------
  // CONSTRUCTOR
  //------------------------

  public B()
  {
    super();
    assoc = new ArrayList<D>();
  }

  //------------------------
  // INTERFACE
  //------------------------

  public D getAssoc(int index)
  {
    D aAssoc = assoc.get(index);
    return aAssoc;
  }
}
```

katcavers:
…ciation that is not a specialization with the same role name as its superclass.

## Description

This PR provides a fix for the case where a subclass declares an association that is named the same as an association in one of its superclasses, but is not a specialization.
Added a new error message (180).

## Tests

Added 3 tests to UmpleParserTest.java to test the cases this error should be thrown.

Closes #961

### DIFF
--- a/build/reference/9180DuplicateAssociationNameSubclass.txt
+++ b/build/reference/9180DuplicateAssociationNameSubclass.txt
@@ -1,0 +1,17 @@
+E180 Duplicate Association Name Class Hierarchy
+Errors and Warnings
+noreferences
+
+@@description
+
+<h2>Umple semantic error reported when a subclass has an association that is not a specialization named the same as a superclass association</h2>
+
+<p>A subclass can not have an association with the same role name as an association of its superclass, unless it is a valid specialization. 
+</p>
+
+
+@@example
+@@source manualexamples/E180DuplicateRoleNameSubclass.ump
+@@endexample
+
+
--- a/cruise.umple/src/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/UmpleInternalParser_CodeClass.ump
@@ -1974,6 +1974,8 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
           break;
         }
         visitedSupers.add(superClass.getName());
+        
+        ArrayList<String> duplicateNames = new ArrayList<String>();
 
         for(Association assoc : uClass.getAssociations())
         {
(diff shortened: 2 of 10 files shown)


vahdat-ab:
Thanks @katcavers
Please consider the following items.
If the file `E180DuplicateRoleNameSubclass.ump` is compiled, it outputs two times the error 180.
Your solution cannot detect the following example: 
```
class A{
  0..1 -> * B;
}
class B{
   isA A;
   0..1 -> * D bs;
}
class D{}
```

vahdat-ab:
In the following example, it shows two times the same error message. It doesn't happen in other cases
```
class A{
  0..1 -> * B;
}
class B{
   isA A;
}
class D{
  isA B;
  0..1 -> * D bs;
}
class C{}
```

The following example just raises one error, which is correct.
```
class A{
  0..1 -> * B;
}
class B{
   isA A;
}
class D{
  isA B;
  0..1 -> * C bs;
}
class C{}
```

vahdat-ab:
Thanks @katcavers

PR #994, opened 2017-02-13
Fix #483 issue for Umple

### LINKED ISSUE #483: Having execution time for compile and code generation in UmpleOnline
Originally reported on Google Code with ID 585
Owned by @mgarzon

---

It would be good to see execution time related to code generation and compile time on
UmpleOnline. It will shows real-compile time without dependency to internet speed (which
some time users think that umple works slow while it's because of their machine). 
The following items, for example, can be shown in UmpleOnline:
1) Parsing time,
2) Analysis time, 
3) Code generation, and
4) total time.
We also need to show config of our server in order to have valid interpretation for
the times.

---

Reported by @vahdat-ab on 2014-06-06 15:43:14

Jaorji:
The test failed.The error message is:
cinst ant 2>&1 | Out-Null
cinst php --version=5.6.29 2>&1 | Out-Null
Command exited with code 1

### DIFF
--- a/cruise.umple/src/Main_Code.ump
+++ b/cruise.umple/src/Main_Code.ump
@@ -22,6 +22,7 @@ class UmpleConsoleConfig {
 
   defaulted boolean version = false;
   defaulted boolean help = false;
+  defaulted boolean performance = false;
   
   internal Optional<String> generate   = Optional.empty();
   defaulted boolean override           = false;
@@ -38,6 +39,7 @@ class UmpleConsoleConfig {
 
     this.version    = optSet.has("v");
     this.help       = optSet.has("help");
+    this.performance = optSet.has("performance");
     this.importFile = Optional.ofNullable((String)optSet.valueOf("import"));
     this.generate   = Optional.ofNullable((String)optSet.valueOf("generate"));
     this.override   = optSet.has("override");
@@ -116,11 +118,16 @@ class UmpleConsoleMain
   depend java.util.Arrays;
   depend joptsimple.*;
   depend java.io.*;
+  depend java.text.DecimalFormat;
 
   UmpleConsoleConfig cfg;
 
   static OptionParser optparser = initializeOptionParser();
 
+  //startTime repersents model run start time while endTime represents end time
+  Double startTime = 0.0;
+  Double endTime = 0.0;
+  
   /**
    * Create a main console. If the args are invalid, an IllegalStateException is thrown. 
    * @throws IllegalStateException If invalid arguments are passed and joptsimple fails to parse. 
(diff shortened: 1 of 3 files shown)


cruise.umple/src/Main_Code.ump
```diff
@@ -116,10 +118,16 @@ class UmpleConsoleMain
...
+  depend java.text.DecimalFormat;
 
   UmpleConsoleConfig cfg;
 
   static OptionParser optparser = initializeOptionParser();
+  
+  //t0 repersents model run start time while t1 represents end time
+  Double t0 = 0.0;
+  Double t1 = 0.0;
+
```
vahdat-ab: I would suggest changing names to ones which are more meaningful.
For example, `startTime` and `endTime`

cruise.umple/src/Umple_Code.ump
```diff
@@ -18,14 +18,23 @@ namespace cruise.umple.compiler;
 class UmpleModel
 {
 
+  //t11 represents Parsing end time
+  //gTime1 represents code generation start time
+  //gTime2 represents code generation end time
+  Double t11 = 0.0;
+  Double gTime1 = 0.0;
+  Double gTime2 = 0.0;
```
vahdat-ab: again use relevant names for variables. You can also use proper modifiers for those variables.  for example, `tParsingTime` for the parsing time.

cruise.umple/src/Umple_Code.ump
```diff
@@ -315,7 +331,18 @@ class UmpleModel
...
+  {
+    return hashMap;
+  }
+  
+    public String getParsingTime()
+  {
+    String parsingT = String.valueOf(t11);
+    return parsingT;
+  }
+
```
vahdat-ab: You don't need to define this method. If you give a proper name to the variable, Umple will generate this method automatically.

cruise.umple/src/Main_Code.ump
```diff
@@ -292,6 +318,7 @@ class UmpleConsoleMain
     
     String languages = String.join(",", UmpleModel.validLanguages );
     optparser.acceptsAll(Arrays.asList("version", "v"), "Print out the current Umple version number");
+      optparser.acceptsAll(Arrays.asList("performance", "p"), "Indicate time to generate files");
```
vahdat-ab: change this `Indicate time to generate files` with the following one:
`Indicate time taken to parse and generate code`

cruise.umple/src/Umple_Code.ump
```diff
@@ -18,6 +18,15 @@ namespace cruise.umple.compiler;
 class UmpleModel
 {
 
+    //ParsingTime represents Parsing end time
+    //gTime1 represents code generation start time
+    //gTime2 represents code generation end time
+    Double parsingTime = 0.0;
+    Double gTime1 = 0.0;
```
vahdat-ab: `gTime1` and `gTime2` must be indicated as `internal`. The reason is that there is no need to them to be available through an instance of `UmpleModel`

vahdat-ab:
@Jaorji  good job. Thanks.

Issue #962, opened 2017-01-21
Bringing Umple's APIs to the metamodel and runtime


vahdat-ab:
Umple generates a hand full of useful APIs for attributes and association. Currently, those methods are generated by UmpleTL. It means they are not available in the runtime model of Umple systems. This limits us regarding many analysis we can do at the runtime and causes to have duplication in UmpleTL's template files. Furthermore,  This will be a great base for the future extensions such as using Language Server Protocol. 
One of the concerns is how to deal with the executable body of these methods. I think we can create those APIs (their signature at the modeling level) and then tag them as auto-generated. Then code generated will know how to deal with that.
This idea can be extended in an advance form to state machines.

More info about those APIs can be found [here](http://cruise.eecs.uottawa.ca/umple/APISummary.html)

vahdat-ab:
Another concern might be the size of the model at the runtime. I think this can be dealt later with some optimization. It might not be even an issue with the current advance in technology.

TimLethbridge:
Sounds interesting, but I think explaining what you mean by example would be important. Give an example trivial umple model with two classes, an association and an attribute, and describe the effect of what you would like to see

vahdat-ab:
This is in the direction relationship with issue #462

vahdat-ab:
Consider the following example, 
```
class A{
  Integer x;
  0..1 -- * B;
}
class B{ 
}
```
Umple generates API for the attribute `x` and also the association `0..1 -- * B;` The list of APIs for different elements are listed [here](http://cruise.eecs.uottawa.ca/umple/APISummary.html)
These APIs are generated by the different code generators (Java, PHP, etc). It means if a modeling element wants to check a method called `void getX()` in the class `A`, the list of methods for the class `A` will not show it. In these cases, the developer of that element starts to hack the code. The person implements another function which reacts like the one exists in the code generators and produce the list of auto-generated methods (sometimes, it's just a manual list). This brings lots of boiling code. Furthermore, it's hard to update to date all those manual codes with their own hard-coded names. This becomes much worse for cases that we do lots of post checking to make sure the model is valid. 
For example, to check if there is no conflict between used-defined methods and auto-generated methods.
By having those methods at the metamodel level, we guarantee that the regular methods such as `getAllMethods` will always show the right number of methods. Then, a normal iteration on this list can satisfy many post-checkings.
Another example,
```
class A{
  Integer x;
}
trait T{
  Integer getX();
  void setX();
  void pMethod(){
  //***
  }
}
[...]

PR #1007, opened 2017-02-27
Initial commit. Add set and get methods for applicable attributes.

katcavers:
## Description

This PR is the first step in bringing generated API methods to the model level for issue #962 . This commit includes changes to add a get and set method for applicable attributes at the model level. Currently, these additional methods are skipped over in the code generation stage, so generation of API methods remains unchanged.

I tried to find all the generators and add the necessary code to skip generation of the get and set methods. If any were missed, please let me know.

## Tests

Added a test in UmpleParserTest called autoGeneratedAPIMethods.


vahdat-ab:
@katcavers The build is failing. Please fix them and request your PR again.

katcavers:
@vahdat-ab It is fixed now. I had forgotten to commit one file, oops!

cruise.umple/test/cruise/umple/compiler/UmpleParserTest.java
```diff
@@ -2560,6 +2560,70 @@ public void methodNameConflictUsingDefaulted()
     assertFailedParse("045_phpDefaultedNameConflict.ump", 48);
     assertFailedParse("045_rubyDefaultedNameConflict.ump", 48);
   }
+  
+  @Test
+  public void autoGeneratedAPIMethods()
```
vahdat-ab: It is better to keep test cases simple.  Each test case should be dealing with a specific case. It helps later to explore the test case easier.  One test case for every stereotype. You can keep the source model the same for all test cases.

vahdat-ab (changes requested):

vahdat-ab:
Great. 
You also need to take care of errors like 1009. The way they have been implemented must be changed because now we have auto-generated methods at the model.
you can get more details about errors related to attributes in `en.error`.
you can create another issue for the metioned case and I'll merge this PR, or you can update the PR and then I'll merge. I prefer the second one :-)

cruise.umple/test/cruise/umple/compiler/UmpleParserTest.java
katcavers: Is there a way to only have to call assertSimpleParse once if I separate them all into separate tests? Or do I have to call it now for each test?

vahdat-ab:
@katcavers the build is failing

katcavers:
@vahdat-ab yes, I know, I didn't mean to push the last commit. I commented on your code review, but perhaps you didn't see it? Just in case, here is the question again:
Is there a way to only have to call assertSimpleParse once if I separate them all into separate tests? Or do I have to call it now for each test?
Thanks!

vahdat-ab:
Yes, you can do it inside `public void setUp()`. However, I would suggest breaking your file to several files in the direct connection with the tests you do. In that case, you parse things (use assertSimpleParse ) related to each test case.

katcavers:
Okay, so should I (1) keep the tests where they are in UmpleParserTest and create separate .ump files to parse for each test case, or (2) create a new test file for these tests and use the setup method to parse the .ump file once, or (3) create a new test file for these tests and use separate .ump files for each of them?

Which would be best? :)

vahdat-ab:
Option 1

cruise.umple/src/Generator_CodeGvClassDiagram.ump
```diff
@@ -129,6 +129,9 @@ class GvClassDiagramGenerator
     if (hasSuboption("showmethods")) {
       boolean isFirstMethod = true;
       for (Method uMethod : inUmpleClassifier.getMethods()) {
+      	if (uMethod.getSource() == Method.Source.fAutoAPI)
+      	  continue;
+
```
vahdat-ab, line 134: I wonder why you wanted to escape these methods here?
Actually, it's a nice way to show what them in the model. By default, we don't shows any methods, but when they are selected they should be shown

vahdat-ab:
@TimLethbridge I wonder if you wanted to have Umple auto-generated methods to appear in generators like Json, scxml, etc. To me, it makes sense to have them there as well.

cruise.umple/src/Generator_CodeGvClassDiagram.ump
vahdat-ab: as we discussed online, disregard my point.

vahdat-ab:
I wonder if you have done sth related to this PR or not? If no, so I want to merge this.
Probably you can add remove methods to this PR so they are all merged together.

katcavers:
Hey Vahdat, yes I had thought that after our meeting a week ago it was going to be pulled. But you are right, adding the remove methods to this PR makes sense. Hopefully I have that ready to go by the end of today.

katcavers:
Hi Vahdat, I saw one of the checks failed, and looked in the logs to find this:
    [junit] Running cruise.statemachine.test.DoActivityTest
    [junit] Tests run: 8, Failures: 0, Errors: 1, Time elapsed: 36.028 sec

I believe this is why it failed? I'm not sure how I can debug the problem? All tests passed locally
@vahdat-ab

katcavers:
I'm also really confused as to why this commit could cause this to start failing...the only difference is a new test and two methods which are actually never called anywhere in the code yet. The previous commit passed both travis-ci builds.

vahdat-ab:
That's ok. The test fails because of pool/queue state machines. This is a random failure.

cruise.umple/src/Umple_Code.ump
```diff
@@ -1966,6 +1966,41 @@ class UmpleClass
...
+  	String firstLetter = attributeName.substring(0,1).toUpperCase();
+  	String rest = "";
+  	if (attributeName.length() > 1)
+  	{
+  		rest = attributeName.substring(1);
+  	}
+  	String getName = "get" + firstLetter + rest;
+  	String setName = "set" + firstLetter + rest;
+  	
+  	ArrayList<Method> methodsCopy = new ArrayList<Method>(this.getMethods());
```
vahdat-ab: It is better to define the variable more general.
`List<Method> methodsCopy = new ArrayList<Method>(this.getMethods());`

vahdat-ab (commented):
I have a question:
do yo deal with constructor at the modeling level or not?
if yes then when an attribute is removed its constructor needs to be modified.

katcavers:
Yes, I am working on the constructor now. I am almost done, but I had not considered this. I will look into it tomorrow. Thank you!

cruise.umple/src/Umple_Code.ump
katcavers: Okay I will make this change.

PR #1013, opened 2017-03-11
jQuery update - 2.2.4

a3994288:
## Description

The older version of jQuery does not work well with JointJS. Update it to a newer version (2.2.4).

## Tests

No test was added, and all tests passed.


TimLethbridge:
OK Good idea. I will merge this; then it will appear in Umpleonline/test. We will see if all is OK there, and also on our local machines.

Issue #935, opened 2016-12-16
the sequence of exit actions are not executed correctly


vahdat-ab:
Here is somehow a complete example shows how exit must be treated.
@TimLethbridge  @opeyemiAdesina  I wonder if you guys can check if the semantics related to execution order is correct or not?
```
class X{
  sm{
    on{
      exit /{exit_on_execute();}
      e1-> off;
      e2-> on;
      s1{
        exit /{exit_s1_execute();}
        e3-> s2;
        e4-> s1;
        e5-> on;
        e6-> off;
        m1{
          exit /{exit_m1_execute();}
          e7-> m2;
          e8-> m1;
          e9-> s1;
          e10->s2;
          e11->on;
          e12->off;
        } 
        m2{} 
      }
      s2{}
    }
    off{}
  }
  void exit_on_execute(){System.out.println("exited on");}
  void exit_s1_execute(){System.out.println("exited s1");}
  void exit_m1_execute(){System.out.println("exited m1");}
}
```
The following part shows the execution oders of exits for each event. It shows which is supported currently by Umple.

//e1-> off;  : exit_on_execute(); //correct
//e2-> on;   : exit_on_execute(); //correct
//e3-> s2;   : exit_s1_execute(); //correct
//e4-> s1;   : exit_s1_execute(); //correct
//e5-> on;   : exit_s1_execute(); -> exit_on_execute(); //Not correct
//e6-> off;  : exit_s1_execute(); -> exit_on_execute(); //Not correct
//e7-> m2;   : exit_m1_execute(); //correct
//e8-> m1;   : exit_m1_execute(); //correct
//e9-> s1;   : exit_m1_execute(); ->exit_s1_execute(); //Not correct
//e10->s2;   : exit_m1_execute(); ->exit_s1_execute(); //Not correct
//e11->on;   : exit_m1_execute(); -> exit_s1_execute(); -> exit_on_execute(); //Not correct
//e12->off;  : exit_m1_execute(); -> exit_s1_execute(); -> exit_on_execute(); //Not correct

vahdat-ab:
@opeyemiAdesina 
The execution order specified above is logical. it's just for explanation. for e1 it should first `exit_on_execute();`  and then if there is an entry it must be executed. The generated code doesn't have issues with entry actions. That's the reason I didn't specify it there.

vahdat-ab:
My comments regarding `//correct` and `//Not correct` meant to say that Umple supports it or not.
[...]

PR #1014, opened 2017-03-15
Resolve issue 935

jblang94:
## Description
This PR addresses [issue 935](https://github.com/umple/umple/issues/935). The first part of the fix is in prepareNestedStatesFor in "Generator_CodeJava.ump", and prepareNestedStateMachine in "GeneratorHelper_CodeStateMachine.ump". This is where the exit actions required for the modular exit method approach are injected. The second part of the fix is in "state_machine_Event". Two helper functions were also added to "StateMachine_Code.ump".

For non concurrent state machines, this is where the call to the super state machine is made to initiate the correct execution order of exit actions, and to set the affected state machines to their Null state.   

For concurrent state machines, additional processing occurs to check for if the transition is taking place between states in a concurrent state machine, or if it is an and-cross transition. If neither of these conditions are satisfied, a call to the super state machine is made to handle the exit actions.

#### Created Tests
checkExternalTransition_[specificCase].ump
checkExternalTransition_[specificCase].java.txt

#### Updated Tests
GeneratorHelperTest.java - prepare_NestedStateMachine
JavaGeneratorTest.java - prepare_postpare_nestedStateMachine
JavaStateMachineTemplateTest.java - multiple tests
UmpleSelfGeneratorTest.java - StateMachines
NestedStateMachineTest.java - ExitFromNestedStates
HistoryStatesTest.java - AutoTransitionToDeepHistory, SimpleDeepHistoryState

#### Ignored Tests
PhpGeneratorTest.java - prepare_postpare_nestedStateMachine
PhpStateMachineTemplateTest.java - multiple tests
AllTests.php - statemachine test

Tracer tests were detecting errors, however, they extend TracerAttributesTest which is an ignored test class:
JavaConsoleTracerTest.java
JavaFileTracerTest.java
JavaStringTracerTest.java

### DIFF
--- a/UmpleToJava/UmpleTLTemplates/state_machine_Event.ump
+++ b/UmpleToJava/UmpleTLTemplates/state_machine_Event.ump
@@ -124,35 +124,29 @@ class UmpleToJava {
         if (state != nullState && state.getHasExitAction()) 
         {
           // Concurrent state machines require additional handling
-          if (state.isInConcurrentSm()) {
+          State fromConcurrentParentState = state.getConcurrentParentState();
+          if (fromConcurrentParentState != null) {
             StateMachine concurrentSmToExit = null;
             
-            if (state.getStateMachine().equals(superStateMachine) || nextState.getStateMachine().equals(superStateMachine)) {
+            if (nextState.getStateMachine().equals(superStateMachine)) {
               concurrentSmToExit = superStateMachine;
             } else if (state.getStateMachine().equals(nextState.getStateMachine())) {
               concurrentSmToExit = state.getStateMachine();
-            } else if (nextState.isInConcurrentSm()){ 
-              State fromConcurrentParent = state.getStateMachine().getParentState();
-              State nextConcurrentParent = nextState.getStateMachine().getParentState();
-              
-              while (!fromConcurrentParent.getIsConcurrent()) {
-                fromConcurrentParent = fromConcurrentParent.getStateMachine().getParentState();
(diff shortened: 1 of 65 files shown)


vahdat-ab:
When an event happens, it is modulated inside its superstate and so it doesn't need to call exit actions of all of its superstates. The following shows the cases that are wrong.

- the event `e10` should not run `exit_on_execute()` because the event happens inside the state `on`.
- the event `e9` should not run `exit_on_execute()` because the event happens inside the state `on`.
- the event `e8` should not run `exit_on_execute()` and `exit_s1_execute()` because the event happens inside the state `s1`.
- the event `e7` should not run `exit_on_execute()` and `exit_s1_execute()` because the event happens inside the state `s1`.
- the event `e4` should not run `exit_on_execute()` because the event happens inside the state `on`.
- the event `e3` should not run `exit_on_execute()` because the event happens inside the state `on`.

jblang94:
@vahdat-ab thanks for the feedback! All fixed I believe

vahdat-ab:
@jblang94 Great. Now, it's perfect. Thanks. However, let's give others to run a few more examples.

I'm asking you guys to run a random example to validate more what @jblang94 has done. We have to change the way exit actions were handled and use a much easy way to deal with this issue.
 
@opeyemiAdesina Please run an example based on your expectation and let us know about the results.
@TimLethbridge Please run an example based on your expectation and let us know about the results.

Issue #1008, opened 2017-03-01
support enumeration types in Umple


vahdat-ab:
Enumeration types are modeling elements and part of many class diagrams,
As long as I 'm aware there is no such implementation for it. Even we cannot see it in the graphical representation of my class diagram.
I'm also aware that someone might want to define an enumeration like below in Umple
```
class A{
  status{married, single, divorced}
}
```
However, I argue that it's not correct.
Generated code:
```java
public enum Status { married, single, divorced }
private Status status;
```
Umple detects that one as a state machine (you can see that a state machine diagram is generated for it). A variable with the same name is generated in the java code and then the same name becomes camelcase and is considered as the name of enumeration. 
These issues must be fixed. The type must be exactly what designer defines. If designers want to use enumerations, they can use it as normal types for attributes.
 Plus, The graphical representation (class diagram) should be improved to present it.

TimLethbridge:
We decided long, long ago that logically a state machine is an attribute with fixed values (each of its states), so logically an enumeration is too. The key difference is that a state machine changes values by way of events.  There are examples in existing systems so we can't change the generated code now. 

If you want to do something different, please explain the generated code you would prefer. Maybe we can add a stereotype.

vahdat-ab:
@TimLethbridge  A question: how we can satisfy this simple general model in Umple based on that philosophy. 
```
enum Course { eng, csi}
class A{
    Course course1;
}

class B{
    Course course2;
}
```
It doesn't matter where we define an enumeration. My issue is how to deal with reusing it.
Furthermore,  sometimes I just need to define it without creating an attribute, but it looks like impossible in Umple. It is forced to have an attribute with any enumeration.

TimLethbridge:
Eight then. So use the keyword enum as a stereotype to generate this. Leave the existing as it is.

jblang94:
@vahdat-ab

I've added the enumeration definition to **umple_classes.grammar** as follows:
`enumerationDefinition : enum [name] { [enumValue] ( , [enumValue] )* }`
and I've ` | [[enumerationDefinition]]` onto `classContent` . 

However, the parser is still parsing enumerations such as 
`enum status { FullTime, PartTime, MidTime }` as `malformedStatemachine2`. 

The rule for `malformedStatemachine2` is defined in **umple_state_machines.grammar** as follows:
```
[...]

PR #1040, opened 2017-04-07
Resolve issue 1008

jblang94:
# Description
This PR addresses issue #1008. Umple now supports enumerations. A user can define enumerations inside or outside of classes. Enumerations defined outside of classes are shared between classes. The `enumerationDefinition` rule is in umple_classes.grammar. A new `UmpleEnumeration` type has been created as well.

An "Enumerations" section is now in the Umple User Manual.

# New Tests
### UmpleParserTest
- parseUmpleEnumerationDefinedInClass
- parseUmpleEnumerationDefinedAtTopLevel
- parseUmpleEnumerationDefinedAtTopLevelAndInClass

### JavaClassTemplateTest
- TestUmpleEnumerations_*


vahdat-ab:
Cool. Thanks.
Please consider the following example:
```
enum Month {january, february, March}
enum Fruit {apple, orange}
class X1{
  Month month;
}
class X2{
   Fruit t;
}
```
When the code is generated, I see that both enumeration types are defined in both classes. However, this should not happen because class `X1` just needs the enumeration Month and class `X2` just needs the enumeration Fruit.
Another example is this
```

enum Month {january, february, March}
enum Fruit {apple, orange}
class X{}
class X2{}
```
The generator still puts the enumeration in these classes.

> The solution

Simply check if the enumeration has been used for attribute types, parameters of methods, or parameters of events of state machines. If it's been used once, then put the definition of enumeration inside the class.

For sure, this is going to be valid just for model-level enumerations

jblang94:
@vahdat-ab Thanks for the feedback! Updated my solution

vahdat-ab:
Good, thanks.
You also need to make sure the names are unique. Please consider the following example:
```
enum Month {x,y,z}
enum Month {o,p,q}
class A{
  Month m;
  Month p;
}
```
The Umple compiler easily passes this. It must be detected and a proper error must be raised.
Another example
```
enum Month {x,y,z}
class A{
  enum Month {o,p,q}
  Month m;
  Month p;
}
```

In this case, there is no error because the local version has a priority over the model-level enumeration. You need to make sure you don't generate the definition for `enum Month {x,y,z}` in class A.
Make sure you also document the error.

jblang94:
@vahdat-ab I also added detection for duplicate enumerations within an Umple class

vahdat-ab:
Great.
Just a note: when you resolve any kind of problem, you need to consider all possible cases:
Check this example:
```
enum X{x}
enum Y{y}
class X{
  X t;
}
```
Now, which type does attribute `t` have? Indeed, the name of model-level enumerations must be unique in the model. The internal enumerations do not need to have this restriction. They cannot have the same name as their classes have.
This example is valid:
```
class X{
  enum Y{y}
  Y t; // Y is not a class here
}
class Y{
}
```
This example is invalid:
```
class X{
  enum X{x}
  X t;
}

```

One more case that can cause conflict. Consider the following example: 
```
class X{
  enum Y{y}
  0..1 -- *Y;
}
class Y{
}
```
Now, `x` has an association with class `Y` or enumeration `Y`!!!. It is enumeration `Y1`, but we cannot have a bidirectional association between a class and enumeration. Indeed, when you generate the code, you will not see a compilation error because the definition of enumeration has priority or the definition of class `Y`. Class `X` will have many enumerations of `Y` and class `Y` will have `0..1` of class `X`. This is wrong because we wanted to make an association between class `X` and class `Y`. You need to detect this case and raise a proper error.
However, this example is valid:
```
class X{
  enum Y{y}
  0..1 -> * Y;
}
class Y{
}
```
In this case, because the association is directional, there is no issue. The class will make an association between class `X` and enumeration `Y`. In this case, don't make an error and just warn the user about the ambiguity and the real output (which is no association between class `X` and class `Y`, but one between class `X` and enumeration `Y`). Make sure the generated code is correct for this cases.
NOTE: enumerations and interfaces somehow have the same restriction while they are used in associations.

jblang94:
Closed this PR because the solution still needs to be refined. All feedback received has been moved to #1008

PR #1050, opened 2017-05-02
Issue 600 - JointJS

a3994288:
## Description

This does not resolve the issue 600 completely but it provides a lot improvement
- Editable JointJS object added
- JointJs for state machine (from umple code to graph)

## Tests

no test has been added.


a3994288:
I fixed most of the serious bugs that you mentioned. However, there are still some left (for example: zoom feature).  @vahdat-ab  
I cleaned up the code and added the comments. JointJS is generally working and I think other bugs can be fixed as separate issues and now it is a good time to submit a PR. I will document the unfixed bugs/features in issue600 later today.

vahdat-ab:
Thanks for the PR.
I see a major issue with paper (you used in jointJS).
It's not usable in many ways. I cannot move the paper to arrange big state machines. The state machines are not drawn automatically in the middle of paper so they are invisible partially. You can check them by just explore examples we have in Umpleonline. 
Without these simple usability features, still, we cannot move to Joint.
Could you please check all examples and make sure the way to put elements on the paper is correct. If it's big so we should be able to got to right or left to see them. I could not make it. maybe it's there and I don't know.

vahdat-ab:
I just checked and it happens to class diagrams as well. For example, `Access Control`

a3994288:
I just tested and I saw the problem. The object (especially the link) is touching the left edge of the paper, which make the edge harder to see. I will fix it.

a3994288:
I made some changes.

This time, I checked all Class Diagram using JointJS, they are all working fine except 'Geometric system' which does not work with Editable Class either. (But works with GraphViz Class)

For state machines, now the objects are centered, and the recursively nested state machines are also displayed. 

However, I am still having problems with the layout, if the state machine is big, the auto-layout is not working properly, I tried different approaches but still could not fix it.

"Home Heating System", "Security Light", "Traffic Light B" are not working because the JSON received from the server are not valid. But this can be fixed as a separate issue.

PR #1055, opened 2017-05-30
this commit adds the first version of description about traits into U…

vahdat-ab:
This adds the first version of description about traits into user manual.


TimLethbridge:
@vahdat-ab Great start on traits for the user manual. It is very timely since I will be lecturing about traits on Thursday and would love to be able to point to relevant pages and code examples in the user manual.

However, could you please make the following adjustments before merging it:

1. Please put traits after 'Methods' in the user manual, not before: Reason: Logically, people will want to learn first about methods, because they are used in traits. That requires a minor adjustment to order.group

2. In all your examples, could you please elaborate slightly on the comment. For example, rather than just saying 'Example 1', say 'Example 1: using template parameters in traits' (or whatever is relevant). Reason: This will ensure that when they are loaded in Umpleonline, or otherwise observed out of context, that people know what is special about the code.

3. The titles (and generated URLs) of many of your pages will end up being too generic. For example
in build/reference/0607traits_state_machines.txt, you have on the first line as page title
Basic Definitions
That will end up generating a URL BasicDefinitions.html
The problem with that is that if someone wants to cross reference it, or looks at the URL, it will seem as though the page is talking about Umple basic definitions in general.

Can you please change this to 
Traits
which will continue to generate Trait.html
(another possibility would be to use 'Basic Definition of Traits'. The problem is your pull request will delete the generated Traits.html page, and I have already pointed people to this URL, and there is a link to in at least one place in the manual, so Google will have indexed it).

Other cases:

Clients -> Clients of Traits
Attributes -> Attributes in Traits
Template Parameters -> Trait Template Parameters
Associations -> Associations in Traits
State Machines -> State Machines in Traits
(Required Interfaces and others I have not mentioned are OK as they are specific enough)

4. The page State Machine Operators is too long. Can you split it into 3 parts:
at points indicated below (I have copied text from your existing titles). The first part of the page would have the material up to:
  Removing/keeping a state machine
     - start a new page here called Selecting Operators
and up to just before the existing heading
  Extending a state by adding a state machine to it
    - starts new page here called Extending a State
Obviously the correct examples would be put on the relevant pages.

5.  You have specified the same syntax elements on every one of the pages. Can you vary this by putting the relevant syntax elements on the relevant pages?

6. If possible, could you add a few hyperlinks in some of the pages so people can jump around more easily (this is lower priority and could be done later; hyperlinks would point to the generated name, which is composed from the title, with spaces removed ... see existing examples)

The internal details (writing) is quite good. I may do some light textual editing later, but if you do the above 5 (or 6) changes you can then merge the PR without further input from me.

vahdat-ab:
@TimLethbridge Thanks for the review.
Item 1 is done.
Item 2 is done.
Item 3 is done. I changed the title to `Traits` to keep links valid. I also provided unique titles for all pages.
item 4 is done.
item 6 is done.

**Item 5**: I could not get the point. Could you please explain more?

TimLethbridge:
By item 5 I meant this:

For @@syntax in each of your pages, you have specified

[[traitDefinition]] [[traitContent]]

That will expand both of on every page. I think you need these two on the first page, but the other pages should not repeat them. It is redundant, from the readers perspective. However on the Trait Template parameters page, it might be reasonable to specify

[[traitParameters]] [[traitFullParameters]] [[traitParametersInterface]] [[gTemplateParameter]] [[AllInclusionExclusionAlias]]
... and so on. In other words, the relevant grammar items for the topics on that page.

TimLethbridge:
One more point I just noticed. 

Please keep umpleonline/ump/manualexamples/Traits1.ump
and please add back this at the end of the first traits manual page

This is a good example of practical use. And it has been mentioned already in teaching materal.

vahdat-ab:
@TimLethbridge 
Item 5 was fixed.
Regarding `Traits1.ump`, it was part of examples. I've moved it to the section "Attributes in Traits". The file name is "traits_example_003.ump".
I also added it to the page "Traits" as you requested. I also added some paragraphs from your previous page about traits to the new one.

I also added new pages, "Traits and Umple Mixins" and "Flattening of Traits"

PR #1079, opened 2017-09-06
Composite Structure

ahmedvc:


TimLethbridge:
I have tested this privately and all seems OK.

PR #1082, opened 2017-09-18
Remove references to absolute system paths in UmpleOnline

### LINKED ISSUE #1081: UmpleOnline expects runtime dependencies to be available at specific system paths
## Brief Description

UmpleOnline fails to run some of its subprocesses on machines that do not have the programs it requires installed at the exact locations it expects.

## Minimum Steps to Reproduce

1. Find a machine where GraphViz's dot is installed somewhere other than `/usr/local/bin`. The Nix package manager can be used for this. Run `nix-shell --pure -p php56 jdk ruby ant git nettools rake graphviz zip` to generate an appropriate environment.
2. Rebuild Umple
3. Launch UmpleOnline (I do this with the builtin PHP server, should be the same with Apache), with default settings.
4. Try to generate a state machine diagram

## Expected Result

The diagram appears.

## Actual Result

```
dot (part of GraphViz) does not appear to be installed on the server in /usr/local/bin/dot
```

fhackett:
## Description

UmpleOnline uses absolute system paths to refer to some of its
runtime dependencies.

This is unnecessarily unportable, since it shouldn't actually matter precisely where these
executables are located as long as they are available on `$PATH`. It is also inconsistent
since only some executables are referred to in this way.

The prefixes `/usr/bin` and `/usr/local/bin/` are removed from the shell invocations that
have them, allowing UmpleOnline to tolerate variations in the location of the executables
it invokes.

## Tests

This change should not affect anything, and is tricky to test since it affects only the relationship  between UmpleOnline and the system on which it is installed.

Closes #1081


TimLethbridge:
This PR does not work on MacOS where by default the path for php exec is very restricted to reduce opportunities for command attacks. /usr/local/bin and /usr/bin have to be specified absolutely by default. There may be a workaround (e.g. setting search path in the script). Meanwhile, this PR cannot be merged.

fhackett:
@TimLethbridge I've added some code to work around the Mac issue - the exact what and why are in the comments. The code should be pretty harmless on any POSIX machine (even NixOS).

I'm not sure how it would fare on Windows though... would someone be able to test and/or clarify what a Windows setup looks like?

TimLethbridge:
Historically, Windows users who want to run UmpleOnline have had to edit their install. I don't think very many people have bothered. Will test your updated PR later.

PR #1083, opened 2017-09-19
Fixes issue 1026

### LINKED ISSUE #1026: A syntax error when setting an attribute as a key, and this attribute is defined in a parent class
1- Define an attribute, id for example, in a parent class. 
2- Extends this parent class, and set this id attribute to be a key at the subclass as shown in the model below. 
3- Generate the model and you will get a syntax error in the Java code. 

The reason is that in the code generated, we access the field directly, which is private at the children levels. In order to overcome this issue, we should use the getter methods instead.

class Test{
  id;
}

class Sub{
  isA Test;
  key {id}
}

### LINKED ISSUE #1089: Equals function outputs syntax error in Java when key is an Integer[ ]
## Description

Using the following Umple code and generating java output will result in a syntax error:
class Test {
  Integer[] vars;
  key {vars};
}

Specifically, this creates an equals() function which compares each value of the array to that of another array. The issue is that the Integer array is cast to an int upon each comparison and then checked if it is null. We need to check if all the values in the array are null (since an Integer can be null), but casting to an int is invalid since an int cannot be null.

JoshuaMcManus:
The equals() and hash() functions for keys within a class now exclusively call get methods
This prevents the issue of trying to access private inherited members directly while also keeping a consistent format

https://github.com/umple/umple/issues/1026

## Description

Applies a fix to issue 1026

## Tests

Had to change existing tests since they were expecting java code to be produced which directly accesses member variables (we now only call the gettters). All other tests passed and those that were modified now pass as well.

Closes #1026  
Closes #1089

### DIFF
--- a/UmpleToJava/UmpleTLTemplates/equals.ump
+++ b/UmpleToJava/UmpleTLTemplates/equals.ump
@@ -11,18 +11,19 @@ class UmpleToJava {
     AssociationVariable as = uClass.getAssociationVariable(memberId);
     if (av != null)
     {
-      String avString = av.getIsDerived()?(gen.translate("getMethod",av)+"()"):gen.translate("attributeOne",av);
+      String avString = gen.translate("getMethod",av)+"()";
       if (!av.isImmutable() || av.getIsLazy()) canSet.append(StringFormatter.format("    {0} = false;\n",gen.translate("attributeCanSet",av)));
       if (av.getIsList())
       {
-        checks.append(StringFormatter.format("    if ({0}.size() != compareTo.{0}.size())\n",gen.translate("attributeMany",av)));
+        String avStringMany = gen.translate("getManyMethod",av) + "()";
+        checks.append(StringFormatter.format("    if ({0}.size() != compareTo.{0}.size())\n",avStringMany));
         checks.append(StringFormatter.format("    {\n"));
         checks.append(StringFormatter.format("      return false;\n"));
         checks.append(StringFormatter.format("    }\n\n"));
-        checks.append(StringFormatter.format("    for (int i=0; i<{0}.size(); i++)\n",gen.translate("attributeMany",av)));
+        checks.append(StringFormatter.format("    for (int i=0; i<{0}.size(); i++)\n",avStringMany));
         checks.append(StringFormatter.format("    {\n"));
-        checks.append(StringFormatter.format("      {0} me = {1}.get(i);\n",gen.translate("type",av),gen.translate("attributeMany",av)));
-        checks.append(StringFormatter.format("      {0} them = compareTo.{1}.get(i);\n",gen.translate("type",av),gen.translate("attributeMany",av)));
+        checks.append(StringFormatter.format("      {0} me = {1}.get(i);\n",gen.translate("type",av),avStringMany));
+        checks.append(StringFormatter.format("      {0} them = compareTo.{1}.get(i);\n",gen.translate("type",av),avStringMany));
         checks.append(StringFormatter.format("      if (me == null && them != null)\n"));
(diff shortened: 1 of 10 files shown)


TimLethbridge (changes requested):
THis PR needs one or two additional tests for the code that resulted in the issue to start with. In other words there should be a test for code
```
class Test{
  id;
}

class Sub{
  isA Test;
  key {id}
}
```
demonstrating that it parses and/or generates correct output.

JoshuaMcManus:
I have added some testbed tests to test the generated java output to verify that it is syntactically correct. I have also added a fix to issue 1089 because it was discovered while writing some of the tests and was blocking the completion of this task.

vahdat-ab:
@TimLethbridge  I think there might be some cache issue with Travis for the PR. I could do a full build without failure on my machine and Joshua did as well. I think we had such an issue before as well.

Nava2:
@vahdat-ab I blew out the relevant cache and re-ran the build, best of luck. 👍

vahdat-ab:
@Nava2 Thanks. Looks like it didn't resolve the issue :-(

TimLethbridge:
I am going to test it on my machine. If it works there too, it might be something to do with the version of the java compiler still being cached. I have seen these PR builds fail on Travis before, then succeed after the merge. Stand by.

TimLethbridge:
OK So this PR fails the following:

AssociationEndTest	equals_checkAllData_nullNowSameAsEmpty	Failure	expected:<true> but was:<false>

junit.framework.AssertionFailedError: expected:<true> but was:<false>
at cruise.umple.compiler.AssociationEndTest.equals_checkAllData_nullNowSameAsEmpty(AssociationEndTest.java:292)

It needs updating and testing again before merging @AlexJoshuaMcManus

TimLethbridge:
It might be that this fails after merging due to some subtlety; perhaps the wrong files end up being present after the merge. When I went to test it, it said that master had diverged. It prompted me twice to merge. I believe my results are the same as the results that Travis is getting. It is not clear why others don't see the same failed tests. Are you searching through the logs of the full build to look for failed tests? Are you looking at the generated QA page?

JoshuaMcManus:
I'll try looking into it more, but I have looked through the full build output as well as the QA page and neither mention any failed tests. @vahdat-ab was also getting the same failed tests, but he did a fresh clone of the repository and saw all tests pass. Did you clone a new version of the repository, or just pull in my branch, @TimLethbridge? 

I have already done a fresh clone of the repository and tested it with success. What I think I'll do if I can't identify the issues is just create a new PR with a new branch and copy over my changes. I'm suspecting that this is because I didn't rebase before committing the first time.

Nava2:
Have you tried manually merging (or rebasing) master into your branch? As a temporary test, try that and see if there's any issues. Re-doing your changes doesn't help find the issue.

JoshuaMcManus:
Yes I have rebased with master and committed the update. I will try doing it again (since master has been updated since I last did my rebase) when I have the time and update you on the results. Thanks for the help.

vahdat-ab:
Just a report:
1) I did make a fresh repository and didn't build it.
2) run the following commands as Github recommended.
```
git fetch origin
git checkout -b fix-1026_jmcmanus origin/fix-1026_jmcmanus
git merge master
```

3) I did a first build.
4) I did a full build.
5) It passes all test cases. I run this on Windows.

TimLethbridge:
OK following these instructions results in a pass on my machine. so I am going to risk it and merge. I think the mixup with wrong editing of history during the sprint caused some issues. Let's hope this works.

JoshuaMcManus:
Okay, thank you for the update. Please let me know if you see any abnormalities and I'll return to this issue and investigate it further.

PR #1100, opened 2017-10-03
Bala issue#322

baala108:
I have removed const modifiers from attributes and implemented the requested feature for java


UmpleToJava/UmpleTLTemplates/JavaClassGenerator.ump
```diff
@@ -60,7 +60,7 @@ class JavaClassGenerator {
     JavaClassGenerator <<!<</*JavaClassGenerator*/>><<@ UmpleToJava.UmpleNotice >>
 <<#
   UmpleClass uClass = (UmpleClass) uElement;
-  globalUmpleClass = uClass;  
+  globalUmpleClass = uClass;
```
TimLethbridge, line 63: You are just takng out spaces at the end of the line here. Almost all other changes in this file are similar to this. What is the point of the changes in this file. It just makes reviewing harder. It is OK to clean up spaces and indentation, but that should be done as a separate refactoring activity.

UmpleToJava/UmpleTLTemplates/JavaClassGenerator.ump
```diff
@@ -389,19 +389,17 @@ for (StateMachine smq : uClass.getStateMachines())
...
-  if (uClass.getExtraCode() != null && uClass.getExtraCode().length() > 0) { #>>  
+  if (uClass.getExtraCode() != null && uClass.getExtraCode().length() > 0) { #>>
   //------------------------
   // DEVELOPER CODE - PROVIDED AS-IS
   //------------------------
-  
-  <<=uClass.getExtraCode()>>
-  
 
+  <<=uClass.getExtraCode()>>
```
TimLethbridge, line 402: This change will delete a space that has always been present in such code, rendering many tests failing

TimLethbridge:
@bala-v09 This PR has some problems: The first is that you have made a huge number of changes that are merely deleting spaces from the ends of lines. That a) makes it hard to review b) results in conflicts with other people's changes (ones just merged, or ones they are working on) and c) in some cases (templates) can made generated code differ from what was previously being generated, resulting if failed tests. All we want to see is the minimal set of code changes.

There are benefits at times to cleaning up code (removing spaces, fixing indents, etc.). But don't do it as part of a separate PR. Maybe you are using an editor that is doing this automatically; you would have to turn that feature off.

Also you do have failed tests.

baala108:
The indentation issue could be because of my editor, I will fix that and update it soon.

Issue #1084, opened 2017-09-20
Umple gets into infinite loop with key in a subclass with infinite cycle error


TimLethbridge:
The following crashes the compiler. Likely the key check is looking through the infinite superclass chain before the chain itself is detected

```
class Watershed
{
  Integer id;

}

class District
{
  isA District;
  Integer id;
  key{id}
}

```

TimLethbridge:
This bug may also occur if non-direct cycles are present

There may be similar bugs in other cases where we search the inheritance hierarchy. The solution is either to put the cyclical check first before walking the hierarchy, or detecting loops during other walks. Please search for possible other cases. I have noticed Umple getting into an infinite loop and crashing once in a while; this is the first time I have actually identified a failing case.

I want one of this terms UCOSP students to prioritize this please.

TimLethbridge:
One more thought: The ideal solution is still to put the cycle check first.

However, if the above is hard to do, or even if it is not hard, for extra protection in *all* searches when doing analysis, I suggest limiting them with a counter at some reasonable maximum. For example you would never have an inheritance hierarchy of 25 levels, so this would be a very conservative maximum. When starting to search set a variable to 25 and count down at each iteration. If you reach zero, then stop the search. That will prevent infinite loops. 

The depth of most inheritance hierarchies is zero or one. Occasionally a hierarchy of 4 or 5 is found. Legitimate hierarchies higher than 12 are almost unheard of. Double that to 25 to get a safe search limit.

vahdat-ab:
This is probably related to the issue #449.
Check the method `private void analyzeKey(Token keyToken, UmpleClass aClass)` in class `UmpleInternalParser`, file `UmpleInternalParser_CodeClass.ump`.
Look for the comment `//Issue #449` and investigate the while loop.

TimLethbridge:
I notice you say in your log that you check for cycles every time a token is parsed. I trust that you have taken account of efficiency and that this will not have a large impact on that aspect.

Also I notice you say that now this error is perhaps appearing instead of some other errors as this is being detected first. Make sure the other errors still have some way of being tested.

JoshuaMcManus:
I ran 3 builds each using the old code and my new code (2 first/quick, 1 first/quick) and noticed no significant changes in run-times. They all averaged out to be roughly the same time. My tests were not very scientific, but they at least didn't show any immediate increase in compile time.

I had two tests that were failing 

- The first was from: 

```
class Student
{
  isA Proffesor;
}

[...]

PR #1102, opened 2017-10-06
Fix 1084

JoshuaMcManus:
## Description

Fixes issue #1084 

The changes factor out the code for adding a key to a class to be executed at a later point in the compiler process. The issue was that if a key is defined in the same class that also creates a cycle in dependencies an infinite loop would occur. This is fixed by delaying adding the keys to a class until after the cycle check is complete.

## Tests

Created two new tests that caused an infinite loop in the old code, but work with this commit. All existing tests pass without modification.
CLOSES ISSUE #1084


TimLethbridge:
Looks good to me

PR #1103, opened 2017-10-07
Fix 1058 runqing

### LINKED ISSUE #1058: Association class gives wrong error messages when reflexive
The following Umple code generates an incorrect warning. It should be naming the associations after the role names

Error on line 3 : Class 'Profile' may not have multiple associations with the same name 'friendships'. More information (19)

```
associationClass Friendship {
  * Profile invitee;
  * Profile inviter;
  status;
}
class Profile {
    name;
}
```

runqingz:

## Description

Implemented new feature that association class will now accept two roleNames on both side of a single end association. 

## Tests

Added a few JUnit tests for expected result.

Closes #1058


TimLethbridge:
Great. I tested it and it worked.

I am going to merge this PR. However, @runqingz, as a separate PR, could you please add an update to the user manual page, describing the extended syntax. You may want to add an executable example, such as the followng:

```
class A{
  f;
  
  public static void main(String [ ] args)
  {
    A anA = new A("theA");
    C aC = new C("TheC");
    B aB = new B(anA, aC);
    System.out.println("The A is " + anA + " and the bs are " + aB + " and the c is " + aC) ; }
}


 associationClass B{
   * A a1;
   * b2 C a2;
 }


class C {
   d;
}
```

PR #1106, opened 2017-10-09
Issue#322 bala

baala108:
- Resolved spacing and indentation issue caused my editor.
-  Fully local build does not cause any failures or errors. 
test-results


TimLethbridge:
Much better. You do, however, need to enhance the tests. There should be a test of generated code to demonstrate that it looks correct. After that I can merge the PR.

The build now passes. I think it was the spacing before that was causing failure.

Note that after the PR is merged, the next step for this issue will be to locate all occurrences of in the compiler itself and convert them.

vahdat-ab:
@bala-v09 make sure you always point at the issue number in your PR.

PR #1107, opened 2017-10-10
Updated user manual for issue#1058 runqing

runqingz:

## Description

Updating user manual to describe the new grammar changes to  association class (described in #1058).

As for example discussed in #1103, @TimLethbridge I think for a such a straight forward example describing the new grammar syntax, I would put it as a manual example rather than a local executable example.


runqingz:
Closed due to using the wrong base

PR #1108, opened 2017-10-10
Updating user manual for grammar changes in issue#1058 runqing

runqingz:

## Description

Updating user manual to describe the new grammar changes to  association class (described in #1058).

As for example discussed in #1103, @TimLethbridge I think for a such a straight forward example describing the new grammar syntax, I would put it as a manual example rather than a local executable example.


Nava2:
Could you cherry-pick your single commit that changes the manual? You've introduced 5+ commits that are mostly merges. 

Alternately, you could rewrite history to remove them: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

runqingz:
Tried a lot of ways to either remove unrelated commit or cherry pick the needed one, however it is hard to do since the needed commit is at a awkward position, I will create a new PR with a clean branch

PR #1112, opened 2017-10-15
Fixing on rendering Class box with Attributes

runqingz:

## Description

* Fix input element height, the one using before cannot display some
attribute text completely(e.g. String)
* Fix Class box height, it used to render the box too thick
* Fix break line between class name and attribute, used to be too thick
* Add margin to the class name input, it used to overlay with class icon
* Cosmetic fixing (typo) on inline comment

## Tests

Manually tested with local umple online


vahdat-ab:
> Add margin to the class name input, it used to overlay with class icon

I checked it and it is there still.
Check example `Banking System A` and look at the class `Employee` and `Account`. It looks like the way it is dealt is not the right one because for some classes there is no such issue.

runqingz:
Hi @vahdat-ab , so the problems occurs when the space we allocated for input (size) is fully occupied by the string. In the code, the initial setup for the size is 9, so className smaller than 9 character will not have any problem. The fix to this issue should be 1. Adding margin so the overlap never occur (I think by doing this I should also trim down the input size when possible to avoid big white space at the beginning of the line) 2. Initially setup the input size to be large enough for any proper class name. 
screen shot 2017-10-17 at 6 35 04 pm
screen shot 2017-10-17 at 6 38 28 pm

vahdat-ab:
@runqingz thanks.
Please fix the conflict and I'll check it then.

vahdat-ab:
@runqingz 
Thanks. It works as you mentioned.
Just curious. does JoinJS support `aligned centered` for titles (or labels)? I feel like this is the right decision for the names of classes.
If there is no such a capability, I'll merge this.

runqingz:
@vahdat-ab , sure, I will dig into that after my midterm today, but just for clarifying, the most expect way we should have here should be align the input box to the centre of outer class box, align the text to the centre of input box, is that correct?

vahdat-ab:
by saying outer class box, you mean the whole box related to a class?
If this is the case, yes. The reason is that the class might become bigger because of having a method with a longer name. In that case, the size of the class will be bigger and so the name of the class should be adjusted to the center of the bigger box.

runqingz:
@vahdat-ab updated, please check

Issue #1101, opened 2017-10-04
Umple json generation fails in certain cases when traits present


TimLethbridge:
The following code fails when attempting to generate json from it as occurs, for example, when drawing editable diagrams in UmpleOnline.

```
class Color {}

trait TDrawable {
  0..1 -> * Color color;
}

class Z {
  isA TDrawable;
}
```
This results in the compiler crashing
at JsonGenerator.extractClassName

Note that a much large example of this is the Geometric Objects example in UmpleOnline. That will generate graphviz. Somebody has special-cased the example to avoid it trying to generate Json.

People using this example regularly crash UmpleOnline, so this problem needs fixing quickly.

JoshuaMcManus:
Was reading through the description of a Trait and noticed that the last line in the first paragraph on this page is repeated: http://cruise.eecs.uottawa.ca/umple/Traits.html

I'm not sure where to report typos like that so I'll post it here for now.

I also found that the link on the line: "Note that traits themselves are subject to being **mixed** in." returns a 404.

vahdat-ab:
> I'm not sure where to report typos like that so I'll post it here for now.

Thank you, Josh. I fixed them and push a new branch. I'll merge it soon.

JoshuaMcManus:
After discussing with Vahdat, there are few things to note about this issue:
1. The association defined in TDrawable still exists in the model's list of associations when generating the json - this is why the null pointed exception is thrown in extractClassName(). I believe that the only association that should exist is 0..1 Z -- * Color color. This association does exist, but it is missing its coordinates since it is applied to the class after postTokenLayoutAnalysis(). When the Json generator executes, it gives it a coordinate of (-1, -1), (-1, -1). 

Vahdat noted that the associations should be maintained at the model level for futrueproofing the project (in case it needs to be used later). So I've created a solution which just skips generating Json for any association that involves traits, which should be sufficient for solving this issue.

2. The coordinates of the associations are never set because the trait's associations are applied to a class in the function applyTraits() which is called after the postTokenLayoutAnalysis() function is executed (which generates all the coordinates for classes, associations, etc.)

I am going to have to rearrange the code somehow, which I will investigate further.

vahdat-ab:
@AlexJoshuaMcManus  I brought back associations added to the model.
Please check the branch named `association_traits` and see if this is what you were looking for or not.
Once you confirm, I'll merge it.
Thanks

JoshuaMcManus:
[...]

PR #1115, opened 2017-10-17
Fixes Issue #1101

JoshuaMcManus:
## Description

Ignores any associations involving a trait when generating the json

Rearranges the order in which postTokenAnalysis functions are called so the traits are applied to the classes that inherit them before layoutNewElements() is called

## Tests

Adds two tests to verify that traits that contain associations now successfully generate json files. The first is a class that inherits a trait which has an association with another class. The second is the same as the first, but with positions of the classes already specified for the json.

Closes ISSUE #1101


vahdat-ab:
Thanks.

PR #1124, opened 2017-10-28
Split the UmpleOnline Dockerfile to save space

### LINKED ISSUE #1121: Squash the size of Docker image by creating a base image and doing minor pushes between releases
We are now pushing Docker images of UmpleOnline. However the image is quite big (300MB ...) and each new branch, pr and merge to master creates a new one. However the most common changes are either replacement of one of the Jars or some other file in the umpleonline hierarchy. These can be pretty small. It should therefore be possible to set up a situation where there is a base umple image, with most pushes only pushing changed files.

The proposal is to push to an umpleonline-base image only on every release (tag on github), and then to only push other changed files between times.

That would require the publishing script to figure out the deltas, but that shouldn't be too hard.

There is a new --squash tag on the Docker build command. Could this be used to merge changes into the UmpleOnline layer? I note there are already two layers created ... one for the COPY and one for the change of permissions. At least these could be squashed.

Some ideas about making smaller Docker images are here: https://developers.redhat.com/blog/2016/03/09/more-about-docker-images-size/
https://opensolitude.com/2015/05/13/docker-images-should-be-small.html

fhackett:
## Description

The header of the Dockerfile that installs all the dependencies is now
split out into an image called umpleonline-base. It should not need to
change much and saves 200MB from the full UmpleOnline image size.

The Travis build has been altered to base its Dockerfile on that image, and
rebuild it on tag builds.

## Tests

As with other build system modifications, all tests are live.

The Dockerfiles have been tested independently, hopefully the tag build logic works.

Closes #1121

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,20 +54,22 @@ script:
 
   # Set up UmpleOnline to be packaged into its Docker image, then build the image.
   - if [ "$HAVE_DOCKER" ]; then
-        $ANT_BUILD -DshouldPackageUmpleOnline=true packageUmpleonline;
+        $ANT_BUILD -DshouldPackageUmpleOnline=true packageUmpleonline &&
+        docker build -t "umple/umpleonline-base:local"
+                     -f ../umpleonline/Dockerfile-base ../umpleonline &&
         docker build -t "umple/umpleonline:$TRAVIS_COMMIT" ../umpleonline;
     fi
 
 after_success:
     if ( [ "$HAVE_DOCKER" ] && [ "$TRAVIS_SECURE_ENV_VARS" ] ); then
-        docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD";
-        DOCKER_IMAGE_NAME="umple/umpleonline:$TRAVIS_COMMIT";
-        docker push "$DOCKER_IMAGE_NAME";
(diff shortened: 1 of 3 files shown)


fhackett:
That's odd... all my stuff worked, but the build failed on testbed. Any thoughts @TimLethbridge ?

TimLethbridge:
So two things here. First the pr build is passing, which is what matters. Not sure why the branch build is failing, but I have seen that before. Something to do with resetting version.

Second issue is that when I pull the docker image for this PR I don't see an image that makes use of the new base image. Not sure why.

fhackett:
@TimLethbridge thanks for checking, I've relaunched the push build just as a matter of principle (need to test the PR builds branches correctly, though I think it does). Looks ok one way or another,

As for you not seeing the new base image, I agree in that I can't really see it either. When I pulled the PR image however, it did only download ~40mb of data so I think it works, it's just not obvious. When (hopefully) the branch build passes, you can pull that too and you should see a significant improvement. Since I imagine the PR build was the first one based on my changes you downloaded, it would have had to download the base image as well at that point. That should be a one time cost.

fhackett:
I give up on the push build, I have no idea how to make it pass.

On the other hand, I reran the pr build so if you pull that again you should only get the UmpleOnline files. If you view it in `docker history` you should see distinct creation times for the 200mb of dependencies and the other files. Your download time should also be shorter.

Issue #1125, opened 2017-10-28
JointJs failed to delete existing association when removed in UML class diagram


runqingz:
## Brief Description

In some cases when using umpleonline, even if the association is removed from UML class diagram under JointJs option, the corresponding association is not removed from the umple code.

## Minimum Steps to Reproduce
Step 1. Sync the graph under JointJs option, using the following simple umple code:
```
class NewClass1
{
}

class NewClass2
{
  * -- * NewClass1;
}

```
Step 2. click the remove button on the association line (removing the association on the diagram)

## Expected Result

```
class NewClass1
{
}

class NewClass2
{
}

```

## Actual Result

```
class NewClass1
{
}

class NewClass2
{
  * -- * NewClass1;
}

```

## Note

This issue is related to Issue #600

PR #1126, opened 2017-10-28
Fix the issue that the removing association in JJs might not work correctly

runqingz:
## Description

This commit fix the problem mentioned in issue #1125.

 However, I am not really sure if there is such a need to do this class name check and set current association name end by using dictionary order of the class names. From my perspective, class names do not really matter, we always remove the association if there is such a subtoken in side a class definition. 

This check will raise the risk of failing to pick up association definition in umple code under cases like #1125 .


TimLethbridge:
This PR is failing a test that is expecting the PlaygroundMain class (the main program that is used in UmpleOnline) to work the old way.

runqingz:
@TimLethbridge updated, now rather than changing the behaviour of complier, Jjs is forced to name association by alphabetically ordering Type name of two end.

PR #1130, opened 2017-11-02
Add feature to Umpleonline JJS class diagram - editing association end

### LINKED ISSUE #1127: Implementing new feature that allows modification in JJs Association is reflected in Umple Code
## Brief Description

Currently, editing association end by moving the association line among different class in JJs class diagram will not result in refreshing Umple code to reflect those changes. However, this feature should be implemented for better User experience.

## Minimum Steps to Reproduce

In Umple online

1. Create multiple classes
2. Create a new association among any pair of the classes.
3. Move the association link to different class

## Expected Result

Umple code should refresh to reflect the new structure of the diagram

## Actual Result

Currently JointJs option does not have such feature

***Note:*** This is relate to issue #600

runqingz:
## Description

Implements feature that allows users to modify association end when modifying the link connection in JointJs class diagram.

Closes #1127

### DIFF
--- a/umpleonline/scripts/jjs/jjs_parse.js
+++ b/umpleonline/scripts/jjs/jjs_parse.js
@@ -114,12 +114,46 @@ var JJSdiagram = {
 			}
 		);
 
+		//The following disconnect and connect event is intended to implement modifying association end, at the moment
+		//the workflow is remove then add, please refer to issue #1127 for detail
+		this.paper.off('link:disconnect').on('link:disconnect',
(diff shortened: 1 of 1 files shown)


vahdat-ab:
@runqingz 
I thought your solution is going to be applied to all kind of associations.
Now, we change the associations graphically we can see the reflection in the code.
This doesn't happen for inheritance.
if it is easy to fix, just give us a patch, otherwise, please create an issue (and a PR) for it.
Thanks.

Issue #1134, opened 2017-11-11
No Umple // line traceback for State Machine do statements


JoshuaMcManus:
## Brief Description

When creating a state machine, the generated Java "do" statements do not produce a comment in the form:
// line ## "<Path to Umple>"

Users who produce invalid Java code will not be able to identify where in the Umple code the issue exists.

## Minimum Steps to Reproduce

```
class LightFixture  
{  
  bulb  
  {  
    On {  
      do { doThisContinuouslyWhileOn(); }  
    }  
    Off {}  
  }  
  
  void doThisContinuouslyWhileOn() {  
    while (true) {  
      System.out.println("Still on");  
      try {  
        Thread.sleep(1000);  
      }  
      catch (InterruptedException e) {}  
    }  
  }  
}
```

## Expected Result

The do statement should look something like this:
```
private void doActivityBulbOn()
  {
    try
    {
      // line 6 "statemachine.ump"
      doThisContinuouslyWhileOn();
      Thread.sleep(1);
    }
    catch (InterruptedException e)
    {

    }
  }
```

## Actual Result
```
private void doActivityBulbOn()
  {
    try
    {
      doThisContinuouslyWhileOn();
      Thread.sleep(1);
    }
    catch (InterruptedException e)
    {

    }
  }
```
This is blocking issue #882

PR #1135, opened 2017-11-12
Fix 1134

JoshuaMcManus:
## Description

Do statements for state machines did not output // line ## "Filename" comments. This fixes this by injecting the comment into the generated Java code.

## Tests

Changes two tests to verify comments.

Fixes Issue #1134

### DIFF
--- a/UmpleToJava/UmpleTLTemplates/state_machine_doActivity.ump
+++ b/UmpleToJava/UmpleTLTemplates/state_machine_doActivity.ump
@@ -12,6 +12,9 @@ if (p != null) {
   {
     try
     {
+      <<#if (p != null) {
+      realSb.append("\n      // line " + p.getLineNumber() + " \"" + uClass.getRelativePath(p.getFilename(),"Java") + "\"");
+    } #>>
       <<= activity.getActivityCode() >>
       Thread.sleep(1);
       <<# if (postTransition != null)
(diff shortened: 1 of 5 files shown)


vahdat-ab:
@AlexJoshuaMcManus please check why your build is failing

JoshuaMcManus:
@vahdat-ab it looks like the tests all passed in Eclipse, but when they're run in test suite they fail because the relative directory changes for the // line output.

How would you suggest I test my changes, if at all?

vahdat-ab:
@AlexJoshuaMcManus 
I tried on my own machine, I faced some failures.
```
    [junit] TEST cruise.umple.statemachine.implementation.StateMachineTest FAILED
    [junit] TEST cruise.umple.statemachine.implementation.java.JavaStateMachineTemplateTest FAILED
    [junit] Tests FAILED
```
When I switch to `master`, it passes all test cases. It means there are issues in your branch.

JoshuaMcManus:
@vahdat-ab 

Sorry, I guess I wasn't clear. Yes I know the tests that I implemented are the ones that are failing. This is because I modified existing tests to look for the new comment that reads:
// line ## "path/to/file"

However, when the tests are run in Eclipse, the path to file is something like "../file.ump", but when the tests are run during a full build they look more like: "../../.../../cruise.umple/test/file.ump"

How would you suggest I go about fixing this? The tests are looking for the comments to match (since that is what I implemented), but it looks like the relative directory changes depending on how the test is called.

vahdat-ab:
I see.
The main priority is the command line. The server always runs tests cases through the command line.
w.r.t to the issue with `Eclipse`, I would say check other test cases in which we test the location and address of files. We have not had any issue with it.

JoshuaMcManus:
@vahdat-ab I've changed the test so it's now passing the build,  but I noticed that 3/4 Travis builds are failing, is that supposed to happen? I remember always failing 2, but not 3.

PR #1165, opened 2017-11-25
Trivial fix jointjs runqing

runqingz:
## Description

* edit button no longer behaves like a check box
* fix class box layout to avoid overlapping on existing umple examples
* Fix some race condition and other issue when using edit box
* Improve the usably of edit box
* Enable delete option now it shows the red cross delete button for the
class box (this is the only way to show it)


TimLethbridge:
I have tested this PR using Docker! Thanks for enabling that.

The delete functionality is certainly improve. (At one point, after doing a series of operations, I was able to get the red x's to no longer appear though ... but can't reproduce).

I note the following problems:

1. When I click on the pencil to edit an attribute, although the attribute is selected and I type new text, no update is applied to the model (this I think was a problem existed before this PR, but I was expecting the PR to fix it as it was dealing with editing attributes).

2. Selecting a class in the diagram no longer highlights the class on the text side (this is clearly new in this PR).

3. When creating a new class, the edit box for the class name is not on the same line as the class name (new in this PR).

4. When I edit a an attribuite and hit return, the edit pencil for 'Add more' is always active (this should only happen after hitting return when adding moew).

4. When I click on the pencil by 'add more' and add a new attribute, the diagram doesn't show the '-' consistent with 'private' as do the others, and the default 'String' does not appear. The only way to get these to appear is to change the diagram type and then change it back.

There are other glitches, but I am not sure they are related to this PR - fix the above for now please.

runqingz:
@TimLethbridge for number 3, what browser are you using? It works find on google chrome on my machine

TimLethbridge:
Aha! Yes @runqingz item 3 is only a problem in Safari. Other browsers work OK (Firefox, Opera, Chrome). But Mac users use mostly Safari.

runqingz:
@TimLethbridge in joint.js, do we allow user to input "- a:String" as attribute? I notice this is not allow in default editing mode, but is not really prohibited in joint.js.

TimLethbridge:
@runqingz In Umple, all attributes are private. Users don't enter - and the system doesn't store the -. My point was that visually, newly added items need to look the same. Your code was taking from the repository and putting the - there. Either all attributes show the - or none do. Same with type. Umple's default type is String. The existing joint.js code is filling this in. Visual consistency is important. Either String should be hidden if the textual Umple omits it (for all attributes), or else you should visually show it for new attributes

runqingz:
@TimLethbridge please check, edit & delete attribute should work as expected now. Similar fix should apply to method but there are more things to consider(scope and parameters). At moment I need to pause and focus on my finals right now, I will try to find some time and fix the problem with editing & deleting method and document my work.

For now, do you think it is okay to merge this PR as a fix to attribute and raise issue for delete & editing method (potential risk of defection) plus the paper offset you mentioned?

Also this is a private issue, as my mentor point out she need feedback grading feedback to the steering committee on the 17th, how many days could I still possible have after 14th (include)?

TimLethbridge:
@runqingz I am going to merge this. There are still other joint.js issues that have been raised, and I judge it still not at the level of usable to be out of 'beta', but this PR makes it better than it was before.

PR #1172, opened 2017-12-04
joint.js photo sync mode should hidden icon&place holder display

### LINKED ISSUE #1139: Photo ready should hide the 'garbage can' and 'add more' in joint.js
The purpose of the photo ready option in UmpleOnline is to enable people to make diagrams that can be used in documents. However this does not work in joint.js class diagrams since the garbage can and 'add more' are still visible. In the original editable diagrams these are removed. The workaround is to use one of the other diagram types, but this is not ideal.

See also #600

runqingz:
## Description

This PR provide feature that joint.js photo ready option should hide all icon, button & place holder display when checked, and show all of these when unchecked.

Closes #1139

### DIFF
--- a/umpleonline/scripts/jjs/jjs_parse.js
+++ b/umpleonline/scripts/jjs/jjs_parse.js
@@ -385,6 +385,25 @@ var JJSdiagram = {
 			}
 		});
 
+		jQuery('#buttonPhotoReady').change(function () {
+			if (this.checked){
+				jQuery('.html-element button.delete').css('visibility', 'hidden');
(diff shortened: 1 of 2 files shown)


Nava2:
Could you use more descriptive titles in your pull requests? It helps when searching for content. "JointJS icon visibility fixes (#1139)" would be a good one.

TimLethbridge:
I tested this. It does what it claims and is a simple agile improvement. Great. One step towards making the joint.js capability fully usable.

PR #1177, opened 2017-12-09
Fixes #1161 null check on delete

### LINKED ISSUE #1161: Cascading delete fails when two compositions are chained
In the following code where B is composed of C, and C is composed of D, a null pointer exception is raised when deleting a B. This should be solvable by adding a non-null check around the offending line in the generated code.

Note that compositions necessarily involve cascading delete, it is the whole point of using a composition.

```
class A {
  // run the main method
  public static void main(String[] args) {
    B aB = new B();
    C aC = new C(aB);
    aB.addMyC(aC);
    // this causes a null pointer exception
    aB.delete();
  }
}
class B {
  1 <@>- * C myC;
}
class C {
  1 <@>- 1 D myD;
}
class D {}

```

This problem was flagged by @gmussbacher gunter.mussbacher@mcgill.ca and is a high priority for his work. He may wish to comment further.

TimLethbridge:
This fixes issue #1161. There was a need for additional null checks in certain delete methods.


TimLethbridge:
This fixes #1161 - When two deletes were chained together, the second resulted in a null pointer to the first.

PR #1178, opened 2017-12-09
Implement issue 1123

fhackett:
## Description

This PR is a full refactor of UmpleOnline's data storage. It implements the interface described here: [https://github.com/umple/umple/wiki/Refactoring-UmpleOnline's-storage-backend](https://github.com/umple/umple/wiki/Refactoring-UmpleOnline's-storage-backend).

All functionality should remain unchanged, the only difference should be the code structure.

What the structural changes give us:
- code now makes no assumptions about having a persistent working directory, requesting one explicitly when needed. This means if we were to change the storage backend we would know when the code expects a working folder and would be able to provide one.
- any PHP code must explicitly request access to any data it needs. This allows putting in place various concurrency control constructs as and when needed.
- any generated data that will be presented to the browser via hrefs will go through an API that abstracts away how we ensure they persist. This becomes a significant issue if we try to scale UmpleOnline past 1 server due to load balancing, but the hooks provided here should make implementing any support for that manageable.

Notable exception to the feature freeze: passing arbitrary paths using the filename query parameter will no longer work as internally UmpleOnline now knows/cares more about where things go. The filename parameters look the same, but are simply parsed for their model ID file name. 

## Tests

As per usual with UmpleOnline all testing is manual. Testing is currently incomplete, merging is not recommended. All HTTP endpoints related to the file store will need testing.

Parts tested so far:
- Basic code editing
- Some very simple bookmarking scenarios
- Compilation to Java
- Downloading the compiled Java artifacts

### DIFF
--- a/umpleonline/404.shtml
+++ b/umpleonline/404.shtml
@@ -1,0 +1,83 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
+  "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+
+<head>
+  <!--#include virtual="files/inclcommonhead.html" -->
+
+  <title>CRuiSE Server - Page not Found - Error 404</title>
+</head>
+
+<body>
+
+  <!--#include virtual="files/incltopuottawabanner.html" -->
+
+  <!-- The following is the main table, split horizontally into
+       a menu and the primary content -->
+  <table border="0" cellpadding="0" cellspacing="0" width="780">
+    <tbody>
+
+      <!-- The following specifies the part of the left bar menu
+           to highlight as current -->
+      <!--#set var="page" value="cruise" -->
+
+      <!-- The following includes the top row with the uottawa
+           banner, plus the leftmost columns of the main row
+           including the left menu -->
+      <!--#include virtual="files/inclleftmenu.shtml" -->
+
+        <!-- Page Content Area table cell: -->
+        <td id="content">
+
+<h1>A page you requested was not found</h1>
+
+<p>A page you requested was not found. If you came to this page via clicking on a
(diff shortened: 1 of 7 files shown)


TimLethbridge (commented):
Testing has gone well so far. I have not uncovered any defects in the PR yet.

However, I did discover a defect that is present in master! When 'bookmark.php' is run without any errors having been created, it displays an invalid failure message. I will raise an issue separately on that (it says "You cannot make a bookmarked page when JavaScript is turned off. Please turn it on.")

I need to do more testing. I have been testing through docker, pulling pr-1178. I might put it temporarily into 'production' on http://cruise.eecs.uottawa.ca/umpleonline/test/ on Monday. Tomorrow, Sunday I am spending in Montreal with my brother's family, so can't get back to this until Monday morning. We need to have every single command tested repeatedly before merging.

fhackett:
@TimLethbridge glad it still appears to work - I'll do some more myself tomorrow once I've taken my next final. I fully expect there to be at least one typo left in a corner somewhere.

Note: I should update this PR with proper commenting, please don't merge it until I do that - the new API has very specific semantics that are not currently obvious.

fhackett:
@TimLethbridge The documentation is done. Now for testing.

fhackett:
@TimLethbridge I've performed more testing and found + fixed one regression (Javadoc was rendering improperly because I forgot to edit part of a string).

I also found several non-regression bugs:
- YUML output is broken in general
- loading a composite structure diagram then reverting to normal diagrams gets stuck on "No composite structure diagram"
- generating Javadoc on the Ping-Pong example fails, though that seems to be due to invalid Java being generated that is rejected by Javadoc

fhackett:
Not sure what happened to Travis there... seems it broke somewhere in testbed?

TimLethbridge:
OK I am going to merge this! Thanks for pointing out the new bugs. Will raise isues.

It will appear on http://cruise.eecs.uottawa.ca/umpleonline/test/  first
I will promote it after a bit more testing

TimLethbridge:
Don't worry about the travis push fail. That happens periodically,

Issue #1151, opened 2017-11-19
Crash in graphviz generation of in certain state machine guard case


TimLethbridge:
The Umple code below (that is set to generate GraphViz from a state diagram with guards) fails and results in the following stack trace:

```
java.util.NoSuchElementException
	at java.base/java.util.TreeMap.key(TreeMap.java:1324)
	at java.base/java.util.TreeMap.firstKey(TreeMap.java:289)
	at java.base/java.util.TreeSet.first(TreeSet.java:394)
	at cruise.umple.compiler.Guard.add(Guard.java:132)
	at cruise.umple.compiler.Guard.getQueuedExpression(StateMachine.ump:301)
	at cruise.umple.compiler.Guard.getExpression(StateMachine.ump:337)
	at cruise.umple.compiler.GvStateDiagramGenerator.appendStateMachineRecursively(GvStateDiagramGenerator.java:580)
```

The method public ArrayDeque<String> getQueuedExpression() in stateMachine.ump is probably the place to start the search.

```
generate GvStateDiagram "./GvStateDiagram/" --override-all;
// Simple UML state machine of a tic-tac-toe or noughts and crosses game
class TicTacToe {
 state {
   Start{}
   Start {
     xPlays [o()] -> xWin;
     xPlays [i()] -> oTurn;
     xPlays [o()] -> tie;
   }
   oTurn {
     oPlays [i()] -> oWin;
     oPlays [true] -> xTurn;
     
   }
   xWin {
   }
   oWin {
   }
   tie {
   }
 }
 // The following need to be implemented 
 Boolean o() {return false;}
 Boolean i() {return true;}
}
```

TimLethbridge:
Fixed in #1184

PR #1184, opened 2017-12-14
Fixes Issue #1151 - solves problem with crashing graphviz generation

baala108:
This PR fixes Crash in graphviz generation when using boolean literal for guards in state machines.


TimLethbridge:
Good job, I have tested this and it works. It would be better to have an implementation test that generates graphviz text from the umple file and validates that it is correct. But that can be done later. This fixes the crash issue.

PR #1205, opened 2018-01-22
Fixes 1181: Multiple Inheritance in Different Class Blocks

daniel-mitchell:
## Pull request title
Fixes 1181: Multiple Inheritance in Different Class Blocks

## Pull request description
Previously, when using multiple class blocks for the same class, having an "isA" statement in each block pointing to different superclasses would not cause an error. This how the check for multiple inheritance is run so that "isA" statements from other class blocks will be detected.

## Pull request content
Umple.ump: Added methods to determine if the current UmpleClassifier is an UmpleClass, UmpleInterface, or UmpleTrait.
UmpleInternalParser_CodeClass.ump: Changed check for repeated inheritance to be run on entire list of extends, instead of just on those added in the current class block. Also changed it to not report the same class being extended multiple times.
422_multipleClassBlockInheritance1.ump, 422_multipleClassBlockInheritance2.ump, UmpleParserTest.java: Test cases

### DIFF
--- a/cruise.umple/src/Umple.ump
+++ b/cruise.umple/src/Umple.ump
@@ -528,6 +528,11 @@ class UmpleClassifier
   1 -> 0..* Constant;
   1 -> 0..* ModelConstraint;
   * -> 0..1 Token extendsToken;
+  
+  // Methods to distinguishe the subclass types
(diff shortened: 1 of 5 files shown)


TimLethbridge:
Excellent work

PR #1219, opened 2018-02-07
Making manual more responsive

TimLethbridge:
This shortens the line-width of some example code so the user manual looks better.

It also makes the css for the user manual track the width of the page.


TimLethbridge:
This also fixes a bug where the 'saved URL' did not work, and makes the main umpleonline page more responsive.

PR #1234, opened 2018-03-09
Fix Issue 1002 and Refactor Methods

hugessen:
Initially, custom methods without the prefix "custom" or "all" threw a 1012 error when before/after functionality was added. This PR fixes that issue and provides a number of code refactors.

- `getAllMethodNames()` now returns custom methods in addition to only generated ones
- Change default operationSource from "generated" to "all" when checking if code should be injected
- Combine 3 similar methods (`checkCodeInjectionValidityAll()`, `checkCodeInjectionValidityCustom()`) into 1, `checkCodeInjectionValidity()`
- Apply DeMorgan's law to Umple_Code.ump for better readability

### DIFF
--- a/cruise.umple/src/StateMachine.ump
+++ b/cruise.umple/src/StateMachine.ump
@@ -93,7 +93,7 @@ class Activity
   
   CodeBlock codeblock = null;
   after constructor {codeblock = aActivityCode!=null ? new CodeBlock(aActivityCode) : new CodeBlock();}
-  before setActivityCode {codeblock.setCode(aActivityCode);}
+  before generated setActivityCode {codeblock.setCode(aActivityCode);}
   after getActivityCode{
   	if (codeblock.getCode()!=null)
   	  return codeblock.getCode();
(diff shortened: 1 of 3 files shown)


TimLethbridge:
So this PR is faliling and here's why: Step 1: Appveyor builds Umple, generating a compiler with the new code. Then it builds the compiler a second time using the compiler generated in step 1. It turns out that your PR creates a compiler that is invalid (step 1). The error is that the following code injections operative now, but were not operative before. You will have debug by doing a repeated build.

```
StateMachine.ump:96:  before setActivityCode {codeblock.setCode(aActivityCode);}
```

```
StateMachine.ump:161:  before setActionCode {codeblock.setCode(aActionCode);}
```

I suspect the issue is that now the injection is was being done in generated Java line 82

```
public boolean setActivityCode(String aActivityCode)
```

But is now being done also in the custom 2-argument setActivityCode.

The likely solution: Add the word 'generated' in StateMachine.ump lines 96 and 161 after 'before'

The specific failure of the step 1 generated code as reported by Appveyor is:

```
 [javac] Compiling 1365 source files to C:\projects\umple\cruise.umple\bin
1126    [javac] C:\projects\umple\cruise.umple\src-gen-umple\cruise\umple\compiler\Activity.java:225: error: cannot find symbol
1127    [javac]     codeblock.setCode(aActivityCode);
1128    [javac]                       ^
1129    [javac]   symbol:   variable aActivityCode
1130    [javac]   location: class Activity
1131    [javac] C:\projects\umple\cruise.umple\src-gen-umple\cruise\umple\compiler\Action.java:246: error: cannot find symbol
1132    [javac]     codeblock.setCode(aActionCode);
1133    [javac]                       ^
1134    [javac]   symbol:   variable aActionCode
1135    [javac]   location: class Action

```

Issue #1028, opened 2017-03-30
Aspect-orientation does not apply to derived attributes


ahmedvc:
Try the model below. The aspect code block will be ignored in the generated code.

class Test{
  name= {"Some value"}
  
  before getName{
    //I will be ignored
  }
}

TimLethbridge:
We realize that the injection should be for the get method. It still does not work. Example:

```
class Test{
   firstName;
   lastName;
   fullName= {firstName+lastName}

  
  before getFullName{
    firstName = firstName+ " ";
   }
}
```

hugessen:
The problem is in the way `JavaClassGenerator` is written. It has multiple `if` blocks for different cases when injecting code. This code should be getting sent to the `attribute_GetDerivedCodeInjection` block, but is instead being sent to the `attribute_GetDerived`. This is because the `if` statement around `attribute_GetDerivedCodeInjection` requires the existence of `after` functionality. 

If I add an `after` statement, both the `before` and `after` code are appended. It also works if I use `after` alone, but does not work with `before` alone.

PR #1240, opened 2018-03-16
Fix issue 1028

hugessen:
Issue #1028 is that `before` functionality can't be applied to derived attributes. After some digging, I realized that it's because the `if` block to inject `before` and `after` validates the existence of `after` functionality. If there is no `after` block, this injection is skipped over.  See my comment on the issue.

This small fix updates the `if` block to check for _either_ `before` or `after`, not just the `after`.


TimLethbridge:
Please add a template test to verify that in a basic case this generates expected output

Issue #1027, opened 2017-03-30
Aspect-orientation (after/before) does not work at child classes


ahmedvc:
Aspect-orientation works fine at a class (see Test), but oes not work in subclasses (see Sub). Try generating the code below, and you will find that aspect-oriented code in the subclass will be ignored.

class Test{
  name;
  before getName{
    //I will work
  }
}

class Sub{
  isA Test;
  before getName{
     //I will not
  }
}

TimLethbridge:
Logically, this ought to result in a getName that calls the super getName, and injects the relevant code.

TimLethbridge:
Here's another case, where the injection is only in the subclass

```
class Test{
  name;

}

class Sub{
  isA Test;
  before all getName{
    //I will not
  }
}
```

Currently nothing happens in the subclass Sub. But it would be logical for the following to be newly generated Java in this case (and in the original example for this issue).

```
public String getName() 
{
  //I will not
  super.getName()
}
```

Similar code would be needed for PhP, Ruby etc.

TimLethbridge:
When done, also add a page to the aspect oriented part of the user manual.

hugessen:
So it turns out there is functionality already built to do this, but it only works for `set` methods, not `get`. I'd like to know if this was done intentionally. The line is at `JavaClassGenerator.java:4740`

For example:
```
class SCTest{
  name;
}

class Sub{
  isA SCTest;
  
  before all setName{
    System.out.println("Before");
  }
  after all getName{
    System.out.println("After");
  }
}
```
Produces the following:
```
// line 5 "SubclassTesting.ump"
public class Sub extends SCTest
{

  //------------------------
  // MEMBER VARIABLES
  //------------------------

  //------------------------
  // CONSTRUCTOR
  //------------------------

  public Sub(String aName)
  {
    super(aName);
  }

  //------------------------
  // INTERFACE
  //------------------------
  /* Code from template attribute_Set_subclass */
  public boolean setName(String aName)
  {
    boolean wasSet = false;
    // line 9 "SubclassTesting.ump"
    System.out.println("Before");
    // END OF UMPLE BEFORE INJECTION
      wasSet = super.setName(aName);
    return wasSet;
  }

  public void delete()
  {
    super.delete();
  }
}
```

ZainabAlShowely:
It seems like it was implemented only for set methods (both before and after), but it is not only get methods that cause the issue, as this code:
```java
class A 
{
  attr;
  * -- * C;
}

[...]

PR #1241, opened 2018-03-16
Fix issue 1027

hugessen:
This PR fixes issue #1027, where inherited subclass methods couldn't have before/after functionality applied. As it turns out, the functionality was already built in, but only for `set` methods, not `get`. This PR extends functionality so it works on both getters and setters.


TimLethbridge:
Please add a test

PR #1236, opened 2018-03-12
Issue#988 case a) i - removed ignored associations

BernardYuan:
##### Description
Fix for issue 988 case a) i, as mentioned in the [thread](https://github.com/umple/umple/issues/988). Removed some ignored tests.

##### Tests
Unit tests written.

Issue #988, opened 2017-02-04
variables of associations are not accessible in subclasses 


vahdat-ab:
Please consider the following example,
```
class A{
  0..1 sth-- * B;
}
class B{
}
class D{
  isA A;
  0..1 sth -- * B;
}
```
It generates the following code
```
  public B getB_B(int index)
  {
    B aB = (B)bs.get(index);
    return aB;
  }
```
The variable `bs` is not accessible in the class D, so there is an error. It's a private varibale in the superclass `A`.
There several other errors in other APIs in association with the variable `bs`

TimLethbridge:
The solution is likely to change the direct variable access to bs to instead call the getbs method.

When solving this probem search for other cases, and add tests for all cases found.

BernardYuan:
One case I found difficult to fix is the adder function generated from the example given above, in class D, we have:
~~~Java
public boolean addB(B aB)
{
    boolean wasAdded = false;
    if (bs.contains(aB)) { return false; }
    D existingSth = aB.getSth();
    if (existingSth == null)
    {
        aB.clear_sth();
        aB.addSth(this);
    }
    else if (!this.equals(existingSth))
    {
        existingSth.removeB(aB);
        addB(aB);
    }
    else
    {
         bs.add(aB);
    }
    wasAdded = true;
    return wasAdded;
}
~~~

The issues are:
1. Class D has no access to **bs**.
2. method *getBs()* does not return the reference to bs, but a duplicate of it.
3. method *addB()* is overriden, so calling *addB()* in D invokes the method itself, rather than the *addB()* defined in class A.

I have tried to change *addB()* into *super.*, however, the templates whose names end with "_relatedSpecialization" may also be applied to non-derived classes, which makes this fix unrealistic.

TimLethbridge:
Yikes. Tricky. These kinds of issues seem to be getting more and more complex.

In an ideal design, the addB method in B should continue to work, and not be overridden, with only calls to any necessary 'checks' made by it (e.g. for verifying multiplicity bounds) being overridden in subclasses.

I notice only trivial details that differ for the addB method in the subclass D as compared to A. Not sure why the 'clear_sth()' is needed in D, for example. Perhaps this could be put into the superclass'es method (i.e. present in all add methods) and then the override is not necessary.

Another thought would be to add a necessary protected service methods into the superclass that the subclass can use to do operations on the private variable (and not a copy of it). But that increases complexity, where as the solution above decreases complexity.

[...]

PR #1242, opened 2018-03-19
Fixing build from error 19

TimLethbridge:
PR #1236 had created a compiler that would result in error 19 being thrown in subsequent builds. It would also raise warning 170 in many places.

This *temporarily* comments out a key line in  cruise.umple/src/UmpleInternalParser_CodeClass.ump
that triggered the error 19 cases. This will need to be addressed by @BernardYuan by most likely changing the associations that were triggering the problem.

A strictness directive is also added to suppress warning 170, since the compiler allows some custom  constructors.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -137,6 +137,22 @@
     <ant antfile="build/build.testbed.xml" target="testRuby" inheritAll="true" />
   </target>
 
+  <target name="sandbox">
+    <deps-declare-taskdef />
+
+    <trycatch property="foo" reference="bar">
+      <try>
+        <antcall target="qaBuildingPage" />
+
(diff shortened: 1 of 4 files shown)


BernardYuan:
This problem happens because some Umple source code violates the rule **b)** concluded from [issue#988](https://github.com/umple/umple/issues/988).

For example, there exist associations like:
~~~Java
class GenConstructor{
	0..1 -- * GenMethodParameter parameters;
	0..1 -- * GenParentMethodParameter parentParameters;
	* -- 1 GenClass owner;
}
~~~
where **GenParentMethodParameter** is subclass of **GenMethodParameter**, having different rolename, however the two associations have different multiplicities. Therefore it is judged as an Error 19.

PR #1243, opened 2018-03-19
Tentative pull request for issue988 rule b)

BernardYuan:
Tentative for issue#988 rule b), just see whether it would pass CI because it passed my local double-full build.

### DIFF
--- a/UmpleToRTCpp/UmpleTLTemplates/Attributes.ump
+++ b/UmpleToRTCpp/UmpleTLTemplates/Attributes.ump
@@ -249,8 +249,8 @@ class Visibilities{
 	const String GLOBAL= "global";
 }
 
-association { 0..1 GenClass owner -- * GenBody constructorContents;}
-association { 0..1 GenClass owner -- * UniqueGenBody uniqueConstructorContents;}
+association { 0..1 GenClass ownerGenBody -- * GenBody constructorContents;}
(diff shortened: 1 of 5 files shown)


TimLethbridge:
Ok this seems to work. But all it does is undo the commenting out. Not sure *why* it works.

Note that you had said
```
class GenConstructor{
	0..1 -- * GenMethodParameter parameters;
	0..1 -- * GenParentMethodParameter parentParameters;
}
class GenMethodParameter {}
class GenParentMethodParameter {
  isA GenMethodParameter;
}
```

Is code found in the compiler that should trigger error 19 because the 0..1 side needs a different role name in one of the associations (otherwise both associations use the same variable on the GenConstructor side which is bug-prone). So why has the problem gone away with this new PR?

BernardYuan:
I have just compiled the given example with this version of Umple, which, in fact, failed. 

It is for sure that this branch should not be merged, however, I think one thing that would require debugging is the effectiveness of the CI procedure. Probably there are some problems with it that it overlooks some certain errors.

TimLethbridge:
I am going to see about fixing the CI tomorrow if I can, so that a regular full build and also a Travis build would fail if they damage the compiler. But it might take longer than tomorrow.

BernardYuan:
After refactoring it seems now the compiler can pass the CI.

PR #1244, opened 2018-03-19
Fix 958: Conflict Constructors for Java

gublan24:
This PR will make umple raise error if a hard-coded constructor matches one that generated by umple. More discussion about this error can be found in[ issue 958](https://github.com/umple/umple/issues/958)

The error will be present to the user as the screenshot shows:


TimLethbridge:
This is an interesting Pull Request. It is the very first time someone has issued an error from a generator. It had been our policy to only issue errors from the analysis phase before attempting to generate any code of any kind.

I notice that when I run UmpleOnline with the example code, the error does not appear until you actually ask to generate Java. At that point the generated code actually appears (with the conflicting  constructor) along with the error message.

I would like to stick with the original architecture of not allowing any conflicting constructor and detecting it during analysis.

PR #1251, opened 2018-03-24
Manual improvement

TimLethbridge:
Improves the user manual:

1. The grammar files had links to their ancient pages on Google code. The links now point to their pages on Github.

2. Prior to this, any attempt to translate manual pages using Google Translate or other web page translators would also translate the code and the grammar, making it unparsable and useless (e.g. variable names might be split into multiple words). With this PR, those sections are marked 'notranslate'. Furthermore a 'Translate page' pulldown is added, allowing the user to read the user manual in any language.


TimLethbridge:
I am going to merge this. The reason why the push build failed is because ManualImprovement doesn't have the updates to UmpleToRTCpp/UmpleTLTemplates files, yet it is using a first-build compiler that expects those refactorings done in issue988-b

The Apprveyor problem is another issue, related to environment updates with ant

Issue #1254, opened 2018-03-28
Some test cases in testbed are never executed


BernardYuan:
In testbed, test cases from package **cruise.association.specializations** and **cruise.association.compositions** seem to have no corresponding JUnit tests. Therefore, it is doubted that those test cases are only compiled, but never executed. Therefore, some runtime errors from the code generated from Umple test cases will not be captured.

BernardYuan:
@TimLethbridge This is the testbed issue I have mentioned in our weekly meeting.

BernardYuan:
Fix this after fixing issue #988 , otherwise lots of testcases can be expected to fail.

RalphNgassa:
There are indeed no existing unit test for **cruise.association.specializations** and **cruise.association.compositions** as it was mentionned. But going through the testbed source files I noticed a comment present in **TestHarnessHarnessAssociationsSpecialization.ump** saying that if cruise.association.specializations test cases compiled then it means they were successful.  I believe this explains why there are no corresponding junit test.

Unfortunately, there no similar comments present in the compositions test source files but that might explain why there are no corresponding compositions junit tests.

The comment:

![image](https://user-images.githubusercontent.com/71353825/124293465-b52f1780-db24-11eb-8d77-326d1600736d.png)

TimLethbridge:
Was closed by #1796

PR #1255, opened 2018-04-03
isse#988 - fixed some templates

BernardYuan:
Fixed some templates concerning [Issue988](https://github.com/umple/umple/issues/988), whenever I know how to trigger them.

The test cases are already there. Just that there are no JUnit tests running them, as mentioned in [issue1254](https://github.com/umple/umple/issues/1254).


BernardYuan:
Some test failed because there are some test cases in *.txt files requires updating.

BernardYuan:
As [PR 1258](https://github.com/umple/umple/pull/1258) is merged, now these tmeplates are only instantiated in the correct places. So this pull request no longer makes any sense and can be closed.

Issue #1153, opened 2017-11-19
After code injection injects code after any line that has "return" in it


JoshuaMcManus:
## Brief Description

When using the *after* statement to inject code into an application, it is supposed to inject the code before any *return* statement. The program will instead look for any line that has "return" in it and inject the code before it.

## Minimum Steps to Reproduce

Create an umple file with the following code:

```
class ReturnIssues
{
  String foo(int a) {

    if(a == 3) {
    	System.out.println("return");
        return "3";
    }
    
    // inject before return 
    return "";
  }

  after custom foo(int) {
    String b = "1";
    System.out.println("Returning from foo, a: " + String.valueOf(a) + ", b: " + b);
  }
}
```

## Expected Result

```
// line 6 "../test.ump"
  public String foo(int a){
    if(a == 3) {
        System.out.println("return");        
        // line 16 "../test.ump"
        String b = "1";
        System.out.println("Returning from foo, a: " + String.valueOf(a) + ", b: " + b);
        // END OF UMPLE AFTER INJECTION
        return "3";
    }
    
    // inject before return    
    // line 16 "../test.ump"
    String b = "1";
    System.out.println("Returning from foo, a: " + String.valueOf(a) + ", b: " + b);
    // END OF UMPLE AFTER INJECTION
    return "";
  }
```

## Actual Result

```
// line 6 "../test.ump"
  public String foo(int a){
    if(a == 3) {
    System.out.println("      
      // line 16 "../test.ump"
      String b = "1";
          System.out.println("Returning from foo, a: " + String.valueOf(a) + ", b: " + b);
      // END OF UMPLE AFTER INJECTION
      return ");        
        // line 16 "../test.ump"
        String b = "1";
            System.out.println("Returning from foo, a: " + String.valueOf(a) + ", b: " + b);
        // END OF UMPLE AFTER INJECTION
        return "3";
    }
    
   // inject before      
      // line 16 "../test.ump"
      String b = "1";
          System.out.println("Returning from foo, a: " + String.valueOf(a) + ", b: " + b);
      // END OF UMPLE AFTER INJECTION
      return    
    // line 16 "../test.ump"
    String b = "1";
        System.out.println("Returning from foo, a: " + String.valueOf(a) + ", b: " + b);
    // END OF UMPLE AFTER INJECTION
    return "";

  }
```

TimLethbridge:
If a comment has return as the first word, there is a crash

PR #1257, opened 2018-04-09
Fix issue #1153

hugessen:
This PR fixes issue #1153, where any mention of the word "return" would trigger after behaviour. Even when the word 'return' was in quotes, a comment, or was part of a longer word (eg 'returning').


TimLethbridge:
This looks good, but it needs tests that cover the various cases!

TimLethbridge:
@hugessen  -- the fix for this PR is simple: You need to update the Ruby and Php versions of the generated code too.

PR #1262, opened 2018-04-23
Partly fix for issue 1260 - added specializedFromVariable to association variable

BernardYuan:
When _relatedSpecialization templates are triggered, some non-existing variables in a class are accessed. Therefore, for this case, we need to know from which association variable is one association variable specialized so that we can call their common code. Templates will be modified in next PR.


TimLethbridge:
Great. The single Appveyor fail is the regular problem we face (issue #831)

PR #1269, opened 2018-05-24
Fixes #1268 Manual pages not generating previous link

### LINKED ISSUE #1268: No previous page link generated for several manual pages
All manual pages that follow a section composed of only one manual page do not contain a link to the previous manual page.
The issue affects the following pages: Association Definition, Singleton Pattern and Umple messages.

The common link between the three pages that are missing the previous link might be something else, but the HTML formatting of these pages seems correct.

ZainabAlShowely:
The Documenter would not generate a previous link if there was was <= 1 page of content in the previous section of a page, and would therefore not generate previous links for some pages. This PR simply changes the condition in which a previous page is generated.


TimLethbridge:
Great this works.

PR #1281, opened 2018-06-01
Fixes #1277 Generating E3504 when referencing a non-existent template

### LINKED ISSUE #1277: Compiler error (analysis) when referencing a non-existing template
When referencing a template that cannot be found, E9100 is thrown instead of E3504 (template reference cannot be resolved).

The following generates the error:
~~~Java
class A {
  temp <<! <<@nonExistingTemplate>> !>>
  emit method()(temp);
}
~~~

A NullPointerException is thrown at the recursive cycle check level.
The issue does not occur if the template is defined but not used in an emit statement.

ZainabAlShowely:
Added a check for the template reference value, an inexistent template no longer
generates E9100. Added manual page for E3504.

### DIFF
--- a/build/reference/9354TemplateReferenceCannotBeResolved.txt
+++ b/build/reference/9354TemplateReferenceCannotBeResolved.txt
@@ -1,5 +1,5 @@
 E3504 Template Reference Cannot Be Resolved
-Errors and Warnings
+Errors and Warnings 1000+
 noreferences
 
 @@description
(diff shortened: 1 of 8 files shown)


TimLethbridge (changes requested):
Please could you add a testcase for E3504 to this pull request. In general we should have a testcase for every error that uses the assertFailedParse assertion. I know that some of them don't, but we ought not to let any new PRs pass that omit such tests.

ZainabAlShowely:
Added two test cases. Was unsure of the format (which prefix number to add for ex) to use for the files, however, was it correct?

TimLethbridge (approved):

PR #1290, opened 2018-06-05
Mixset Parser & Use Statement 

gublan24:
This PR has several pieces of code that can be summarized by the two points:

1- Allows Umple to parse mixsets partially. Umple adds mixsets to the umple model with the proper line numbers and UmpleFiles. But still there is no parsing for the body of mixsets.

2- Handles use statement that does not match any umple file. Umple will save all potential mixsets use statements in arraylist called "possibleMixsetList" in Umple model. Then at last step, the possible mixset use statements should be handled.
TO DO : Umple should fail when a use statement does not either match a file, or mixset.

** The work on mixsets is still on progress.


cruise.umple/src/UmpleInternalParser_CodeMixset.ump
```diff
@@ -0,0 +1,178 @@
+/*
+
+Copyright: All contributers to the Umple Project
+
+This file is made available subject to the open source license found at:
+http://umple.org/license
+
+This file analyze core umple meta model tokens such as which language to generate,
```
TimLethbridge: This comment is wrong

cruise.umple/src/UmpleInternalParser_CodeMixset.ump
```diff
@@ -0,0 +1,178 @@
...
+
+class UmpleModel{
+  depend java.util.stream.*;
+
+/*
+ * returns a mixset from UmpleModel based on its name.
+ * @Abdulaziz   
+ */
+ Mixset getMixset(String name) {
+	Mixset mixset = null;
```
TimLethbridge: Please use two spaces throughout Umple, never tabs. You may find tabs in some places but that is not our standard.

cruise.umple/src/UmpleInternalParser_CodeMixset.ump
```diff
@@ -0,0 +1,178 @@
...
+*/
+
+
+namespace cruise.umple.compiler;
+
+class UmpleModel{
+  depend java.util.stream.*;
+
+/*
+ * returns a mixset from UmpleModel based on its name.
```
TimLethbridge: Point out in the comment that it returns null, if no mixset matching the name is found

cruise.umple/src/UmpleInternalParser_CodeMixset.ump
```diff
@@ -0,0 +1,178 @@
...
+  	aToken.addSubToken(new Token("entityName",className) );
+	aToken.addSubToken(new Token("entityType","class") );
+	analyzeMixset(aToken);		
+	}
+
+/*
+ * last action in the parse.	
+ */	
+ private void parseMixset(){
+ 	for (Mixset mixsetUse : model.getPossibleMixsetList())
```
TimLethbridge: This loop seems excessively inefficient. All that we need to do is to process each element of the mixsetOrFile Association and it it is a mixset and their are fragments not already used, then process it. What you are doing is making a list and then searching the model for an element you have already found.

cruise.umple/src/UmpleInternalParser_CodeMixset.ump
```diff
@@ -0,0 +1,178 @@
...
+
+/*
+ * returns a mixset from UmpleModel based on its name.
+ * @Abdulaziz   
+ */
+ Mixset getMixset(String name) {
+	Mixset mixset = null;
+	for(MixsetOrFile mOrF : getMixsetsOrFiles(name))
+	{
+		if (mOrF instanceof Mixset)
```
TimLethbridge: Best practice would be to have a Boolean something like 

```
Boolean isMixset = false;
```

is MixsetOrFile

and override the constructor in Mixset using after constructor to set the Boolean to true. Then you can simply say here mOrF.isMixset()

cruise.umple/src/UmpleInternalParser_CodeParserHandlers.ump
```diff
@@ -21,6 +21,14 @@ external interface AnalyzerGeneratorHandler{}
  * dispatches a `RuleBasedParserThread` on the discovered token, which
  * effectively reads the new file into the token location
  */
+ 
+ 
+ class UmpleModel{
+   
+   lazy ArrayList<Mixset> possibleMixsetList = new ArrayList<Mixset>();
```
TimLethbridge: This breaks Umple standards. We should be only using associations for lists. And anyway, we don't need this. The existing model I defined has all that is needed. Please remove.

cruise.umple/src/UmpleInternalParser_CodeParserHandlers.ump
```diff
@@ -29,39 +37,69 @@ class UseStatementParserAction{
...
+	String path = "";    
     int index =-1;
     if (data!=null && data.getFullFileAddress()!=null)    index = data.getFullFileAddress().lastIndexOf("/") ;
     if (index==-1) index = data.getFullFileAddress().lastIndexOf("\\");  
     if (index!=-1){
     	path = data.getFullFileAddress().substring(0,index+1 );
     }
     
-    synchronized(data.getHasParsed())
+synchronized(data.getHasParsed())
```
TimLethbridge: No need to take out the indentation here

cruise.umple/src/UmpleInternalParser_CodeParserHandlers.ump
```diff
@@ -29,39 +37,69 @@ class UseStatementParserAction{
...
       fileName = data.getAnalyzer().getFile().getAbsoluteFile().getParentFile().getAbsolutePath() + File.separator + value;
       //I used this technique because I noticed this function is used to several purposes.
       // The correct implementation is just to set fileName = path +value; and remove the above code.
       //TODO: this issue must be investigated more. Currently it is working with the following solution. 
       File f = new File(fileName);
       if (!f.exists()) fileName = path +value;
-      if(!data.getHasParsed().contains(f.getAbsolutePath()))
+     
+     /*  
+      *	when a file is not exist, consider it as a possible mixset.
```
TimLethbridge: I would say definitively: Any thing ending with '.ump' must be a file. Anything not ending with .ump must be a mixset.

cruise.umple/src/UmpleInternalParser_CodeParserHandlers.ump
```diff
@@ -29,39 +37,69 @@ class UseStatementParserAction{
...
+      	int useLineNumber = token.getPosition().getLineNumber();
+      	UmpleFile mixsetUseFile = umpleModel.getUmpleFile(); // the file in which "use" is mentioned. 
+      	Mixset tempMixset = new Mixset(mixsetName);
+      	tempMixset.setUseUmpleFile(mixsetUseFile);
+      	tempMixset.setUseUmpleLine(useLineNumber);
+      	umpleModel.getPossibleMixsetList().add(tempMixset);
+      	
       }
+      else   {  
+	      if(!data.getHasParsed().contains(f.getAbsolutePath()))
```
TimLethbridge: My design intent was this:

The use statement would immediately parse the mixset fragments that have been accumulated so far using the same code as below (modified as needed so it can operate on a mixset, not a file). Later on, if additional mixset logic was found, it would be parsed when found, again using the same logic as below.

TimLethbridge (changes requested):
I have made quite a few comments. Clearly a lot of work needs doing before this is merged. We will talk tomorrow.

cruise.umple/src/UmpleInternalParser_CodeMixset.ump
gublan24: Yes.

cruise.umple/src/UmpleInternalParser_CodeMixset.ump
gublan24: OK.

cruise.umple/src/UmpleInternalParser_CodeMixset.ump
gublan24: OK.

cruise.umple/src/UmpleInternalParser_CodeMixset.ump
gublan24: I agree.

cruise.umple/src/UmpleInternalParser_CodeParserHandlers.ump
gublan24: This happens because I added some code, then I removed it.

cruise.umple/src/UmpleInternalParser_CodeParserHandlers.ump
gublan24: This will make the work much easy.

gublan24:
I've updated the branch with the suggested modifications. I believe that the build checks should pass with no failure.

TimLethbridge (approved):

Issue #1299, opened 2018-06-13
Some diagram edits leave the diagram 'grey'


TimLethbridge:
If in Umpleonline one causes an error while editing this results in the diagram appearing grey. However, after fixing the error, the diagram is not always being returned to its regular 'black' state even theough it is 'active'.

TimLethbridge:
The issue is more than 'diagram staying grey'. It seems that when the diagram is in the grey state after correcting and error there are two other consquences:

1. Classes can be moved around, but the associations don't 'follow' them.
2. An attempt to connect an association to another class doesn't work, as you can't click on the destination class and the 'rubber band' association continues until you click on the same class (making a reflexive association) or you hit escape.

TimLethbridge:
Making a textual edit can change 'grey' to 'black' But it doesn't solve the problem mentioned in the last comment!

TimLethbridge:
If I am in the 'grey state' and I add two classes, I can then create an association between them. But the existing classes on the diagram seem unrecognized by the linking code, even though on can drag them. They must be somehow omitted from some kind of list.

PR #1302, opened 2018-06-18
Fixes #1298 #1299 Adds compilation check after diagram edit in UmpleOnline and fixes issues relating to grayed out state

### LINKED ISSUE #1298: Missing immediate validation for graphical edits in UmpleOnline
When adding new attributes to classes in the Umpleonline editable diagram, several checks are missing, causing errors as the code will no longer reflect the diagram content.
An error message indicating that the Umple code is unable to be updated is added to the text editor multiple times, but the expected behavior is not allowing adding the attribute (as with non-alphanumerical names).

Duplicate attributes name and attributes names conflicting with association names or state machines cause the issue.

Other checks may need to be added to method names as well.

ZainabAlShowely:
Adds missing compilation check after a diagram edit in UmpleOnline.


ZainabAlShowely:
Added a partial fix for 1299 to this PR to test the diagram being grayed out after a diagram edit, by using a different method to add the 'diagram sync needed' message.
The grayed out state has a remaining issue, present in the live version: Removing attributes/methods is possible in that state (but it does not update the code).

PR #1313, opened 2018-06-22
Fixes #884 Adds annotations that pass through from Umple to generated code

### LINKED ISSUE #884: Umple parsing and output of annotations needs some more work
Consider the following example:

```
class X{

 @SuppressWarnings("unchecked")
  public void analyzeToken(Token token)
  {
    analyze(token);
  }
public void analyze(Token token){}

public void analyze2(Token token){}
}
```

If we generate the code we won't see the method `analyze` while it has got the method `analyze2`.
The issue comes to the place because of `@SuppressWarnings`. I think it should be related to the Umple grammar.

TimLethbridge:
This PR allows all Java annotations starting with @ to be passed through to generated code if they are present in Umple source. When generating php and Ruby they appear as comments. Users can use annotations on methods in order to use Umple with tools that process annotations, or to mark a method as @Deprecated etc. Annotations on classes also appear before the generated class. Annotations on attributes appear before the get method. Annotations on internal attributes appear before the private variable.

An annotation can appear in Umple anywhere a comment can appear. In fact, they are treated just like comments, except that in Java they are output without being wrapped in a comment.

This feature has been requested by several people in the end-user community.

### DIFF
--- a/UmpleToPhp/UmpleTLTemplates/class_MethodDeclaration.ump
+++ b/UmpleToPhp/UmpleTLTemplates/class_MethodDeclaration.ump
@@ -59,7 +59,7 @@ class UmpleToPhp {
         }
         
 				appendln(realSb, "");
-				if (aMethod.numberOfComments() > 0) { append(realSb, "\n  {0}\n", Comment.format("Method Javadoc",aMethod.getComments())); }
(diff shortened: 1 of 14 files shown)


TimLethbridge:
Failures were all related to DoActivityTest, so all is good to merge

PR #1366, opened 2018-09-18
Fixes #460 - Add support for nested generics

### LINKED ISSUE #460: Error in attributes with nested generic type
Originally reported on Google Code with ID 562

---

class A {
  depend java.util.*;
  Map<String, String> a;
  Map<String, List<String>> b1;
  Map<String, List<List<String>>> b2;
  Map<List<String>, String> c1;
  Map<List<String>, List<String>> c2;
  Map<List<Map<String, String>>, String> c3;
  Map<List<Map<String, String>>, List<String>> c4;
}

**What is the expected output? What do you see instead?**

c1 to c4 cannot be processed while it is possible to be fixed.

---

Reported by @nwoedf on 2014-05-11 15:54:08

pwang347:
## Description
The current system does not support (all) nested generics, e.g. `Map<List<String>, String>`, though `Map<String, List<String>>` was already supported.

This change aims to ensure that the behaviour is consistent across most cases.

Further note that array conversions will only occur at the outermost layers; for instance, `ArrayList<String[]>[]` generates `List<ArrayList<String[]>>` for Java.


TimLethbridge (changes requested):
Great work.

You will need to figure out why the build is failing and push relevant additional changes to the branch, so they appear in this PR.  Other than that, what you are doing looks reasonable. I am curious as to why the Map<String,String> wasn't causing any problems before.

As for your question about validation: Umple doesn't verify the existence or correctness of user-specified types, as that would be too complex. Such problems would need to be detected when the java is subsequently compiled.

TimLethbridge:
Appveyor is giving this, but wasn't before. Not clear why the error occurs. This same code seems to compile on the other CI platforms and locally.

```
   [javac] C:\projects\umple\cruise.umple\src-gen-umple\cruise\umple\compiler\StructureDiagramGenerator.java:586: error: method _createJavaScriptCall in class StructureDiagramGenerator cannot be applied to given types;
   [javac]     return this._createJavaScriptCall(0,sb,modelName,bndList).toString(); 
   [javac]                ^
   [javac]   required: Integer,StringBuilder,String,List<ComponentDescriptor>,List<BindingDescriptor>
   [javac]   found: int,StringBuilder,String,List<BindingDescriptor>
```

The relevant line is supposed to be generated as

```
return this._createJavaScriptCall(0,sb,modelName,cmpList,bndList).toString(); 
```

The cmpList list seems to be being left out when it wasn't before.

TimLethbridge:
So appveyor is compiling the PR twice. The first compile (using the previous compiler) is OK. The second compile (using the just-compiled-compiler) fails. This suggests that the changes in the PR are at fault and it is not a configuration issue in Appveyor.

TimLethbridge:
@pwang347 I can confirm that the Appveyor errors are a result of this PR and not a configuration problem. Appveyor does two compiler builds in succession, and the second one fails. In your branch on my machine I can replicate that by simply doing who quick builds in succession. See my other comments above. But the problem does not occur in master.

pwang347:
@TimLethbridge Thanks for this! Just curious, are the build definitions in source control as well? I can see the script used for TravisCI but not for Appveyor it seems.

TimLethbridge:
@yes, The build definitions are in source control. The appveyor one is appveyor.yml at the root

TimLethbridge:
Great job

PR #1378, opened 2018-09-26
Fixes #1362 - Do not generate error markup by default in UmpleOnline

### LINKED ISSUE #1362: UmpleOnline keeps certain generated outputs when generating others
If you generate certain outputs (that appear at the bottom of the screen) in UmpleOnline, previous outputs are not replaced. This happens in certain combinations only. The person who solves this issue should search for other cases.

Known cases:

ER Diagram followed by Event sequence
ER Diagram followed by CRUD User Interface

pwang347:
## Context
Generated code that does not have ```<p>URL_SPLIT``` within the markup is automatically treated as error code, and dumped to the messageArea div. For instance, ER diagram does not include this tag and so the generated diagram is always treated as an error output.

This change follows the philosophy that content lacking this delimiter should still be treated as valid code instead of error output.

## Tests performed
* ER Diagram followed by Event sequence - no longer overlapping
* ER Diagram followed by CRUD User Interface - no longer overlapping

## For followup
* Details on how to test for the following comment in #1362:
> ... after generating valid output, and the valid output is not removed


pwang347:
TODO: test that error outputs of previous code generations are cleared

TimLethbridge (approved):
I have tested this and it seems to solve the problem. I will merge

PR #1385, opened 2018-10-09
Fixes #1351 - Add ability to declare state-dependent method bodies

### LINKED ISSUE #1351: State-dependent methods / internal events.
Sometimes a user wants to create behaviour that varies based on state. But if the method has side-effects one needs to set aside a variable to store the result as events in umple currently don't have return types.

```
class Portal {
  statusWord = "";
  description = {"Current status: "+getStatusWord();}
  before getStatusWord {describe();}

  sm {
    closed {
      describe / {setStatusWord("closed");} -> closed;
      open -> open;
    }
    open {
      describe / {setStatusWord("open");} -> open;
      shut -> closed;
    }
  }
}

```

The above is wordy and the user must explicitly code the self transition that might clutter up the diagram, might be error prone, and is awkward. It would be better to be able to define a method in a state thus:

```
class Portal {
  description = {"Current status: "+getStatusWord();}

  sm {
    closed {
      String getStatusWord() {return ("closed");}
      open -> open;
    }
    open {
      String getStatusWord() {return ("open");}
      shut -> closed;
    }
  }
}
```

Any Umple method syntax should work.

The generated code would generate a single method getStatusWord with a case statement that would execute the right block of code depending on the state (with priority given to substate versions overriding higher level state versions. If no method version is defined in a state then a default of doing nothing and returning an empty string, null, zero, false or nothing, depending on the method;s return type.

The above would serve to implement UML's internal event capability.

In a queued or pooled state machine, should such a method be in the queue or be a direct call? I tend to think the former.

The above could be done by 
a) Allowing any event method to return a value initiaally (with consistency enforcement, and default enforcement)
b) Adding the ability to tag any self transition as internal e.g. ~>
c) Allowing an ordinary method signature to implicitly define an internal transition.

pwang347:
## Context
The goal of this pull request is to add the ability to declare method bodies that depend on state.

## Tests performed (within test file)
* Order of method declaration (default methods should be merged regardless of whether they are declared before or after a state-dependent method is declared)
* Method distinguishing via parameters
* Nested states (tested only 2-deep)

## TODO
* Test that different modifiers don't result in strange behaviour (currently, will take the first encountered modifier)
* Test more complicated state nesting
* Test multiple top-level state machines
* There is a weird extra space added for the second state-dependent method in the test file which needs to be removed


cruise.umple/test/cruise/umple/compiler/JavaGeneratorTest.java
```diff
@@ -1325,6 +1325,7 @@ public void constraintComparisonWithConstant()
     String actual = model.getGeneratedCode().get("X");
     File generatedFile = new File(pathToInput+"X.java");
     generatedFile.delete();
+    UmpleModel test = model;
```
pwang347: Oops, will remove this

UmpleToJava/UmpleTLTemplates/state_machine_StateDependentMethods_All.ump
```diff
@@ -0,0 +1,143 @@
...
+        // Prints the state-dependent method as defined within the UmpleClass
+        public void traverse(Method method)
+        {
+          printBefore(method);
+          Map<State, MethodBody> stateMap = uClass.getStateDependentMethodBodyMap().get(method);
+          MethodBody defaultBody = stateMap.getOrDefault(null, null);
+          for (StateMachine sm : uClass.getStateMachines())
+          {
+            StringBuilder smBuilder = new StringBuilder();
+            append(smBuilder, "\n  {");
```
pwang347, line 64: This needs to be fixed, should be moved out of loop otherwise will fail with multiple sms

TimLethbridge:
@pwang347 Fantastic job! I am going to merge this.

However, the task isn't finished as you mention yourself. There is a to do list.

I am going to add that tie #1351 and reopen that issue (it will be temporarily closed by this).

PR #1391, opened 2018-10-15
Fixes #1351 - State-dependent method bodies part 2

### LINKED ISSUE #1351: State-dependent methods / internal events.

pwang347:
## Context
Completes remaining tasks for #1351.
* Different modifiers now generates warning 82 (added associated manual files)
* Top-level state machines now enabled
* Fixed bug with extra space (space is now trimmed in modifiers)

## Tests performed
* Tested more complicated state nesting
* Tested multiple top-level state machines
* Tested the new warning generation


cruise.umple/test/cruise/umple/statemachine/implementation/java/stateDependentMethodDeclaration_2.java.txt
```diff
@@ -0,0 +1,237 @@
...
+  {
+    return sm;
+  }
+
+  public void pstatus()
+  {
+    switch(sm)
+    {
+      case init:
+        System.out.println("init");
```
pwang347, line 47: Would probably be a good idea to test injections for methods with return values as well

TimLethbridge:
@pwang347 Are you able to look into the conflicting file UmpleParserTest.java that was also, I think modified recently by another PR. You simply need to merge your changes into what is current. I am not sure why it can't happen automatically.

cruise.umple/test/cruise/umple/statemachine/implementation/java/stateDependentMethodDeclaration_2.java.txt
TimLethbridge: Yes,I agee

build/reference/9082ConflictingMethodModifers.txt
```diff
@@ -0,0 +1,16 @@
+W082 Conflicting Method Modifiers
+Errors and Warnings 51-99
+noreferences
+
+@@description
+
+<h2>Umple sematic warning when do activities, exit actions, outgoing transitions, and/or nested state machines are removed by the compiler from final states.</h2>
```
TimLethbridge: This block is leftover from error 72 and is not correct in this file. Please change so it is correct.

TimLethbridge (changes requested):
Please resolve the comment I left, and also deal with the merge conflict

TimLethbridge (approved):

TimLethbridge:
Great job!

PR #1398, opened 2018-10-23
State-dependent methods manual page and minor fix

pwang347:
## Context
* Added manual page with examples
* Fixes case where injections weren't working with void methods (and to guard against strange fall-down behaviour with switch cases that lack break and return)
* Fixes return adding an extra space for void types in the default case


TimLethbridge:
Brilliant! I love the dining philosophers example

PR #1402, opened 2018-10-26
Fixes #225 - Add tabs to UmpleOnline

### LINKED ISSUE #225: Allow multiple files umple in UmpleOnline using tabs.
Originally reported on Google Code with ID 326

---

Currently UmpleOnline operates on a single model.ump file when it is editing.

This project would do the following:
1. If there are additional .ump files in the directory, it would create tabs in Umple
   Online. Clicking on a tab would load the file into model.ump (for compatibility) and
   display it. Saving model.ump would save the file also to the appropriate named file.
2. There would be the ability to add a tab, and tabs would be renewable, although the
   .ump extension would be automatically part of the name of the underlying file.
3. Code generation would compile the current file, but use clauses would be used to
   enable compiling of all the files. See also issue 325 for the ability to use files
   in standard libraries.
4. It should be possible to export all the ump files by downloading as a zip file,
   as can be done with generated code. Similarly the ability to upload should be possible
   for a set of files.
5. The diagram displayed would always be the one associated with the current file,
   although a mechanism could be incorporated to allow for use of different diagram files
   with the same model file.
6. The entire Umple system compiler should be able to be displayed in this manner in
   UmpleOnline (although don't copy the files).

---

Reported by @umple on 2012-08-16 17:31:21

pwang347:
## Context
* Added tabbing; supports creating new tab, double-clicking tab name to rename, deletion, download all .ump files, server-side reordering via tab_index file
* Made sure that bookmarking works as intended (new files are copied over and tabs are created in the new URL)

image

## For discussion
* After playing around with generating `use` statements, it seemed like the least obtrusive way was to simply let the user type it themselves. This avoids the issue of dropping markers to actually correctly update the statements, but there are also issues if the user explicitly changes these or if we want to avoid cyclic dependencies.
* How it works - I decided to treat model.ump as the "viewing frame"; ie. tabs are loaded into model.ump and navigation events trigger saves. For improved performance, only saves are sent to the server while reads will read directly from local storage (except for when loading from a bookmark, in which case all files need to be loaded).
* This means that generation always only generates model.ump as usual (the current viewing frame); this is a consequence of use statements needing to be explicit.
* As part of the change for downloading all .ump files (except model.ump, we just save the current tab before downloading), I added a new dependency on php-zip.
* I am also a bit concerned about potential race conditions--I'm more familiar with the async-await programming paradigm and not having a callback to handle error cases (as far as I'm aware callbacks are just dropped for errors) seemed a bit problematic. Initial testings of each feature seemed to be working, but I'm sure that with some latency the client and server can go out of sync and do weird things.

## TODO (Not meant for this PR)
* It would be nice if the toggles and dropdown states are also kept on a per-tab basis, but possibly better as a subsequent task
* Add hotkeys for tab navigation


TimLethbridge:
Great work! Works so far. I will put my comments as a review.

TimLethbridge (changes requested):
Here are my comments so far:

1. I agree with your viewing frame approach. And having user put in use statements by themselves. It seems to work well

2. I am concerned about screen real estate for small devices, both vertical space, and also space in the upper message line. Could you make the following changes. a) With your changes the error messages are default off the bottom of the screen, that needs correcting so there is still space to see them without scrolling. b) Is there a way to reduce the amount of vertical space used by the tabs (less vertical space in the tabs above and below the labels; perhaps smaller font; less veritcal space above the tabs; c) Add to the options menu 'show/hide tabs', so people working on small devices have this option; d) Add 'Download Umple Files' to the Save menu and consider removing from the icon line.

3. The 'Download files' doesn't seem to work, at least when running in Docker. Seems just to open a new blank Tab

4. When a bookmarked URL is reloaded, the tabs are in alphabetical order. This might be the best choice, but it will be different most likely from what the user expects.

5. There seems to be a limit of 11 tabs, which is reasonable I suppose, but the '+' then does nothing.

6. It would be great if when loading an example, the tab could automatically take on the name of the example if it had previously started with 'Untitled_'. Same when loading a file or example as a URL option.

7. Speaking of URL options, the following did not work in Docker 'http://localhost:8000/umple.php?example=manualexamples/HelloWorldExamples1'. This is used to load manual examples and is very important. Similarly, loading ordinary examples doesn't work, such as 'http://localhost:8000/umple.php?example=AirlineSystem'. This is critical. Check loading files as well (there is an option to load any ump file stored on the internet that has a .ump suffix)./

8. When I make a persistent URL, create a new tab and typed into it. Then reload the URL. What I typed had not been saved in the new tab. But if I load an example from the menu, all seems OK.

9. One time I loaded an example fro, a menu into a tab, and then hit + to create a new tab, and the example appeared there as well

10. A couple of times when loading examples, into a new tab, I got into a freeze state.

11. Please try every single option in the menu, generate all outputs, do all saving and loading methods, edit diagrams manually in multiple tabs, etc.

Once the above are all dealt with, I will get other UCOSP students to test too using Docker and also using a branch checkout. This needs to be very thoroughly checked before merging as it is a fundamental change. I might make a new minor release with this when I do merge it.

TimLethbridge:
I notice also that your branch is behind master. That means that certain recent changes to the way Appveyor behaves are not being accounted for.

pwang347:
@TimLethbridge Thanks for this!
For your comments above:
1. Sounds great.
2. Will work on this! I'll test reducing the font size (and possibly adding the file extension to be more explicit for `use` statements). I'll also test just reducing the padding around the text, since it's important that the text is still clickable for renames.
3. I believe this should be resolved by adding php7-zip to Dockerfile-base, will add this soon.
4. I'll also work on this. Could be possibly done by supplying an ordered list of names in the query and the server doing its best to accommodate for it.
5. I might adjust this cap to account for point 1
6. Yes, that's a great idea--I'm thinking of adding a flag that indicates that the name is ephemeral.
7. Will investigate this! Thanks
8. Yes, this is an issue: currently saves to remote are only triggered by actions like tabbing, bookmarking and downloading. Maybe periodic save requests should happen for the currently active tab as well so changes aren't lost. I'm also thinking of adding a "dirty" flag if there are changes and if so confirm navigates away from the web editor.
9 + 10. Most likely client-server went out of sync. I think the freeze happens whenever there is a bad request code, so maybe instead I should return an error message with status 200 and work with that in the callback.
My plan is to make the server less of an "on-demand" service (latency is too high to block the user interface on remote calls) but rather to queue up events and replay them as a state-tracking entity. If necessary, we could even have the client send the entire tab state over periodically to re-sync the server in the worst case.
11 + 12 (master branch). Will do!

Thanks for the very insightful feedback!

TimLethbridge:
Not 100% sure what is going on with issue 8. In the old UmpleOnline saves were done from the browser to model.ump whenever typing occurred (within seconds), and in other contexts such as editing diagram. Now, as I understand it, it has to save to both the named file and also model.ump, but I would have thought this could be essentially simultaneous. I don't see why it has to be as complex as it sounds as though you think it might need to be.

pwang347:
Yes, it should work as a simultaneous update. The forced saves are intended to slightly reduce the delay from autosaving, though a more robust solution may be to actually include everything in the payload rather than hoping the request completed. Dirty state now that I think of it wouldn't work too well with autosaving in general, but was something I was thinking about.

pwang347:
For 9+10 I was exploring the idea of using a queue and workers setup for the server, but I assume that introducing separate worker threads is not something we want. Currently considering some sort of client queue in favour of a heftier sync operation for tab state. Certain operations such as bookmarking or downloading would then block (display progress spinner) until all the requests complete

TimLethbridge:
I think we need to think simplification. Rather than 'queue and workers'. First there are two servers. Here we are talking about the php backend (not the internal server that that calls). I just want to be clear.

I am happy you are using local storage to prevent constant reloading of tabs as the user clicks back and forth.

But I thing that for the saving to the server, you should use symlinks. Essentially:
- When activating/loading/whatever a tab, a symlink is made from that file to model.ump (replacing any existing symlink)
- From that point on, all operations that operate on model.ump would work as they always have with no need for any change whatsoever. 
- All edits always write through instantly to model.ump (and the relevant tab file) as they have always done, so there is no need for any queuing, and there shouldn't be any races.

pwang347:
I did actually consider using a symlink initially, and if I understood correctly you mean that model.ump would point to the file for the currently active tab, so saves to model.ump save directly to the active tab.

My biggest concern here was that the updates to model.ump happen periodically (as you mentioned) and if the symlinks are not updated properly we risk writing the current changes to the wrong tab.

I tend to think of the (PHP) server operations as an untampered replay of the client operations. This sequence of operations naturally lags behind the client due to network latency (and is not guaranteed to receive everything). Fundamentally, the server just needs to make sure it has the correct file system structure for the following tasks:
* `use` statements
* downloading the .ump files and
* bookmarking

[Client-dedicated] queue and worker perhaps wasn't the best description or model to use, but clients must be able to order the operations so that deletions, renames followed by data writes don't fail. And following this line of thought I think symlinking may be violating the atomic nature of these operations (that is, the server's knowledge of what is the "active tab" is not a source of truth, but the client's is).

Given the freedom for tab naming as well, I've been hesitant to create any form of intermediate symlinks in the directory (for example, I considered using a symlink for renames so payloads directed to the old name could also redirect) but there seemed to be too much maintenance needed (e.g. name clashes) for marginal utility.

So with the above, my conclusion to address 9+10 was to either replicate each operation on the server (which may be somewhat brittle, but is certainly more lightweight), or we could send larger sync payloads which contain essentially all the information the server needs to know (leading to redundant work). The latter is probably easier to implement (can use existing ajax callback setup) and handle edge-cases with, I'm just not sure how slow it will be. A combination of the two approaches may also be interesting, e.g. accumulating diffs and sending them over in bundles so the server can replay them.

Point 8 as mentioned earlier should simply be to update the autosave hook to include an extra parameter with the ID of the current tab, and using this to duplicate the changes over as well. The local cache would not need to be updated since it's updated on tabbing away and reloads will fetch the information away anyways.

Please let me know what you think!

pwang347:
Alternative solution to height issue - display error/warning icon at top (where it says see message)?
Another issue: history - these should also be tab specific.

TimLethbridge:
Good steady progress.

We will need to fix the Dockerfile issue so we can test downloading properly in Docker.

Here are some problems I note (continuing numbering sequence)

10. Sometimes (often), clicking on the + does nothing. Sometimes clicking on a tab activates a different tab. Sometimes, clicking on the x to delete, deletes the wrong tab! I don't know how it got into this state. If I closed completely, and reloaded the URL it went away. But we need to keep testing until this can be reproduced.

11. If I double-click to edit a name, it would be better if it would select the entire name, at least if it is 'Untitled-n'. Otherwise, the user has to click again, or if they just type then they get the previous name at the end.

12. Critical: If I load an example, and start editing it. The cursor suddenly jumps to the top of the text pane. This was an old problem we got rid of long ago.

pwang347:
Description of latest changes:
* Removed unnecessary blocking while things are loading - if the tab context has changed then some callbacks should simply be omitted.
* Zip is now using the command line utility, so should be testable
* Added ordering via the jQuery Sortable API. The order is then sync'd to the server which to a file tab_index which is an ordered list of names. Subsequent hits to the same model will try to read from an existing index file, otherwise default to original action of finding all .ump files in the dir except model.ump
* Moved download button. Side note: the other features in the same pane only work for the currently active tab, which makes sense for the encoded option I suppose, but others may need to be extended to work with tabs. Shouldn't be too hard considering we're already storing content in localStorage, but I'll defer it for now.
* Request chaining should finally be working as intended now, the ordering of dequeueing was still problematic.
* Tab names are now also sanitized to ensure that they work as filenames, so that we don't have to do odd mappings to and fro to deal with such instances
* Switching tabs now also resets the caret position (also noticed a bug that can be reproduced on live umpleonline: loading examples seems to reset the caret but not the line number text field; I'm experiencing something similar to that in the tabs)

vash-evg:
If I open a new tab and use State machines there, going back to original tab returns states machine representation instead of a class diagram.
Steps to reproduce:
1. Open UmpleOnline
2. Click Examples->class diagram-> 2DShapes. Class diagram representation is displayed on the right
3. Open a new tab
4. Click Examples -> state machines -> Agents communicating. State machine is displayed on the right
5. Go back to old tab. "No state machine found in the umple file" error is displayed. In some cases, state machine representation is displayed instead of a class diagram

pwang347:
@vash-evg Could you elaborate on how to repro it and what was expected (the diagram should update periodically I thought)? Thanks

TimLethbridge:
Regarding @vash-evg comment. Yes, the 'state machine' setting is currently global, and opening a state machine example sets it to 'state machine'. @pwang347 knows about this. We agreed that it would be done as a separate issue later.

TimLethbridge:
I made a minor fix to an obvious bug in line 12 of umple.php that had been there for several months. It had no ill-effect on operations, but was visible in the test server because it is set to display all php notifications.

As it stands I am finding the this PR reasonably stable, although after extensive testing every now and then I am able to get a reload of a saved URL to load code into the wrong tab. But I can't reliably reproduce it. I will keep trying.

pwang347:
Thanks, I will look into and test this particular case. My conjecture for this as of now: the client may be simply triggering saves at inopportune times as the tabs are created.

TimLethbridge:
I have just found that the 'Restore Saved State' doesn't seem to restore it, and sometimes gives the following:

```
<br />
<b>Fatal error</b>:  Call to a member function readData() on a non-object in <b>/h/ralph/umplebranchtest/umple/umpleonline/scripts/compiler.php</b> on line <b>71</b><br />
```

It is supposed to be able to restore whatever was being worked on, so that users who accidentally lose internet connection, or refresh their browser, or simply want to reload, don't lose their work.

YiranShu:
If I create a class and try to generate CRUD user interface, it will fail and display "Notice: Undefined variable: errhtml in /h/ralph/umplebranchtest/umple/umpleonline/scripts/compiler.php on line 208" on the page.

Steps to reproduce:

1. Open UmpleOnline
2. Click Examples->class diagram-> 2DShapes.
3. Click Generate->CRUD User Interface
4. "Notice: Undefined variable: errhtml in /h/ralph/umplebranchtest/umple/umpleonline/scripts/compiler.php on line 208" is displayed.

TimLethbridge:
Thanks @YiranShu -- The problem you encountered is present before these changes. It was simply that the php information messages were being suppressed. Actually this is something you could fix as part of your work.

vash-evg:
Hi, I did more testing and couldn't spot anything new. It looks good to me.

TimLethbridge:
Great. I will do one more round of testing while at the conference, then merge it and do a minor release.

pwang347:
@TimLethbridge Apologies for the inactivity--I can confirm that the 'Restore Saved State' button isn't working because of issues with how cookies are being read, but not the changes of this PR. Deleting the cookie from the normal URL makes it function normally.

The other issue seems to be coming from the loadFile hook kicking in after the tabs are loaded, but I have yet to come up with a clean solution for it. Hopefully I can figure something out tonight.

pwang347:
The issue seems to be no longer reproducible for me. Please let me know if it occurs again!

pwang347:
As followup, something that is a bit cumbersome is the fact that rapid changes followed by refreshes will cause those changes to be lost (the refreshed page loads the old info and then saves it, effectively discarding the changes).

Some ideas to deal with this:
* We could create a lock file in the model directory when we're performing operations on the server side, so that new clients must wait before all the changes have been written. If the URL has saved data, then it makes sense to block until we receive the most recent changes.
* We could try to make use of localStorage and timestamps to determine which change is the most recent; the data received from the server or that which is known locally.
* Probably the cleaner solution: add a file lock when writing, on load when we're reading the contents it should block on the server side.

That being said, it doesn't seem that concurrent editing works for single-tab umple at the moment. So the concern here is less that there are new users joining in, but that accidental refreshes don't lose work. This makes localStorage a viable path, though it would mean that concurrent editing would be harder to implement in the future the more we rely on the client being unique.

YiranShu:
When I try to generate Java API Doc, "A page you requested was not found" is displayed.

Steps to reproduce:

1. Open UmpleOnline.
2. Click Examples->class diagram-> 2DShapes.
3. Click Generate->Java API Doc.
4. "A page you requested was not found" is displayed.

pwang347:
@YiranShu Whew, looks like it's also a bug using the alternate URL (I can't repro it locally, and it seems like it's hitting the wrong endpoint):
image

pwang347:
Note that I've reverted the download to use the native PHP zip again, so it shouldn't work for now. I've tested that it works locally, but feel free to revert if the other method is preferred.

TimLethbridge:
I can confirm that the problem @YiranShu noted with Javadoc is present in all versions of UmpleOnline running in URLs that have a path with /something/ after the domain name, so it isn't related to this PR.

I am still going to do more testing, and might not merge this before I get back to Ottawa.

YiranShu:
I did more testings. I think this PR is OK.

TimLethbridge (approved):

TimLethbridge:
I am going to merge this now.

PR #1403, opened 2018-10-30
Minor fix for issue 535. Step 1: Format the output of each object

YiranShu:
It is a minor fix for issue 535. Make the output of each object created by the CRUD user interface generator look nice (not as json, but attribute: value, one per line).


TimLethbridge:
I tried to use this latest version. I used the Accidents example, created an AccidentType in the CRUD with arbitrary strings, and then seemingly in infinite loop occurred. A message with the following information appeared: 

Strong offset case occurred in show_element.php on line 179

And then a series of the following appeared. I had to close the tab to stop it.

Uninitialized strong offset: 41 in show_element.php on line 183

YiranShu:
> I tried to use this latest version. I used the Accidents example, created an AccidentType in the CRUD with arbitrary strings, and then seemingly in infinite loop occurred. A message with the following information appeared:
> 
> Strong offset case occurred in show_element.php on line 179
> 
> And then a series of the following appeared. I had to close the tab to stop it.
> 
> Uninitialized strong offset: 41 in show_element.php on line 183

Thank you! I may know what the problem is.

YiranShu:
@TimLethbridge 
I have updated my commit with Paul's code.

PR #1418, opened 2018-11-23
Progress on #751: Initial attempt at Gradle builds

pwang347:
## Context
The goal of this change is to add the ability to build the project using Gradle as described in the issue.

## Functionality by completion of this PR
- First build is working [o] e.g. `gradle firstbuild`
- Quick build is working [o] `gradle quickbuild`
- Full build [o] `gradle fullbuild` - tests not working
- Default build action should be managed by travisci afaik, so make sure that it runs smoothly in the check [o]

## Elaboration
- Currently, all gradle components are managed in dist/gradle to avoid conflicts with the default build directory
- Generated files are listed under the subprojects that they were compiled from in dist/gradle/src-gen
- Classfiles are compiled to dist/gradle/bin
- The final output jar file is found at dist/gradle/libs/umple-VERSION.jar

## For discussion
- There seem to be many issues with using the plugin in its current state (these issues are reflected when using the JAR as well) due to the default values that are supplied to the generated language and path, so it may be a better idea to port the build logic over later (and design with that in mind)
- A cleaner approach may be to access the internal Umple API directly to execute the compilation akin to the plugin, though doing things by commands for now seems more agile


TimLethbridge:
@pwang347 Thanks for the great work. When you have stabilized this, please let me know. I am finding myself snowed under, and getting further and further behind, and so need to budget my time carefully. My assumption is from your comments that you will have almost everything in place so people can build with either gradle or with the old process. But I imagine the devil is in the details ... for example I wonder if the umple-version numbering will be working, and whether the instructions for doing a release in the wiki will need to change. I hope I can count on you being able to assist if I run into problems over the coming weeks or months!! I will try to find some time to test once this is stabilized.

pwang347:
Will do! Yes, the current plan is for most of the core workflows (ie. at least umple-version.jar and jUnit tests) are working correctly and provide an alternate build process. Definitely feel free to ping me for any issues afterwards, and I will thoroughly document the current state of the change most likely after my last exam this Dec 13th.

As it is currently, for versioning I maintain the original scheme of naming the jar file `umple-<umple version>.<number of commits>.<hash of head commit>`, as well as replacing the `@UMPLE_VERSION@` string with it in the java files bundled in the jar file using a global replace operation. I'm currently attempting to fix all the jUnit tests as many are broken from running from a different directory (broken relative paths).

Off the top of my head, the missing portions are testbed  tests, eclipse plugin builds and the sandbox project. But I'm sure there may be more.

Additionally, there are some things that may need changing before the build workflow becomes a viable option. For example, I've been copying source files into a single directory under dist, which is not great performance, but helps identify what exactly is being compiled.

It also seems that Travis CI is running `gradle assemble` by default the moment it sees a build.gradle file, which builds the jar but doesn't run any unit tests (which may be good for our purposes, though in the ideal case we can run a separate job with its own test suite so that we don't add _any_ overhead to the current builds). It's also failing currently because of a missing hook to copy rtcpp libraries, so I'll fix that too.

TimLethbridge:
@pwang347 Eclipse plugin build can be dropped, as we are using a different process (plugin is done from the umple.har. Sandbox was only to test the build process itself and get a fast fail if not working, so lower priority. Thanks for all your work

pwang347:
Sounds great. Thanks Tim!

TimLethbridge:
Great! I see a wiki page https://github.com/umple/umple/wiki/Gradle-Workflow-and-Migration-Document too. Thanks. I will test in the next day or so.

TimLethbridge:
This comment is to ensure there is a link to issue #751

PR #1440, opened 2019-02-26
Graphviz feature diagram in umple online

gublan24:
A new **DIAGRAM TYPE** with the name "GraphViz Feature" is added in this PR to show the feature diagram. Its under **OPTIONS** in umple online. It might be better to name it "Feature Diagram" instead of "GraphViz Feature" but the later maintains the consistency with the other diagrams' names.

The GraphViz Feature diagram will look like the below diagram:  

umpleonlinegvfeaturediagram_example

### DIFF
--- a/cruise.umple/src/Generator_CodeGvFeatureDiagram.ump
+++ b/cruise.umple/src/Generator_CodeGvFeatureDiagram.ump
@@ -28,6 +28,15 @@ digraph FeatureModel {
 
   emit graphStart()(graphStart(UmpleModel.VERSION_NUMBER));
   
+  // Template for what is output if there is no feature diagram 
+  nofeatureDiagram <<! 
+  node [shape=plaintext];
+  message [fixedsize=true label ="No feature found in the umple file." width=4];
+!>>
+  emit nofeatureDiagram()(nofeatureDiagram);
+
+
+
   public void generateFeatureNodeShape(FeatureLink featureLink , StringBuilder code)
   {
     FeatureNode featureNode = ((FeatureNode) featureLink.getTargetFeature(0));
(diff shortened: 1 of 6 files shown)


TimLethbridge:
@gublan24 I have tested this using the docker image that is created for each PR. I used the dev-tools script udock, with the following
```
udock pr-1440
```

There are several problems:

State diagrams fail entirely. Something in this PR breaks them (whether I use ctrl-s, or hit the 's' button at the top.

I notice that for a 'blank' feature diagram nothing appears. Even in an empty Umple model there should be something sensible generated.

I notice that the following generates does in fact generate a diagram as expected:

```
mixset M1 {
  class A {}
}
use M1;
require [M1];
class B {}
```

However, if I leave out the require clause, nothing appears, not even an error message.

When there is an error in the model (any error), and the diagram is in feature diagram mode, the error messages and warning messages should be displayed, just as they are in other modes. So, for example, the following should display the error 12 as normal.

```
mixset M1 {
  class A {isA C;}
}
use M1;

require [M1];

class B {}

class C {isA A;}
```

In addition, all the other diagrams can be generated 'below' the main model using the 'generate' menu. Normally for new diagram types we prioritize in generating this (e.g. this is done for ERDs.

gublan24:
Think you Tim for finding these issues. I've found the cause for the failing of the state diagrams. For the rest, I will try to solve them and then update the PR. 

I do agree with adding the feature diagram in **Generate** and not allowing empty feature diagram. I will add them to this PR. I've not implement them here because I was considering them in another PR.

gublan24:
@TimLethbridge Could you test the branch **testBranch** or **GvFeatureDiagramUmpleonline2**. I've tested it using docker on my computer and its OK.

I use the following docker command to run a specific tag in umple/umpleonline:

 docker run --rm -ti -p 8000:80 -v /Users/myuserid/umodel:/var/www/ump/umodel umple/umpleonline:**testBranch** >/dev/null

PR #1441, opened 2019-03-12
Update 1001attribute.txt

Emmacho:
Insertion of omitted multivaluedAttribute definition in the main user manual page regarding attributes


TimLethbridge:
@Emmacho Congratulations on your first merged PR in Umple. I tested this PR in docker to see that the manual looked correct. It should go live in the manual in less than an hour

PR #1458, opened 2019-05-24
Fixes #1456 Runtime Exceptions

### LINKED ISSUE #1456: Make runtime exceptions explainable
Currently there are several places when Umple at runtime raises an exception due to various situations occurring.

For example the following code is found in https://cruise.eecs.uottawa.ca/umple/AssociationDefinition.html
```
class A {}
class B {
  1 -- * A endTwo; 
}

```
An A can only have one associated B, so in the constructor of A there is code that may fail the construction due to the presence of the B. The generated code thus throws an exception as follows:

```
throw new RuntimeException("Unable to create endTwo due to b");
```

The trouble is, when code is executing, and this message appears in a stack trace, the end-user will not have a clue what is going on. Even the beginner Umple developer may not understand, since they are not generally supposed to be looking at the generated code.

What is needed is:
1. Create a new 'Runtime Exceptions' section of the user manual (after 'Errors and Warnings 1000+ by modifying order.group

2. For each runtime exception:
  a)  assign a user manual page explaining the problem, including a various cases where this can occur, and a good explanation of what is going on and how to resolve it.
  b) on each of the places in the code where exceptions are thrown. Add '. See URL' after the textual descripton, where the URL is the pointer to the user manual. This will require changing various tests that were not expecting to encounter the URL.

To find the places where runtime exceptions are thrown in Java go to umple/UmpleToJava/UmpleTLTemplates and do

```
grep RuntimeException * 
```

The result lists 8 places, including
```
constructor_AssociationAssignOne.ump:6 constructor_AssociationAssignOneToOne.ump:5
constructor_AssociationAssignUndirectionalOne.ump
```

fjtheknight:
Uncomplete fix for issue 1456 Make runtime exceptions explainable:

**Work done:**

* Added 'Runtime Exceptions' section of the user manual. 

* Assigned manual pages for each group of issues.

* Added URLs for the user manual pages explaining the issues on templates and test files for JAVA only.

=> All tests now recognize the URLs and are successful

**Remaining task:**

* Add description and examples to the manual pages.

* Fix the issue for PHP and Ruby


TimLethbridge:
Excellent work so far.

For the manual pages, just draft them and I can edit further before we merge the PR, so that the English is perfect.

In the description you should explain that most violations (e.g. of multiplicity) are reported at run time by methods that return false. Programmers should therefore check for this. However when constructing a new object there is no way to return false; the constructor has to fail, hence these runtime exceptions are thrown.

For the examples, you should give examples that cause code from each template to be drawn. Ideally you would include a main method in at least one of the examples on each page that generates objects and then calls a constructor in a try-catch block, and where the exception would actually be triggered.

Tim

TimLethbridge:
The latest group of commits (May 31 morning) looks great.

TimLethbridge:
@fjtheknight Your recent commits have broken some manual pages. I think it is because you moved the @@syntax I have not checked all pages, but Trait Template Parameters page now fails to appear (because it doesn't parse properly). Same thing with the Method Definition page. Check other pages.

fjtheknight:
Ok thank you I will check that.

TimLethbridge:
@firas I will merge this once it builds correctly. I fixed the wording of 9601ViolationofImmutability.txt a little

PR #1459, opened 2019-06-06
Fixes #1444 Ensure comparators in sorted associations are serializable

### LINKED ISSUE #1444: Ensure comparators in sorted associations are serializable
From end-user Gunter Mussbacher: have recently discovered the "sorted" keyword in Umple. it's quite handy. however, there is an issue. in my course, we are using the Serializable interface to persist the application to a file.

This works quite well, but the Comparator used by the "sorted" keyword does not implement the Serializable interface. therefore, we have to change the generated code and make the instance variable using the Comparator transient. it would be great if the Comparator could implement the Serializable interface by default.

fjtheknight:
Uncomplete fix for issue #1444 Ensure comparators in sorted associations are serializable:

**Work done:**

* Changed comparator construction to Comparator.comparing(' + attributeType + '::get' + sortedName + ')'
* Added required modifications to the test files

**Remaining:**

* Add the "serializable" stereotype

* Add "transient" keyword to generated comparators when a class is declared serializable

* Add "readObject(java.io.ObjectInputStream in)" method to serializable classes


TimLethbridge:
@fjtheknight Some of the code being newly generated seemed to only work in Java 1.8 and higher, so I changed the build for tests of examples to work with 1.8 (I think at this point in evolution, that is reasonable)

TimLethbridge:
My suggestion to  enable adding 'transient' to the generated Java when we are generating a comparator attribute:

1. Add a boolean attribute internalTransient to UmpleAttribute, with the default being false;

2. set `priority.setInternalTransient(true);` in  UmpleInternalParser_CodeClass.ump line 2922 approx

3. In javaClassGenerator.ump line 285, add an || for getInternalTransient() to ensure the isTransient code is set.

TimLethbridge:
While doing the work to make the priority transient, can we also set internal to be true as well. (later on, perhaps as another issue we could consider fixing the ineration of attributes used to generate the three kinds of diagrams so that the internalTransient ones do not appear)

fjtheknight:
@TimLethbridge 

Pushed the changes made to the branch issue1444f:
* Added a boolean attribute “isTransientInternal” to the Attribute class in cruise.umple/src/Umple.ump
* Set its value to “true” for the comparator created in cruise.umple/src/UmpleInternalParser_CodeClass.ump
* Added “transient” to attributes for which getIsTransientInternal() returns true in UmpleToJava/UmpleTLTemplates/members_AllAttributes.ump
* Fixed issues with expected outcome for sorted associations tests

https://github.com/umple/umple/compare/issue1444...issue1444f

“transient” is added to the attribute declaration this way because there is already a string isTransient that is used for distributed classes, and just changing the value of isTransient to “transient” will indesirably affect other attributes (the string isTransient is accessible by all the methods used to generate the java code, see UmpleToJava\UmpleTLTemplates\JavaClassGenerator.ump line 284).

Distributed classes use the transient keyword and implement the serializable interface as well (see cruise.umple\src\Generator_CodeJava.ump line 2578/2671 for example)

fjtheknight:
Also adding the "internal" modifier to the comparators in cruise.umple/src/UmpleInternalParser_CodeClass.ump 

`
Attribute priority = new Attribute(yourEnd.getRoleName()+"Priority","Comparator<" + attributeType +">", "internal", "", false, yourClass);
`

 results in errors:

`
    [javac] G:\Git\Umple\umple\cruise.umple\src-gen-umpletl\cruise\umple\cpp\gen\GenClass.java:3228: error: cannot find symbol  
`
`
    [javac]                             Collections.sort(methods, getMethodsPriority());  `
`
    [javac]                                                       ^  `
`
    [javac]   symbol:   method getMethodsPriority()  `
`
    [javac]   location: class GenClass`

TimLethbridge:
@fjtheknight   OK. So internal is getting rid of getMethodsPriorty(). Two possible ways forward. Change the generation to simply access methodsPriority directly (I think this is best), but if this causes other problems then take out the internal.

fjtheknight:
@TimLethbridge 

I tried different approches to add the readObject method to the generated sorted classes, no success for now. I uploaded a branch (issue1444rOMethod) with the code I used and the errors I got.

1- I created a method object "readObject" and added it to the class (tried to use the same approach used to add a comparator as an attribute)
2- Add the full code for the method as extra code. This lead to a conflict with the readObject method produced for the distributable classes.
3- Because of that, I thought maybe we can just use the distributable keyword for this problem, but there will be a lot of extra code generated and some code needs to be changed (distributable solves the issue with transient, readObject method and serializable implementation)

My questions:
- I couldn't figure out how to properly add a custom method to a class and make sure it's code is generated with the rest of the class code
- To solve the issue with the readObject method of the distributable classes, should we make a template for each case? (changes to all the distributable class generation files are required)

Link with generated errors:
https://github.com/umple/umple/compare/issue1444...issue1444rOMethod

TimLethbridge:
@fjtheknight  It seems like you are running into problems that will require a lot of my time, and I won't have time for the coming week or so other than a few minutes here and there (I have to focus 99% on paper deadlines). We mustn't disrupt distributed. If you can figure out how to use the same generation that distributed uses, that would be good (but without the other unneeded code). Other than that, you might have to wait until I can give this considerable time, which might not be until I get back from my first vacation period in mid July. I will task you with other issues in the meantime.

fjtheknight:
@TimLethbridge I followed what was done for distributed and was able to add "import java.io.Seriablizable" to the generated class files (pushed the changes to issue1444, tests not yet fixed)

I had some issues with adding "implements java.io.Serializable" to the classes ([link](https://github.com/umple/umple/compare/issue1444...issue1444serializable1)):

- I tried to add "implements java.io.Serializable" to the generated java code (in JavaClassGenerator.ump, the output didn't change)

- I tried to add it as an interface named "java.io.Serializable" to the sorted association classes (in UmpleInternalParser_CodeClass.ump, commented). This lead to correct and wrong code generation at the same time (example: <class Mentor[]> but was:<class Mentor[ < java.io.Serializable]>)

fjtheknight:
@TimLethbridge 

The changes at branch issue1444 allow generation of transient comparators and classes that implement the serializable interface. This will allow serialization and deserialization of these objects, however modifying them after deserialization needs a readObject method for each class.

I tried to add the code for readObject methods to the classes as extra code, It didn't solve the issue totally, I need to check some conditions.

https://github.com/umple/umple/compare/issue1444...issue1444addingRO

fjtheknight:
@TimLethbridge 
The issue should be fixed now. I added a template for the readObject method and the code is added to the generated Java class code when needed.
However I did not add an automated test for serialization yet.

TimLethbridge:
@fjtheknight Fantastic! I will look at it when I have some time. I will be working on the weekend but by focus is on paper deadlines Monday and Tuesday that will require 20 hours of work before then! Meanwhile you can work on the tests

fjtheknight:
@TimLethbridge 
I fixed a small bug generating a  readObject method for serializable classes with no transient attributes.

Also I made a test file for serialization using a modified version of the sorted association example.
https://github.com/umple/umple/compare/issue1444...issue1444tests
The test works fine, but upon building umple, two java files are generated (Academy.java and Registration.java) and I couldn't figure out what's causing it.

TimLethbridge:
@fjtheknight Regarding the two files that are being built. Academy.java and Registration.java there is supposed to be an instruction to delete the files. Look through the other tests to figure out how this is done. If you can't do it, then I will. But an tight for time as you know. Please confirm that other than this, you believe all is now ready for final review?

fjtheknight:
@TimLethbridge 
I am thinking of adding a new test case, and fixing some formatting, but the code should be ready for final review within few hours (after I check the new test case). I will send you a zip file with a serialization/deserialization java example to make sure the output is what's expected.

TimLethbridge:
@fjtheknight Don't send a zip file. Put it in github and point me to the file there.

PR #1468, opened 2019-06-27
added condition to fix the case when the first code injection languag…

fjtheknight:
After some more testing, I found out that the problem appears whenever the first language specific code snippet is different from the language in which you want to generate your code

For example:

```
class TestMultiLang {
  name;
  public String customGetName () {
  } 
  before customGetName Ruby {
  //Ruby
  }    
  before customGetName Java {
  //Java
  }
  before customGetName Php {
  //Php
  }
}
```

will result in "null" generated for the first code injection if you generate code in Java or Php.

What I understood is that the problem comes from getCode() method of the CodeBlock class (cruise.umple\src\Umple_Code.ump lines 1163 and 1167), returning "null" if the first code specific language (as in customGetName Ruby{}) is different from the generated code language (generate Java code).

What I did is that I added a condition to check wether what getCode() returns is null or not, if it is, it doesn't add it to the generated code (cruise.umple\src\GeneratorHelper_CodeClass.ump line 87). This fixes the problem as the same previous example will now generate:
```
  // line 7 "model.ump"
   public String customGetName(){
    // line 9 "model.ump"
    
    // END OF UMPLE BEFORE INJECTION
    // line 15 "model.ump"
    //Java
    // END OF UMPLE BEFORE INJECTION
    // line 18 "model.ump"
    
    // END OF UMPLE BEFORE INJECTION
    
  }
```
which works, but I think the "header" and "footer" for Php and Ruby injected code (// line 18 "model.ump" and // END OF UMPLE BEFORE INJECTION for example) should not appear in the Java generated code (and vice versa). Should I investigate it more?

### DIFF
--- a/cruise.umple/src/GeneratorHelper_CodeClass.ump
+++ b/cruise.umple/src/GeneratorHelper_CodeClass.ump
@@ -61,21 +61,22 @@ class GeneratorHelper
     {
       for (CodeInjection inject : allCodeInjections)
       {
+        String comment = "//";//(inject.getSnippet().getCode("Ruby") != null)?"#":"//";
         String positionString = "";
         Position p = inject.getPosition();
         Position codeP = inject.getCodePosition();
         String injectFooter = "";
         String injectType = inject.getType();
         if (codeP != null) {
-            positionString =  "// line " + codeP.getLineNumber() + " \"" + inject.getUmpleClassifier().getRelativePath( p.getFilename(), inject.getUmpleClassifier().getSourceModel().getDefaultGenerate() ) + "\"\n";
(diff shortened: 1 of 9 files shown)


TimLethbridge:
@fjtheknight This looks like a good change. Proper policy now is to add a test so that we don't have regressions: In other words a test of the kind that would have caused the problem, that shows the correct output.

fjtheknight:
I was working on the test files for this issue and realized that the comments used for the header and footer of the injection (//something) are not compatible with Ruby (should be #something). This is a problem with any code injected into a Ruby method

PR #1470, opened 2019-06-29
Fixing issue#1467: Inline mixset error 

gublan24:
This PR fixes the case of an inline mixsetes that consists of one element, which is placed inside Umple classes, traits, or any Umple entities. 
Also, the parser will raise an error for unsupported inline mixsets. An error page has been added to explain the error and the way to solve the error.

### DIFF
--- a/build/reference/4102InlineMixsets.txt
+++ b/build/reference/4102InlineMixsets.txt
@@ -1,0 +1,19 @@
+E1511 Inline Mixset Is Not Supported 
+Errors and Warnings 1000+
+noreferences
+
+@@description
+
+<h2> Inline mixset is not supported error reported when the Umple parser can't interpret an inline mixset.</h2>
+
+<p> Umple allows inline mixsets for certain entities in Umple. Inline mixsets accepts classes, traits, and interfaces. If you encounter an error regarding the support of inline mixsets, you can easily add a pair of curly brackets around the code that a mixset introduces. The code bellow shows how to use inline mixsets.  <br/>
+</p>
+
+
+@@example
+@@source manualexamples/E1511InlineMixsetIsNotSupported.ump
+@@endexample
+
+@@syntax
+[[mixsetInlineDefinition]] [[mixsetInnerContent]] [[extraCode]]
+
--- a/build/reference/4102InlineMixsets.txt
+++ b/build/reference/4102InlineMixsets.txt
@@ -6,7 +6,7 @@ noreferences
 
 <h2> Inline mixset is not supported error reported when the Umple parser can't interpret an inline mixset.</h2>
 
-<p> Umple allows inline mixsets for certain entities in Umple. Inline mixsets accepts classes, traits, and interfaces. If you encounter an error regarding the support of inline mixsets, you can easily add a pair of curly brackets around the code that a mixset introduces. The code bellow shows how to use inline mixsets.  <br/>
+<p> Umple allows inline mixsets for certain entities in Umple. Inline mixsets accepts classes, traits, and interfaces. However, empty code classes, interfaces, and traits are allowed for inline mixsets. If you encounter an error regarding the support of inline mixsets, you can easily add a pair of curly brackets around the code that a mixset introduces. The code bellow shows how to use inline mixsets.  <br/>
 </p>
 
 
--- a/build/reference/4102InlineMixsets.txt
+++ b/build/reference/4102InlineMixsets.txt
@@ -4,16 +4,16 @@ noreferences
 
 @@description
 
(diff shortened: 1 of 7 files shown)


TimLethbridge:
I notice that the following code raises the error, but still the compiler then crashes!

```
class X {
  mixset m1 a;
}
```

TimLethbridge:
I changed the manual page slightly to improve it, and to make it render (the syntax information was in the wrong location, but I just removed it since we generally haven't been putting that in the error message pages)

gublan24:
> I notice that the following code raises the error, but still the compiler then crashes!
> 
> ```
> class X {
>   mixset m1 a;
> }
> ```

gublan24:
Great. I just put things with the intention to improve them before submitting the PR. 

Let me know if you have some test cases, I can add them to in the tests that we have in this PR.

TimLethbridge:
Just for the record, the third example in the associations in traits manal page (traits_example_multifeature) gives the following errors when you add use f1; use f4;

```
Error on line 65 : Inline syntax for the mixset : 'f5' is not supported. Please enclose the body of 'f5' with curly brackets.. More information (1511)
Error on line -1 : Compiler Error (Analysis). 
Exception java.lang.NullPointerException in at 
cruise.umple.compiler.UmpleInternalParser.analyzeMixset(UmpleInternalParser.java:15478) at 
cruise.umple.compiler.UmpleInternalParser.analyzeMixsetBodyToken(UmpleInternalParser.java:15270) at 
cruise.umple.compiler.UmpleInternalParser.analyzeClassToken(UmpleInternalParser.java:2022) at 
cruise.umple.compiler.UmpleInternalParser.analyzeToken(UmpleInternalParser.java:1311) at 
cruise.umple.compiler.UmpleInternalParser.analyzeAllTokens(UmpleInternalParser.java:1267) at 
cruise.umple.compiler.UmpleInternalParser.analyzeClass(UmpleInternalParser.java:2958) at 
cruise.umple.compiler.UmpleInternalParser.analyzeClassToken(UmpleInternalParser.java:1854) at 
cruise.umple.compiler.UmpleInternalParser.analyzeToken(UmpleInternalParser.java:1289) at 
cruise.umple.compiler.UmpleInternalParser.analyzeAllTokens(UmpleInternalParser.java:1240) at 
cruise.umple.compiler.UmpleInternalParser.parseMixsetWaitingFragment(UmpleInternalParser.java:15424) at 
cruise.umple.compiler.UmpleInternalParser.analyzeMixset(UmpleInternalParser.java:15499) at 
cruise.umple.compiler.UmpleInternalParser.analyzeMixsetToken(UmpleInternalParser.java:15438) at 
cruise.umple.compiler.UmpleInternalParser.analyzeToken(UmpleInternalParser.java:1298) at 
cruise.umple.compiler.UmpleInternalParser.analyzeAllTokens(UmpleInternalParser.java:1240) at 
cruise.umple.compiler.UmpleInternalParser.analyze(UmpleInternalParser.java:1129) at 
cruise.umple.compiler.UmpleModel.run(UmpleModel.java:1594) at 
cruise.umple.PlaygroundMain.processCommand(PlaygroundMain.java:763) at 
cruise.umple.PlaygroundMain.handleMessageFromClient(PlaygroundMain.java:459) at 
cruise.umple.AbstractServer.receiveMessageFromClient(AbstractServer.java:556) at 
cruise.umple.ConnectionToClient.run(ConnectionToClient.java:335) . More information (9100)
```

gublan24:
Right. "traits_example_multifeature" fails because of: 
 `mixset f5 after custom m2() {System.out.println("m2 occurred");}`. 
The main difference to the following code: 
` mixset f6 { BAD CODE %^%$^$%^ a }`
is that f5 code violates the syntax of inline mixsets since the aspect (before/after) is not supported, while f6 does not violate the syntax of regular mixsets. But it has a bad code that will be detected only when using f6.

An option of delaying the errors of unsupported inline mixsets until the time of their use. The parsing should continue and errors will be detected and saved in the mixset meta-model. This option has a bad design since it goes against early detection of errors in code. 

Therefore, I've update f5 in "traits_example_multifeature"  with curly brackets.

TimLethbridge:
Fixing the user manual example is good.

But the compiler should not crash on bad input EVER. It still crashes if given the original bad input.
That null pointer exception in at UmpleInternalParser.analyzeMixset should still be handled in some sensible way.

gublan24:
Right. I am using some JUnit tests to test my code and Its always shows the fail trace when there is an error. I see your point when using umpleonline or console. Sorry to this misunderstanding.

PR #1472, opened 2019-07-05
Issue1444testbed

fjtheknight:
Added testbed tests for sorted associations serialization (issue #1444)

The test is aimed at making sure that:
- the Serializable interface is correctly imported and implemented by sorted associations classes.
- the serialization of objects is done correctly
- the deserialization results in a "modifiable" object of the class, which means you can not only read the object, but also modify it (through add methods for example, if applicable)
- the objects obtained after deserialization are still sorted correctly


TimLethbridge:
Fantastic

PR #1476, opened 2019-07-12
fixed extraneous URL_SPLIT message in GraphViz views

fjtheknight:
An extraneous error message was displayed in GraphViz views in case the result from executing a command was an empty string.

I suspect the extraneous error message was used before the implementation of the current error display system and wasn't updated.

So my solution was that, in case a command execution returns an empty string (which means an error occured), check whether there is no associated error message (which means that the error is not known and there is no error file for it), if so, display the extraneous error message that asks the user to send the code to the Umple team, else, display the predefined error message related to the error.

The difference between Windows and other OS seems to be what executeCommand($command) returns. If there is an error, it returns an empty string, but not in Windows, and thus another message is displayed.

### DIFF
--- a/umpleonline/scripts/compiler.php
+++ b/umpleonline/scripts/compiler.php
@@ -320,24 +320,30 @@ else if (isset($_REQUEST["umpleCode"]))
   $sourceCode = htmlspecialchars($sourceCode);
   
   $errhtml = getErrorHtml($errorFilename);
-  
+
   if ($sourceCode == "")
   {
     if($input == "//$?[End_of_model]$?") {
-      $html = "
+      $msg = "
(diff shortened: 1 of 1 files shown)


TimLethbridge:
This PR does not work properly as it introduces a different problem. With this PR, if one is in 'E' (editable class diagram) mode and has no text, and then generates java, one gets a different wrong message.

I have looked into it and it turns out that you should scrap this PR.  The solution instead is to replace line 1058 of umple_page.js with

```
    output = code.replace(/<p>[\s\S]*/, "");
```

The regular expression was incorrect. I have solved this with #1477 so this PR will be deleted

TimLethbridge:
Closing as superceded by #1477

PR #1481, opened 2019-07-17
Fixes #1376 new error on multivalued vs. derived

### LINKED ISSUE #1376: Initialized multivalued variables fail to generate variables and initialization
emails and incoming number are not initialized as variables in the following, however the get and other methods appear, and the attributes appear in diagrams

```
class Office {
   Integer number;
   String[] emails = {"abc@umple.org", "def@umple.org"}
   Integer [] incomingNumber = {765432, 987654}
} 
```

Regular multivalued attributes (without initialization) seem OK.

TimLethbridge:
This PR is actually from changes made by @Emmacho 

Add new error 83 when creating a derived attribute that is also flagged as being a list.

### DIFF
--- a/build/reference/9083InvalidMultivaluedAttributeInitialization.txt
+++ b/build/reference/9083InvalidMultivaluedAttributeInitialization.txt
@@ -1,0 +1,20 @@
+E083 Invalid Multivalued Attribute Initialization
+Errors and Warnings 51-99
+noreferences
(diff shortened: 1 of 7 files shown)


TimLethbridge:
The core of PR has passed. Waiting for @Emmacho to add a user manual page before merging to fix #1376

Issue #1315, opened 2018-06-23
Need for namespace reset override


TimLethbridge:
Currently if you say 

```
namespace n;
class X {}

namespace m;
class Y {}

class X{}
```

you get in UmpleOnline
```
Warning on line 7 : Redefining the namespace of class X to m. More information (30)
Warning on line 9 : Redefining the namespace of class X to m. More information (30)
```

There are two problems here:

1. You might want in the second redefinition to not redefine the namespace to m. This is a serious problem when using a new file or mixset. 

2. In UmpleOnline the second message appears due to the hidden diagram information. The warning and redefinition will *always* appear.

Solution alternatives

a) Reverse the situation. Only take the first namespace, and leave it in place. Also add a 'namespace -;' syntax that would indicate that following this point no namespace is active. That could be injected to step putting subsequent classes in a namespace. This could be injected at the start of the diagram code.

b) Add an extra optional second argument to the namespace directive 'final' (or something similar) to prevent overrides and warnings about them.

c) Leave the first namespace as the final namespace. Don't override on any redefinition by default, and hence drop the warning. Also add a 'override' as a possible second argument to force an override should this be desired. The idea of having namespace - would still be a good idea but would not be needed to be injected in UmpleOnline for diagrams.

Suggest option c

vahdat-ab:
The issue you opened is one of the issues I have faced a while ago and tried to fix this with some constraints and warning. There are many scenarios you may imagine to mix namespaces. It becomes more complicated when you use `use` statements that have several namespaces defined inside the used files. Even the order of using statement will cause some concerns (with respect to other use statements or other names spaces)
The person who wants to fix this issue must also consider these cases. 
Fixing this issue with new definitions may cause some problems in building UmpleCompiler because there are some dependencies about how use statement works.

TimLethbridge:
OK. So here is the proposed solution, reflecting Vahdat's issue and respecting the need to enable existing systems to still compile.

1. Add 'final' as an optional second argument to the namespace directive. This prevents classes set with this namespace directive from being reset by a later mixin or mixset triggered by a use statement.

2. Add  '-' as an optional argument to the namespace directive. This cancels any outstanding namespace. Subsequently defined classes default to no namespace (unless a redefinition is encountered in a subsequent mixin/mixset for them. Note that '- final' makes no sense.

3. Add a warning if an attempt to redefine a final namespace is encountered (saying it is ignored), The existing warning would still be in place for non-final namespace redefinition.

[...]

PR #1482, opened 2019-07-22
added --redefine to namespace to redefine class namespace + fixed err…

fjtheknight:
issue #1315 

1)    Warning 30 needs to be adusted to say Ignored attempt to redefining the namespace of class {0} to {1} . Use --redefine to force this'; **done**

2)  A --redefine option following the namespace needs to be added to allow redefinitions. **done**

3)   A - option for the name of the namespace needs to be added to cancel the current namespace specification **done**

4)    Examples need to be fixed so those with hidden layout no longer specify error 30. **not yet**

The following example shows the differences:

```
class Z{}

namespace x;
class B{}

class C{m;}

namespace u --redefine;

class C{}
class J {};

namespace y ;
namespace -;

class C{h;}
namespace l;
class C{x ;}
class K{}

namespace -;
```

Classes Z and K should be in package l
Class B should be in package x
Classes C and J should be in package u
Package y should contain no classes
A warning at line 19 should appear to suggest redefinition of namespace for class C:
` Warning on line 19 : Ignored attempt to redefining the namespace of class C to l . Use --redefine to force this. More information (30)`


fjtheknight:
Documentation should also be modified.

TimLethbridge:
@fjtheknight  I would expect the following code to put X outside of any namespace; Y in namespace p, and Z outside of any namespace. But all are in namespace p. The `namespace -;` doesn't seem to be working as expected. 

```
class X {}

namespace p;
class Y {}

namespace -;
class Z {};
```

Also, in the example you gave, class Z shouldn't be in any namespace, because line 21 `namespace -`' should prevent the namespace from being applied to any classes in the diagram code.

fjtheknight:
@TimLethbridge As of now Umple puts classes that aren't in any namespace in the last detected namespace by default. So this should be changed.

cruise.umple/src/UmpleInternalParser_CodeClass.ump line 1195

EDIT: I need to verify this again

TimLethbridge:
@fjtheknight So we discussed and I think you understand. Yes, Umple should continue to put classes that aren't in any namespace in the last detected namespace. But in the new setup `namespace -` would cancel the last detected namespace.

fjtheknight:
@TimLethbridge namespace -; now cancels the last namespace.

* There is a small issue however (might not be an issue):
```
namespace y ;
class H{}
namespace -;
namespace l;
class F {};
namespace -;
```

Here the first namespace -; is not necessary, but because nothing was added between namespace - and namespace l, we will get a warning:
`Declared namespace l was not used before declaring l.`

The right way to write this code is 

```
namespace y ;
class H{}
namespace l;
class F {};
namespace -;
```
Should this be considered an issue?

* Adding namespace -; at the end of the code will prevent the warning message due to hidden diagram code. Should we expect the user to add namespace -; whenever a namespace is used in the project or should we add one by default at the beginning of the diagram code?

TimLethbridge:
@fjtheknight You are right. The error 31's that appear are not valid. In the case you cited, there should be no such warning, as there should be no active namespace,

The following code produces two such problems.

```
namespace y ;
class H{}
namespace -;
namespace g;
namespace l;
class F {};
namespace -;
```

And the following code, even in the original, before your changes, raises this erroneous warning. It should say Declared namespace g was not used before deglaring I.

```
namespace y ;
class H{}
namespace g;
namespace l;
class F {};
```

If this is a trivial fix, then fix it in this PR. Otherwise raise a new separate issue, assign yourself, reference this PR in ot and create a new PR for the new issue.

Regarding whether `namespace -;` should appear at the start of the diagram code. The answer, in my opinion is that this is probably the best thing to do.

I notice that the end_of_model string is defined at umple_page.js:18

I suggest that the correct place to change this is umple_page:722 which currently is 
```
  var umpleCode = modelCleaned + Page.modelDelimiter + positioning;
```

this might perhaps be changed to something similar to
```
  var adjustedPositioning;
  if(!positioning.equals("") {
   // prepend namespace cancellation to prevent namespace redefinition errors
    adjustedPositioning = "namespace -;\n"+positioning;
  }
  else
  {
    adjustedPositioning = positioning;
  }
  var umpleCode = modelCleaned + Page.modelDelimiter + adjustedPositioning;
}
```

fjtheknight:
@TimLethbridge 

The warning message is fixed now, and namespace -; is added at the start of the diagram code.

The only remaining issue is when to make warning 31 appear. The problem for now is that the code that adds the warning to the warning only works with previously declared classes (around line 5490 in UmpleInternalParser_CodeClass.ump) which means that:
- whenever one of these classes appear in the code after the namespaces, a warning will appear, and thus repeating the warning for the same namespace multiple times
```
namespace w;
namespace x;
namespace y ;
namespace z ;
class H{}
class J{}
```
gives the following warning messages:
```
Warning on line 2 : Declared namespace z was not used before declaring x. More information (31)
Warning on line 3 : Declared namespace z was not used before declaring y. More information (31)
Warning on line 4 : Declared namespace z was not used before declaring z. More information (31)
Warning on line 2 : Declared namespace z was not used before declaring x. More information (31)
Warning on line 3 : Declared namespace z was not used before declaring y. More information (31)
Warning on line 4 : Declared namespace z was not used before declaring z. More information (31)
```
so 3 warnings for each new class

- if a class has already been added, the warning won't appear:
```
namespace w;
namespace x;
class H{}
namespace y ;
namespace z ;
class H{}

```
gives the following warning messages:
```
Warning on line 2 : Declared namespace x was not used before declaring x. More information (31)
Warning on line 6 : Redefining the namespace of class H to z. More information (30)
Warning on line 9 : Redefining the namespace of class H to z. More information (30)
```
so the warning for namespace y is ignored.

This means that the code for warning 31 should be elsewhere, and the way tokens are analyzed makes this not trivial I think.

TimLethbridge:
@fjtheknight Yes, I see that the warning 31 code should be elsewhere. I think the last comment should be put in the main issue; or alternatively in a separate issue.

fjtheknight:
@TimLethbridge 

(current fix is pushed on branch issue1315w31b and will be merged to this branch if approved)

The issue with warning 31 is fixed now as well as some other issues:

**1. Warning 31:**

There was a problem with the solution you suggested to fix this issue:
When a namespace is redefined, the previous namespace could no longer be used, and thus a warning should be issued. So simply checking namespaces on parsing won't fix the problem. So what I did is that:

- I added a new class (NameSpace) to save namespaces' values, positions and how many times they have been used, and added a NamSpace attribute to the UmpleElement class
- Whenever a namespace is parsed, it is added to a list of NameSpaces
- If a namspace is used, increase its usage counter
- If a namespace is redefined, decrease its usage counter
- After parsing is done, check the parsed namespaces list and issue a warning whenever a namespace has been used 0 times



- I also replaced the old method (getCurrentNameSpace) used for getting the current package name (which required looping through all the tokens whenever a class is parsed) with a simple NameSpace variable (lastParsedNameSpace) which highly improves performance
- The warning message was simplified to just say that the declared namespace isn't used

**2. Interfaces' packages:**
There was an other issue with Interfaces and namespaces: 
In the following example
```
namespace x;
interface I
{
  String getA();
}
namespace y;
interface I
{
  String getB();
}
```
interface I should be in package x, and contain two methods, however, currently the parser will ignore previously defined parts of the interface and will create an interface with only the last part parsed in the wrong package, so the code generated will be:
```
package y;
public interface I
{
 public String getB();
}
```
instead of 
```
package x;
public interface I
{
 public String getA();
 public String getB();
}
```
This is now fixed.

3. **Old test examples:**
Error and warning tests are fixed.


The following example demonstrates the changes:
**Example:**

```
interface I1 {}
class C1{}

namespace n1;
class C2{}

namespace n2;
interface I2{}
class C2{att1;}
class C3{};
  
namespace n3;
namespace n4;
class C4{}
namespace -;

class C5{};

namespace n5 --redefine;
interface I2{void m1();}
class C2{att3;}
```

In this example
- no namespaces were declared for I1, C1 and C5
- the namespace for I2 and C2 was redefined to n5, namespace for C3 wasn't changed (n2)
- n3 wasn't used, n1 is no longer used after redefining the namespace for C2

So we have 2 unused namespaces (n1 and n3 => 2 warnings 31), and an attempt to redefine the namespace for C2 to n2 (warning 30)

This is to show the changes made to solve the issue with warning 31:
https://github.com/umple/umple/compare/issue1315redefine...issue1315w31b

I have some questions:
- Should the NameSpace and NameSpace[] attributes be internal?
- Should the documentation be changed?

TimLethbridge:
@fjtheknight . Your changes in branch issue1315w31b look good, as far as I can tell. 

As a rule, I prefer the use of directed associations as opposed to list attributes, since the associations appear in the diagram. I am not going to insist on this change, however.

No the NameSpace and Namespace[] should not be internal.

I would like you to add at least one more test to test the changes to interfaces. Perhaps create a complicated test that has a bunch of interfaces and classes in the same namespace, multiple declarations of the same namespace that is used. Multiple declarations of a namespace that is not used, with one case being where it is declared and then cancelled by `namespace `.

Then merge into this branch and let me know. I would like to do manual tests of all the cases at that point before merging the PR into master.

fjtheknight:
@TimLethbridge 

There is an issue with associations and classes that have no namespace.

In the following example:
```
class Out{}
namespace p;
class In{*--* Out;}
```
class In is in package p, but class Out is in the default package. But because there is an association, there will be an attempt to import a package for class Out in class In file (which is just an empty string in this case) :
```
//%% NEW FILE In BEGINS HERE %%

package p;
import java.util.*;
import .*;

public class In
{
  ...
}

//%% NEW FILE Out BEGINS HERE %%

import java.util.*;
import p.*;

public class Out
{
  ...
}
```

The problem is that in Java classes in the default package cannot be imported by classes in packages.

The issue occurs only when we have an association with a class that is in the default package. We can either raise a warning saying that classes in default package can't be imported by classes in packages, or we can just add a "association" package by default to classes with no packages if this case is detected. There are other solutions that will add a lot of extra code.

**EDIT:**
This also happens when we have a class in a package that extends a class that is in the default package. So the generated code for class X2 in following example:
```
class X{};
namespace p;
class X2{isA X;}
```
will be:
```
package p;
import .*;
public class X2 extends X
{
  ...
}
```

TimLethbridge:
@fjtheknight  I assume the issue with packages and associations is a longstanding issue and not introduced by your changes? If so, then I think we should raise it as a new issue rather than tag it on to this one. Raising a warning I think is not a good idea, as programmers should not be restricted in that way. if there are solutions that would require a lot of code, please describe how you think that code might look when you raise the issue.

fjtheknight:
@TimLethbridge 
No this issue is introduced by my changes. Currently Umple just puts classes that have no namespace in the last parsed namespace, this way the issue is avoided. But now, classes with no namespace are put in the default package.

There are possibly two workarounds for this issue:

Let's say we have this example:
```
class A{
//code here
};
namespace p;
class B{isA A;}
```
1. The first one is to generate a class in package p, and make class A and B extend it. So the above code should be interpreted by Umple as follows:

 ```
class A{
isA A2;
};
namespace p;
class B{isA A2;}
class A2{
//code here
}
```
This way, class A and B will have access to class A2 because it is either in the same directory or in a sub-directory. But that's against the idea of having class A code in the default package.

2. The second one is to use Reflection. The problem is that I couldn't find a way to solve the issue with reflection for the example with superclasses, and that I couldn't find similar concepts in Ruby.

**Comment by TimLethbridge** [comment]
@fjtheknight 

Firstly, we should avoid solutions such as renaming classes and adding extra supervlasses as suggested above.

We should also avoid adding complex solutions such as using reflection. Reflection tends to result in slow code and code that is hard to understand. It is also more prone to security issues. Some previously acceptable reflection is now deprecated.

So back to the original: classes that have no namespaces should just be in the last parsed namespace. Why can't we do that? In fact it is core to the design. I should be able to say:
```
namespace X;
class Y {}
class Z {}
```

with the result that both Y and Z are in X. Does this no longer work in your changes? Can you give an example of code that will result in changes to which namespace they are in?

**Comment by fjtheknight** [comment]
> ```
> namespace X;
> class Y {}
> class Z {}
> ```
> 
> with the result that both Y and Z are in X. Does this no longer work in your changes? Can you give an example of code that will result in changes to which namespace they are in?

@TimLethbridge 

This still works. The problem only occurs when we have a class that has no package name (by that, I mean classes that are declared either before any namespace is declared, or after namespace -;) :
```
class A{} //will be in default package
namespace n;
class B{} //will be in package n
class C{} //will be in package n
namespace -;
class D{} //will be in default package
```
As for your example, both Y and Z are in X. To change this, we need to use --redefine; :
```
namespace X;
class Y {}
class Z {}
//both are now in X

namespace X2 --redefine;
class Y{}
//Y is now in X2
namespace -;

class Z{}
// Z is not affected by X2 because of namespace -;

namespace X3;
class Z{}
// Z is not affected by X3 because there is no --redefine; and is still in package X. A warning will be issued because of this
// Another warning is issued because X3 was not used
```
Also, currently, if a class is in a package, there is no way to put it back in the default package. Should we add that?

**Comment by fjtheknight** [comment]
@TimLethbridge 
I merged all the changes from the other branch to this one.

There was a problem with interfaces: if an UmpleClassifier implements an interface that is in an other namespace, "import interfaceNameSpace.*;" won't appear in the generated code.

For example, the following code
```
namespace f;
interface F{}
namespace j;
interface J{isA F;}
```
gives
```
//%% NEW FILE F BEGINS HERE %%

/*This code was generated using the UMPLE 1.29.1.4607.2d2b84eb8 modeling language!*/
package f;
// line 3 "../../../../ump/tmp821300/model.ump"
public interface F
{
  
}



//%% NEW FILE J BEGINS HERE %%

/*This code was generated using the UMPLE 1.29.1.4607.2d2b84eb8 modeling language!*/
package j;
// line 5 "../../../../ump/tmp821300/model.ump"
public interface J extends F
{
  
}
```
I was able to fix the issue with classes that implement an interface from another namespace, but there is still a problem with interfaces that extent interfaces from other namespaces. UmpleClass has GeneratedClass that has addMultiLookup method which can be used to add imports. The code generator uses that later to generate the import code. I couldn't find anything similar for interfaces.

The latest changes added a check for imports from the default namespace (class-class and class-interface implementation, extension and association). In this case, a warning is issued and the namespace is changed.

Tests and examples should be added after manual tests.

EDIT:
There is a small language mistake in the warning that should be fixed.

PR #1485, opened 2019-08-07
temporary fix for issue 1483

fjtheknight:
This is a temporary fix for issue #1483 

The infinite loop occurs whenever there is a method defined inside a standalone state machine.The current fix checks whether the parsed method is associated to a class or not, if not, it will show a warning message:

`Methods in standalone state machines might not be correctly generated. State machines can be used in traits or put directly in a class to achieve the same effect.`

and the code for the state machine might be ignored in order to prevent an infinite loop induced when the parser is looking for the class of the state method being analyzed (UmpleInternalParser_CodeStateMachine.ump line 1990)


TimLethbridge:
I am not sure the fix needs to be temporary. This is a rare problem.

Issue #1442, opened 2019-03-17
Problems with constructors and setting with unique construct


TimLethbridge:
1) in the example below, there is an issue with the second constructor
generated for A. the constructor just assigns aName to name instead of calling
setName(aName) as is done in the first constructor. because of that, there is
no check for duplicate names and the hashmap asByName is not properly updated.

```
class A {
   1 -- 1 B; 
   unique name;
 }
 
 class B {
   anotherName;
 }
```

Similarly, if a
state machine is defined for A, then the second constructor also does not set
the initial state.

The above was communicated by email from Gunter Mussbacher

2) there is another issue with the setName(aName) method. assume that an
instance has "myName" as the value for the name attribute. if setName(aName)
is called and the same name, i.e., "myName", is passed into it, setName(aName)
fails, because it detects that "myName" already exists in the hashmap, but
does not realize that it belongs to the same instance that is being changed.
this issue occurs for any unique attribute, i.e., it also occurs for a simple
model with one class and one unique attribute.

TimLethbridge:
For point 1, this is a feature interaction between
* the 1-1 association capability that constructs both objects to be linked woth the 1-1 association at the same time
* the unique capability that prevents duplicate objects from being created. See constructor_AttributeAssignUnique.ump which is invoked from constructor_DeclareDefault.ump:177

When both features are present, aspects of both constructors need to be present.

PR #1486, opened 2019-08-12
Issue1442

fjtheknight:
Issue #1442 :
- Added check (setAttribute) for the second constructor generated for classes in 1-1 associations with unique attributes to prevent duplicate attribute values.
- setAttribute now returns true if the argument's value is the same as the unique attribute's current value of the same object.
- fixed tests for unique attributes accordingly

I tried also to do control for primitive types in Java (see branch issue1442), added conditions to generate code according to whether the attribute is primitive or not, than found out that Map and Hashmap don't work with primitive types, which means wrappers need to be added conditionally so I stopped working on that until we discuss it's utility and just added control for Umple builtin data types


TimLethbridge:
Looks OK, except no indentation in Java case.

PR #1527, opened 2020-02-11
Disable auto-generated constructor and delete()

gublan24:
This PR allows disabling default constructor and the method 'delete()' within a class by `strictness disable constructor` and `strictness disable delete()`. The grammar ` strictnessDisableAuto : disable [**expression]` accepts general expression and can be extended (with minor modification in UmpleParser) and used be as a first class entity in Umple.  

However, the issue comes with auto-generated methods (setters, getters, adders, etc). They are produced in Umple generators as string output. Its extremely difficult to disable these auto-generated methods since the logic to produce these methods is spread over many places. Refactoring Umple generators into modular ways is very crucial to allow strictness capability. For example, JavaGenerator should translate UmpleModel into objects of javaGeneratedClass(s) that represent classes, method, access modifiers, etc. Then javaGeneratedClass(s) can be written into Java files at last stage.


TimLethbridge:
I notice that right now:

a) You are just allowing suppression of constuctor and delete. I was hoping for the ability to suppress any generated method pattern, not just delete (by filtering for patterns in the generated output stream perhaps). The code in this PR is super-specific.

b) You have enabled the strictness clause to be in a class (which is great), but what would happen if the 'strictness disable delete();' was found outside a class, or if 'strictness ignore 99;' was found inside a class. At the very least there should be no crash. In the ideal case, the outside-class case would make it so for all classes, and the inside-class case for the ignore class would make the ignore of a warning/error specific to that class (which would actually be better than the current approach).

c) It only works with generated Java. 

I can understand that totally general solutions to the above would take more work, but we need to strike a balance between making it perfect and making it a special-case-only change. One possibility would be to raise an issue (which should be done anyway) and then implement a partial solution, leaving the issue open for the rest of the solution to be done.

gublan24:
For (b) its possible and not difficult to make strictness general for all classes when it appears as top level entity. 

The two point (a) and (c) are related. Yes, It only works with generated Java because most of the logic of handling associations, state-machine, etc are language-specific in Umple. They produce different methods at generation time. The parser does little things before generation. Therefore, there is no way currently to ignore a method related to association before code generation.  The ideal case would be a lightweight generators while the association logic should be moved to the parser. I believe  getters and setters of attributes can be switched off after parsing. I try to include them to be filtered by strictness.

PR #1531, opened 2020-02-19
issue 1517-item 1,2,3,4 

HaowenShi:


TimLethbridge:
Your change to point 1 (tabs) is incomplete. The instructions said 'It is slightly darker when the tabs are active' and lighter when they are not (same as the A, M and other buttons.'

Take a look at the 'A' button. This toggles display of attributes. When you mouse over it, it goes very dark. When attributes are visible (the default) it is of medium darkness, and when attributes are hidden it is lighter. The medium darkness gives feedback to the user regarding whether attributes are visible. As you can see from line 234, it runs 'javascript:Page.clickToggleAttributes()' to achieve this. Take a look at the code in umple_page.js; you will see this function works with buttonToggleAttributes. Search elsewhere for buttonToggleAttributes to see how this works, including the initHighlighter function.

The code you are changing calls 'javascript:Page.toggleTabs()' and needs to be enhanced to do the same thing in umple_page.js.

I notice you have also made lots of changes to remove extra spaces. Although this is not 'bad' it made it hard to review these specific changes. Generally cosmetic changes of this magnitude should be done in a separate commit. Don't worry about that this time, but pay attention in future.

Please email me to let me know that you are noticing comments like this.

umpleonline/umple.php
```diff
@@ -241,7 +241,7 @@
     &nbsp;
 
     <span style="font-size: 30%; white-space:nowrap;">
-    <a class="button2" href="javascript:Page.toggleTabs()" title="Toggle tab visibility">Tabs</a>
+    <a class="button2 active" href="javascript:Page.toggleTabs()" title="Toggle tab visibility">Tabs</a>
```
TimLethbridge: This makes the button active. But it does not switch between active and inactive appearance as it should between clicks.

umpleonline/counter.php
```diff
@@ -35,8 +47,18 @@
...
 	fclose($log);
-	$count_new=$count/1000;
-	echo " | $count_new K commands run since July 2019" ;
+	if ($count>=1000 && $count<1000000) {
+		$count_new=$count/1000;
+		echo " | $count_new K commands run since July 2019" ;
+	}
+	elseif ($count>=1000000) {
+		$count_new=$count/1000000;
+		echo " | $count_new million commands run since July 2019" ;
```
TimLethbridge, line 56: This does not round to two decimal places as I had asked. So for example if the log file  was 123456789013 the output is 123456.789013, when it should be 123456.78

Same problem above for countlog.txt for the number of visits.

TimLethbridge:
I am going to close this without merging. I have included the functional bits in the release code for 1.30, but there were too many changed lines that just changed spaces, messing up the rest of this

PR #1556, opened 2020-06-04
Fixes Issue#1488

WeiqiMaster:
Now if a method block only adds constraints, testcases, or assertion (but not codelangs) i.e. no new executable code, then warning 49 would be suppressed.


TimLethbridge:
There are some print statements in the testcases that need to be removed (print statements in testcases clutter the build transcript)

PR #1558, opened 2020-06-06
[skip ci] Fixes issue#1554. Now providing a html link rather than using js window.location.href

WeiqiMaster:
Now the Download Files button will give a static link rather than using js window.location.href to directly download the zip file. 
Whenever the code in the left panel is synchronized (whether manual sync or auto sync), the download link will disappear.


TimLethbridge:
Good job on this so far.

Here are a few adjustments I need made.

- The 'Download Zip File From Here' appears under the 'Save to Dropbox', yet it is relevant to the 'Download Files' so it should be right below that. Actually I think it is time to get rid entirely of the 'Save to Dropbox'. Dropbox keeps changing their API, and the current implementation does not work, and hasn't been useful for some time. Please just get rid of it.

- The 'Download Zip File From Here' should disappear after 30 seconds. The way to do this is to create javascript function that runs after a timer. There are several examples of this that get rid if status messages after a timed delay.

- In any PR involving Javascript, you need to commit the minified file along with the others, so that when someone installs the system (or when it appears on Docker) it works.

- You should not use [ci skip]. Reason: We need the Docker image built. [ci skip] should only be used when the only changes are in comments or in user manual pages that are purely relatively small text changes (even manual pages with extensive comments and all manual changes with code changes need to have CI run since in fact all the examples are tested by the way).

WeiqiMaster:
I think removing the link whenever the code in the left panel gets synchronized is a better solution compared to making the link disappear after 30 seconds. So we can make sure users always get the latest zip even if they modify the code in 30 seconds.

PR #1559, opened 2020-06-10
Feature model examples

gublan24:
In this PR, "Feature Diagram " is added to the Umple tool webpage under "EXAMPLES" with two examples to show mixsets and feature modeling.


TimLethbridge:
You should put at the end of the file BerkelyDB_SPL.ump the following to stop the testing system from trying to see if code generated from it is valid. 


```
//$?[End_of_model]$?
// @@@skipjavacompile - Generates no language code
// @@@skipphpcompile - Contains no language code
// @@@skiprubycompile - Contains no language code
```

TimLethbridge:
I added the needed lines since it was trivial to do so and I want to keep progress moving

PR #1579, opened 2020-07-03
Fixes #1254 "Some test cases in testbed are never executed"

### LINKED ISSUE #1254: Some test cases in testbed are never executed

WeiqiMaster:
This PR adds some testbed specialization tests.


TimLethbridge (changes requested):
Lots of instantiation of the model is not what was needed here. What was needed is creating links of the associations (sufficient numbers of links to test the specialization rules)

PR #1608, opened 2020-07-22
Fixes #1602 "Uncompilable Java - variable and method already defined when duplicate after timeout in different state machines"

### LINKED ISSUE #1602: Uncompilable Java - variable and method already defined when duplicate after timeout in different state machines
If there are two or more state machines, each with two identically-named states, and where one of the states has a timeout (after) that transitions to the other state (and this after has the same origin and destination state in both/all state machines), then the code generator generates duplicate variables and methods

The following code will produce an error since the same variable timeouts1Tos2Handler and method of the same name is produced for both state machines. 

Solution would be to generate both of these with a name that is prefixed with the state machine name. This will require extensive testing, and modification of multiple tests, so be very careful.

There is a workaround for users, which is to simply rename states in situations where an 'after' statement goes from one state to another and where both state machines have the same states with the same transition. Having identically named states in different state machines is, however, not an uncommon situation.

An alternative possibility would be to produce a warning, but there is really no logical reason why people should not be able to do this in umple code, and in fact if using traits it might be particularly common.

```
class X {
  sm1 {
    s1 {
      after(1) -> s2;
    }
    s2 {
    }
  }
  
  sm2 {
    s1 {
      after(1) -> s2;
    }
    s2 {
    }
  }
}
```

Produces the following errors when Java is generated:

```
Oven.ump:16: error: variable timeouts1Tos2Handler is already defined in class X
  private TimedEventHandler timeouts1Tos2Handler;
                            ^
Oven.ump:141: error: method startTimeouts1Tos2Handler() is already defined in class X
  private void startTimeouts1Tos2Handler()
               ^
Oven.ump:146: error: method stopTimeouts1Tos2Handler() is already defined in class X
  private void stopTimeouts1Tos2Handler()
               ^
```

WeiqiMaster:
This PR fixes the issue by prefixing state machine name to the timeout variable and method.


TimLethbridge:
Would you please add a test for this case? (a template test showing the output produced).

PR #1625, opened 2020-08-11
Fixes #1599 "Add a button to re-indent code"

### LINKED ISSUE #1599: Add a button to re-indent code so it looks nice and makes it easier to understand
In UmpleOnline, add an command in the menu (perhaps even a button) to re-indent the code. So that it looks nice.

This would apply a rule of indentation of two spaces.

It would not shorten any lines of text if they are long. It would also not adjust any other aspects of style, such changing the location of starting curly-brackets. But ending curly brackets that are preceded by text on the same line would be moved to a line of their own, unless the starting curly bracket was earlier on the same line.

It would match curly brackets (e.g. for classes, methods, mixsets, and all other elements. Lines of code that are between curly brackets would be indented by 2 spaces.

The code would have to be careful not to be tricked by escaped curly brackets, or curly brackets within strings.

If there is additional special indentation within blocks (e.g. on conditions), this would be preserved

WeiqiMaster:
Adds a button 'Reindent Code' in the 'Draw' menu below 'Redo' to indent code in the code panel.

### DIFF
--- a/umpleonline/scripts/umple_action.js
+++ b/umpleonline/scripts/umple_action.js
@@ -2758,11 +2758,8 @@ Action.reindent = function(lines)
 
     var indexOfFirstQuote  = trimmedLine.indexOf("\"");
     var indexOfSecondQuote = trimmedLine.indexOf("\"", indexOfFirstQuote + 1);
-    console.log(i + ":::: ");
     if (indexOfFirstQuote != -1 && indexOfSecondQuote != -1)
     {
-      //var openCurlyBraceExits = trimmedLine.indexOf("{") != -1 && trimmedLine.indexOf("{") < indexOfFirstQuote && trimmedLine.indexOf("{") > indexOfSecondQuote;
-      //var closeCurlyBraceExits = trimmedLine.indexOf("}") != -1 && trimmedLine.indexOf("}") < indexOfFirstQuote && trimmedLine.indexOf("}") > indexOfSecondQuote;
       var indexOfOpenCurlyBrace = trimmedLine.indexOf("{");
       if (indexOfOpenCurlyBrace != -1 && indexOfOpenCurlyBrace > indexOfFirstQuote && indexOfOpenCurlyBrace < indexOfSecondQuote)
       {
@@ -2777,14 +2774,11 @@ Action.reindent = function(lines)
       var indexOfOpenCurlyBrace = trimmedLine.indexOf("{");
       var indexOfCloseCurlyBrace = trimmedLine.indexOf("}");
     }
-    
-    //continue;
 
     if (indexOfOpenCurlyBrace != -1 && indexOfOpenCurlyBrace != trimmedLine.length - 1)
     {
       lines.splice(i + 1, 0, trimmedLine.substr(indexOfOpenCurlyBrace + 1));
       lines[i] = lines[i].substr(0, lines[i].match(/^\s*/)[0].length + indexOfOpenCurlyBrace + 1);
-      console.log(lines);
       Action.reindent(lines);
       return;
     }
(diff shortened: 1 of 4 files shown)


TimLethbridge:
I just went through a large number of examples of Umple files to test the new functionality. Overall it is pretty good.

But we need to make a few changes. For the ones that are not quite so important and are also harder, it would be OK to do those as separate PRs, so we can get this one merged.

1. (Very Important and likely very easy) If the start and end curly brackets are on the same line as the start brackets, and the block is less than 40 characters, then that block should be left 'as is'. Examples of this are in the 'Access Control' example in UmpleOnline. 

```
  description { Menu, Record, Screen }
```

Should not be reindented.

Nor should

```
  key {id}
```

Another example of this is in example Airline where

```
class PersonRole{}
```

should be unchanged.


2. (Very Important and likely not too difficult) If code is commented out, any curly brackets in it should be ignored. Example:

```
class X {
  y = "cat"
    //  }
  a;
}
```
This should apply to  /* */ comments as well.


3. (Moderately important and probably moderately hard) In long statements, a programmer may want to wrap them to a second and subsequent lines. This should be preserved. See the example State Machines / Canal Lock. Line 13 wraps to line 14, and line 14 has extra indentation.

```
      boatRequestsToEnterAndGoDown
        -> OpeningUpperGate;
```

The rule should be:

Where:
  in(n) is the original indentation of line n  e.g. 7
  nin(n) is the new indentation of line n      e.g. 4
  in(n+1) is the indentation of line n+1       e.g. 9
  non(n+1) is the new indentation of line n+1 after adjustment

If line n does not have as its last nonblank character a semicolon or start or end curly bracket
and has indentation in(n), and after adjustment would have indentation nin(n):

and the following line (n+1) does not have any indentation adjustments itself resulting from curly brackets,

then the following line (n+1), which would normally also be indented nin(n), should instead be indented 

  nin(n) + max(0, in(n+1) - in(n))
  
Which would be 6 in the above example.

This calculation would need to continue for subsequent lines, referring back to the original line n, until the next line ending with a semicolon is encountered, or an indentation change resulting from a curly bracket is encountered.

Note that as soon as a semicolon or curly bracket at the end of a line is encountered, these 


4. (Very important and easy) If I hover over the Reindent Code, it should have a grey background, like hovering over the Undo does.


5. (Moderately important but easy) Please add a 'tooltip' when I hover over it saying, "Click this to adjust the text so that blocks marked by { and } have their contents indented by 2 spaces.


6. (Moderately important but perhaps hard) If I reindent twice in a row in most examples, and then undo, the layout information is lost. If this is hard, leave it and raise an issue, as this is rather obscure and an unusual case.

7. (Moderately important and maybe hard) If at all possible, could the cursor be put back on the same code that it was on (which may not be the same line). This may need to be done as a separate issue and pull request if it is too difficult.

WeiqiMaster:
I have fixed them. Also now the "Sync Diargam" button in the "Draw" menu will also display the cursor.

TimLethbridge:
This is getting closer. However

1. The /* stuff */ type of comment doesn't work quite right. Look at the class Canvas in the GeometricSystem example. There is a comment /* implementation */ where the end comment seems not to being noticed.

2. If there is a line like State Machines / Collision Avoidance line 47.

```
					Neutral { selectPark/{ setDriveSelected(false); } -> Park; }
```
That has more end curly brackets than starts, it won't work. The rule about keeping the start and end curly brackets on the same line should only be followed for matching pairs. In this case the second closing one needs to be on a new line and follow normal indenting rules.

I could tell there was a problem of type 2 above by looking at the last line of each example, that should have the } in column 1 after reindenting.

PR #1628, opened 2020-08-20
Fixes #1606 "State machine called timer will generate bad Java if there is an after statement somewhere"

### LINKED ISSUE #1606: State machine called timer will generate bad Java if there is an after statement somewhere
The following code will fail java compilation because a variable with an enumerated type called timer will be generated as will an internal variable for the after statement.

Possible solutions:

1. Add a warning to avoid state machines called 'timer' and suggesting the user rename them something else. 
2. Avoid the clash by renaming any state machine called timer, e.g. to _timer
3. Renaming the clashing variable in the generated code (might require changing a lot of tests tests)

Option 3 might be best

```
class X {
  timer {
    s1 {
      after(1) -> s2;
    }
    s2 {
    }
  }
}
```

Will generate:

```
Processing -> timercrash.ump
timercrash.ump:103: error: enum types may not be instantiated
      timer = new Timer();
              ^
timercrash.ump:104: error: cannot find symbol
      timer.schedule(this, (long)howLongInSeconds*1000);
           ^
  symbol:   method schedule(TimedEventHandler,long)
  location: variable timer of type Timer
timercrash.ump:109: error: cannot find symbol
      timer.cancel();
           ^
  symbol:   method cancel()
  location: variable timer of type Timer
```

WeiqiMaster:
Generate an error message when a state machine is named as "timer". Update manual page.

### DIFF
--- a/build/reference/9075UserDefinedStateMachineCannotBeNamedTimer.txt
+++ b/build/reference/9075UserDefinedStateMachineCannotBeNamedTimer.txt
@@ -4,10 +4,10 @@ noreferences
 
 @@description
 
-<h2>Umple semantic error reported when a user-defined state machine is named 'timer' keyword</h2>
+<h2>Umple error reported when a user-defined state machine is named 'timer'</h2>
 
 <p>
-In Umple, 'timer' is a reserved keyword. It is used to define the Timer in TimedEvent.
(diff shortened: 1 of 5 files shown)


TimLethbridge:
I just made a couple of minor adjustments to wording in the manual page and the error message.

There are leftover System.out print statements in 
cruise.umple/src/UmpleInternalParser_CodeStateMachine.ump at lines 1081 and after 1085

Please fix these.

PR #1637, opened 2020-08-27
Fixes #1629 "Umpleonline readOnly argument setting for embedding allows certain changes such as tab adding"

### LINKED ISSUE #1629: Umpleonline readOnly argument setting for embedding allows certain changes such as tab adding
The following page shows some features of UmpleOnline allowing embedding. This is a general feature and the following shows examples only

http://cruise.eecs.uottawa.ca/umpleonline/iframeexample.html

However, certain changes are needed;

* It should not be possible to add or delete tabs in embedded mode
* The visit count at the bottom should not be displayed if noChrome is set

Also in read only mode it should not be possible to load examples from the menu

WeiqiMaster:
Disable createTab and loadExample feature in readOnly mode.


TimLethbridge:
It is still possible to delete a tab when readOnly is set. It is also possible to rename a tab. Both of these need to be blocked. This PR fixes adding a tab but not deleting or renaming.

TimLethbridge:
After the latest changes, I am still able to double-click on a tab to rename it.

TimLethbridge:
OK. I see the renaming has no effect. I think that is OK.

Issue #1634, opened 2020-08-27
Avoid the 404 error when generating javadoc, and actually show failed compilation results


TimLethbridge:
When generating Javadoc, even if the Umple model is perfect, the documentation might not be generatable.

This can happen if a state machine refers to a non-existant attribute or method, or if there are any compilation errors in methods in the model. It is simply that javadoc generation actually attempts to compile, and fails.

Currently a page appears 'A page you requested was not found' ... in other words a 404 error page appears.

The solution would be to detect the failed compilation and actually show the javac error messages. It could look something like 'Unable to generate Javadoc because there is uncompilable Java code in the model. This can happen when methods cannot be compiled, or when state machine guards or constraints refer to non-existant attribiutes or methods. The following are the javac compilation results. You should try to correct the embedded Java that is resulting in these messages'. Ideally line number references would result in clickable items that could select the relevant line in the model.

TimLethbridge:
New versions of javadoc seem less sensitive, so polishing this issue is not worth it. Once we update the main server to newer Java, this issue will not be a problem any more.

PR #1638, opened 2020-08-27
Minor fixes. Contrb to #1634

TimLethbridge:
As per issue #1634 if there is an embedded method in Umple code that is uncompilable, Javadoc can not be generated and the UmpleOnline feature to generate javadoc would fail and display a 404 message.

This PR fixes that by displaying a sensible message. It also displays the actual syntax error. However, currently it does NOT display the Umple line where the syntax error occurred. That would require extra work (compiling with -c - will map the java error number to the umple line number, but javadoc doesn't do that). So there is more work to deal with #1634 properly, but this will at least avoid the confusion that currently results.

 The 404 message was also misleading, so it is fixed here too. It referred to Models 2015, and needed updating.

A glitch in a manual page from a recent PR is also fixed.

### DIFF
--- a/build/reference/5140SettingUpTasks.txt
+++ b/build/reference/5140SettingUpTasks.txt
@@ -90,9 +90,9 @@ noreferences
 <br/>&nbsp;
 
 
-<h2>How do responders respond to a task</h2>
+<h2>How do responders respond to a task?</h2>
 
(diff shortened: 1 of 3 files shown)


TimLethbridge:
It turns out that newer versions of Javadoc are less sensitive to bad methods, so this issue solves a problem that may no longer occur in recent java versions. Nonetheless it does no har,

PR #1651, opened 2020-09-25
Mixin youtubes to StructureofUmpleCode page

gublan24:
This is update to user manual. The page "Structure of Umple Code" will include YouTube examples for mixins with Umple classes, associations, state machines and aspect oriented programming.   

I don't see real benefit to add the code used in the example. But I will add it to the YouTube description instead.


TimLethbridge:
@gublan24 You modified the wrong file! You committed the generated file, not the source file. You should edit  build/reference/0122structureOfUmpleCode.txt instead

See 0101gettingStarted.txt for an example of how the @@videoURL tag can be simply added

Issue #1401, opened 2018-10-25
Generate natural language descriptions of associations, etc.


TimLethbridge:
Students (and developers) often have a hard time understand the true consequences of their decisions regarding associations. It is suggested that when hovering over an association in the diagram (all kinds), the text that pops up should describe it in detail. Also a generation output could be created that would describe the whole model in natural language.

For example a
```
class Company {1 -- 1 BoardOfDirectors;}
class BoardOfDirectors {}
```

Could generate

```
A company has a Board of Directors. It must always have exactly, one, and can never have none.
A Board of Directors must have a Company.  It must always have exactly, one, and can never have none.
There must always be an equal number of Companies and BoardsOfDirectors, connected as Pairs. 
When a Company is created is BoardOfDirectors must be created at the same time.
When a Company is deleted, its BoardOfDirectors is deleted at the same time.
```

TimLethbridge:
For a 1--* multiplicity, such as the following

```
class Mentor {  name;}
class Student { name;}
association { 1 Mentor -- * Student;};
```

We should see
```
A Mentor has some Students. It can have none, and there is no upper bound defined.
A Student has a Mentor. It must always have exactly one, and can never have none.
When a Student is created its Mentor must be assigned.
```

In all cases there should be a method created in the class association to obtain this data. The data should also be transmitted to the diagram generation tools, so that when the user hovers over any association, the information appears. This should at least work in GraphViz view.

TimLethbridge:
To generate text, use Umple's template capability.

TimLethbridge:
Good progress in this in PR #1668 

A few more cases need to be dealt with.

1. The 'TEMPDEBUG' now needs removing and replacing with just a blank line

2. In the 'Canal' example in Umpleonline there is a 1..* -- 2 association and several others of interest.
For the 1--* end between SegEnd and Segment, for example, there aren't any messages displayed. Use this example and other to find cases that need improving

3. For 1-* ends:

```
A SegEnd has some Segments. It must have at least 1 and there is no upper bound defined.
When a SegEnd is created, at least 1 Segment must be assigned.
```

4. For cases where the 1 in the above is greater than 1, the relevant integer should be present.

5. For the case where there is an upper bound (e.g. 1--3 instead of 1--*, or 0..* or simply *, instead of saying 'there is no upper bound defined', say (where n is the upper bound, and X is the relevant class) 'there is an upper bound of at most n Xs'

[...]

PR #1668, opened 2021-03-10
Changes the label and tooltip variables on the files

Stacie0805:
This PR is intended to resolve issue #1401 by improving the tool tips on Graphviz associations so they can express natural language concepts


TimLethbridge:
I ran full build and notice that the following message is displayed:

```
    [junit] TEST cruise.umple.implementation.gv.GvCdGeneratorTest FAILED
    [junit] TEST cruise.umple.implementation.gv.GvErdGeneratorTest FAILED
```

Looking at the QA log in dist/qa/cruise.umple/cruise/umple/implementation/gv/150_GvCdGeneratorTest-fails.html I see

```
display_compositions Failed at:29 expected:<...tip="A 0..1 -> * C["];]> but was:<...tip="A 0..1 -> * C[]> 
```

and also a failed test for display_associations

The relevant tests are in 

cruise.umple/test/cruise/umple/implementation/gv/GvCdGeneratorTest.java around line 65

and related files. You will see there is a comparison of generated output with a file in Associations.gv.txt

You would need to update that .txt file to correspond with what is being generated now.

However, note that there are additional changes needed to finish this issue. I will add another comment. for those in the issue itself

TimLethbridge:
Here are my comments so far:

1. A function called hasRoleName should be returning a Boolean. Any function starting with 'has' should be a Boolean Function. Change this to 'getClassWithRoleName'

2. I see you have an almost identical function at line 395 that adds the plural. It would be better to use a single function, with an argument 'isPlural', so that code only has to be changed in one place. Also, this is putting an extra s on role names that already end in an s.

3. Your comparison of class with role name to see if there should be a separate role name is not working because role names are lowercase! You need to do a case insensitive comparison.

Do these fixes and then show me again.

TimLethbridge:
I am cancelling this PR. It is replaced by #1683  which has the same code, is mergable.

PR #1679, opened 2021-04-12
Refactoring GvEntityRelationshipDiagram generator

gublan24:
This PR refactors GvEntityRelationshipDiagram generator into a feature "EntityRelationshipGenerator". Since this generator does not have test cases, I did two full compilation of Umple; one with enabling the feature (EntityRelationshipGenerator) and another without using the feature/mixset. The two compilations give the same result.


TimLethbridge:
There is indeed a test for this:

cruise.umple/test/cruise/umple/implementation/gv/GvErdGeneratorTest.java

gublan24:
I see there are 16 failures that always there. However, when I delete the line "use EntityRelationshipGenerator;", which should ideally cause tests in GvErdGeneratorTest.java to fail has no effect. 

Are these test cases excluded from checking when build Umple? I use dev-tools/bumple to do the complete compilation.

TimLethbridge:
I know the student working on class diagrams caused the ER testcases to fail, since there is some common code (take a look). So I am sure the test cases are being run. I don't have time to investigate.

gublan24:
The test cases not failing also the generator is not there because of the code `System.exit(-1);` in the UmpleModel. This code kills the thread of compilation. And because there is no errors, Junit gives false positive results. See the effect when removing it in https://github.com/umple/umple/tree/disabledCodeGvEntityRelationshipDiagramAsFeature.

PR #1716, opened 2021-05-25
Gradle improvements for Issue1609 spring2021

RalphNgassa:
Fixed most gradle build errors and added new commands to be up-to-date with the cheatsheet


TimLethbridge:
It seems that a lot more progress is needed.

For reference I am running under Gradle 7.0.2

1. There are two warnings about features that are scheduled to be removed in Gradle 5.0 (a bit odd since we are already on 7.0). The first is the 'Separate output directories for each language'. I would like to see some analysis of this. What would have to be done to fix it.

2. The second warning is. FileCollection.stopExecutionIfEmpty() method has been deprecated.

3. The file build/umpleversion.last.txt is getting modified in the build. This is only supposed to happen during a release.

4. testall still fails with an 'illegal start of expression' error. This needs to be analysed and explained. Why is it not happening in the ant build? 

5. Until item 4 above is fixed I can't see whether the other tests are being run by testall (all the regular junit tests as well as the manual and example tests). 

6. The packageumpleonline seems to create a slightly different minified file than in the ant build. I would need an explanation of this as it could cause problems

For reference, the commands I ran were

```
git checkout issue1609_spring2021
gradle --version
./gradlew --warning-mode all fullbuild
git status
./gradlew --warning-mode all packageumpleonline
./gradlew --warning-mode all testall

```

When pushing more commits to this branch and pr, please make comments here that reference the above items. I at neast need a resolution of items 3 and 4 and confidence that the rest of the tests are being run before we can merge this. If Items 1 and 2 aren't resolved then please comment on why resolution is not necessary and/or what would be needed to resolve them.'

TimLethbridge:
I just ran the same commands on the main linux server. Exact same results, I am enclosing a complete trace file. Note that the very first thing gradle 7.0.2  does is download gradle 4.10.2 ... presumably because of the deprecation isssue? This needs explanation.

[FullGradleLogInPR1716.txt](https://github.com/umple/umple/files/6598678/FullGradleLogInPR1716.txt)

TimLethbridge:
I have discovered why gradle 7 downloads gradle 4 on start ... it is because of the gradle/wrapper/gradle-wrapper.properties file

See https://docs.gradle.org/current/userguide/gradle_wrapper.html

Presumably it would be necessary to update the proper to point to a newer gradle. But then you would also need to do the necessary migrations to make it work in the newer gradle

TimLethbridge:
You can use the following command to try to figure out problems

```
gradle help --scan
```

TimLethbridge:
Yes another problem, After running the commands described above, umpleonline doesn't work (likely wrong umplesync.jar or some similar problem). No diagram appears for example. Also the version displayed at the bottom left of the screen says v@UMPLE_VERSION@ which implies that the cleanup at the end of the build is not happening properly. I had to run an full build using ant and then a promote to umpleonline using ant to restore to the correct state.

When testing the gradle builds please verify UmpleOnline afterwards as well as the command-line compiler. Check the times of update of the jars. etc. as well

TimLethbridge:
Because of the changes to certain parts of the compiler the test cruise.umple.implementation.EqualsTest is failing. The build therefore cannot run until the relevant test is also fixed.

PR #1734, opened 2021-06-02
Ensuring URL example argument selects dropdown example. Most of #1703

### LINKED ISSUE #1703: Direct browser navigation does not make dropdown selections based on the URL in UmpleOnline
## Summary

Within the UmpleOnline, the users can directly use a navigation URL, paste it into the browser, and press ENTER for easy access to a specific example page.  When the user lands into that specific URL, the UML code and the class diagram are displayed according to the example mentioned in the URL ([example](https://try.umple.org/?example=2DShapes)).

But the dropdown values are not selected according to the URL. The example dropdown is selected as "Select Example". For an URL (https://try.umple.org/?example=2DShapes), the example dropdown should be selected with the value "2DShapes".

## Steps to Reproduce

1. Copy the URL (https://try.umple.org/?example=2DShapes), paste it into the browser, press ENTER.
2. Now, under the "Tools" section, the example dropdown is loaded with "Select Example".
image

## Expected Feature

The system should show the specific example as a selected value in the dropdown.

RalphNgassa:
Updated umpleonline scripts to make the examples dropdown menu selection match the example loaded through the URL. Fixes #1703


umpleonline/scripts/umple_page.js
```diff
@@ -1092,6 +1108,7 @@ Page.getSelectedExample = function()
   if(theExampleType == "cdModels") {
     var requiresGvClass = false; // Some class diagrams  are too complex to edit
     inputExample = jQuery("#inputExample option:selected").val();
+    sessionStorage.setItem("savedSelection", inputExample);
```
TimLethbridge: Please avoid using sessionStorage for this feature. I believe this is leftover code from an earlier attempt.

TimLethbridge (changes requested):
This does not work in Safari. It works on a Mac when I tried only in Chrome, Opera, Edge, Firefox, but it needs to work in Safari

Possible causes:
  - The jquery code in setExamples might slightly different behaviour in that browser.
  - The code is attempting to set an example that doesn't exist in 3 out of the 4 menus (it only exists in one of them); maybe when it fails safari behaves differently, Would it be possible to only set if it exists?
  - Maybe initexamples is being done after setexamples?

umpleonline/scripts/umple_page.js
RalphNgassa: Sessionstorage was indeed residual code and I made sure to remove it. I also modified the logic on which setexamples is operating so it might hopefully work on Safari now: Rather than searching for the appropriate value in all of the 4 menus, this search will stop once the requested example is found. This means that if the requested example is found in inputExample2, it will first go through inputexample and then to the former before returning nothing and thus not going through inputexample 3 and 4.

TimLethbridge (approved):

PR #1735, opened 2021-06-02
Issue1695_Survey

Svetlana154:
Added survey feature that may occasionally appear in the same section as alerts and TOTDs


TimLethbridge:
We are very, very close now. However I am still having a hard time with actually consistently testing it, and with 'testability".

So here's my sequence. The items marked TOFIX need addressing (I have put this tag afterwards).

1. Reload, sc, reload, sc. Not sure why/if I had to do that twice. But I believe my browser should be 100% in the 'pure' state now. This should not normally be needed in a new browser. But I believe it is just to clear any previous state.
2. Do a little work
3. sp (I believe this should set the timer and edit count. I am NOT expecting it to refresh the screen and get rid of my edits, but it does. TOFIX. It would be better if it was silent other than setting some variables so that one more edit would bring up the message. Furthermore it *immediately* brings up the survey message after the refresh, when I was expecting to do one more edit (to verify that editing in fact brings it up). TOFIX.
4. I hover over the one-line message and the larger box pops up. Good. I move cursor off it and it goes away. Good.
5. I click on 'not today'. The box goes away temporarily but I was expecting the one-line message to also go away. TOFIX

For the backdoors, I was expecting them to be super easy to understand and to use for manual (and later on, automatic) testing.
 * sc. I was expecting this *just* to clear cookies and clear any state
 * sp. I was expecting this *not* to set any cookies at all, and not to have any effect after a reload, but just to affect the current session 

The fact that sp sets a cookie is, to me, not helpful and confusing. If after doing sp and you refresh, the fact that sp was done should, I think be entirely forgotten. If you have a reason for storing and refreshing, let me know, but it doesn't seem to so that I was looking for and doesn't seem to be useful.

I think I was not expecting cookies to be used as much in this feature. I was expecting to have:
* A cookie to indicate that a each particular survey had been started (to ensure that it wasn't prompted-for again)
* A cookie to indicate that a survey had been specified as 'not today' that it wouldn't appear today (but perhaps if the dice rolls, then maybe tomorrow and afterwards).

PR #1737, opened 2021-06-07
Improving colour of selected states in UmpleOnline more S less V Fixes #1705

### LINKED ISSUE #1705: Selected top menu items should be highlighted in UmpleOnline
## Summary

Within the UmpleOnline, the users can switch the page layout and change the diagram options. Users can click at any of the letters ("E", "G", "T", "D", etc.) at the top left corner to view and retain the change. Once clicked at any of these letters and the mode is changed, the system does not highlight the current selection. The system should highlight the current selection of letters (mode) so that the users are always aware of the mode that they are working on.

## Steps to Reproduce

1. Click at any of the letters in the top left corner.
image
2. Multiple modes have been toggled but no mode is highlighted. The system does not explicitly show the current selection set of modes.

## Expected Feature

The system should highlight the current selection set to explicitly display the user's mode.

shoevancar:
Updated the Umpleonline code to highlight the top menu in deep color when selected. Fixes #1705

### DIFF
--- a/build/umpleversion.last.txt
+++ b/build/umpleversion.last.txt
@@ -1,1 +1,1 @@
-version :   1.30.2.5469.b99e94bb0
+version :   1.30.2.5248.dba0a5744
--- a/build/umpleversion.last.txt
+++ b/build/umpleversion.last.txt
@@ -1,1 +1,1 @@
-version :   1.30.2.5248.dba0a5744
+version :   1.30.2.5469.b99e94bb0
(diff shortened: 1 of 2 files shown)


TimLethbridge:
The umpleversion.last.txt should not have been committed, so I reverted it.

The new proposed colour was made darker but asince it was too unsaturated (trending towards grey) it didn't look nice. I changed so it was more saturated and I also reduced the change in brightness. This should still stand out better than the previous colour, but looks nicer.

Also the PR description was not as useful before. I added the fixed message, and more clearly state what it does.

PR #1746, opened 2021-06-11
Modified a few number of umple source files to prevent vml from being generated (Draft)

RalphNgassa:
I excluded vml.ump from the files used to generate java in cruise.umple/src-gen. Also deleted some lines of code in the ant build files (build.umple.xml and template.xml) to prevent the packaging of umplevml.jar in the dist folder.
I also noticed there is a vml folder present in cruise.umple/test so to completely remove vml from umple, I believe it has to be deleted too. That's also the reason why the CI checks are failing.


TimLethbridge:
I would like to clean out all references to Umple.

If I do in this branch

```
grep -i vml *
```

I get references to VML in gradle

I also note

* umpleonline/vml.php
* mentions in cruise.umple/src/DiagramMaster.ump around line 103
* file cruise.umple/src/Vml.ump
* file cruise.umple/src/Vml_Code.ump
* several that match umpleonline/scripts/*vml*

There may be others. All need deletion followed by testing to ensure there are no negative impacts

TimLethbridge:
This looks pretty good
I noticed that build.gradle and a Readme file still mentioned vml so I removed it.

PR #1762, opened 2021-06-14
Fix gradle build error messages

RalphNgassa:
I tried fixing the minor issues that were recent since I last worked on gradle. Still some depracation messages appearing but that can be easily fixed. I will try fixing those tomorrow during my first work hours, as I working on other issues today. You'll also notice some changes present in two template files that were made to resolve failed unit tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,9 +50,11 @@ def buildAliases = [
    ],
    'quickbuild' : [
         'cleanUpUmple',
+        'codeGen',
+        'umpleParser',
         'resetUmpleSelf',
         'compileJava',
-        'jar',
+        'packageAllJars'
     ],
    'fullbuild' : [
         'cleanUp',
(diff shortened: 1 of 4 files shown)


RalphNgassa:
I edited template.xml to ensure that the Association sorted test is run by the ant build script and I also made sure that it was the only test file not included that why it's the only additional test file. Given that the template files were already modified on this branch I found it appropriate not to create a new branch.

TimLethbridge:
OK. Major improvements.

Some more suggestions.

1. I notice that when I say gradlew tasks it does not list fullbuild in custom tasks. This is important since people will otherwise not know about it.

2. Is there a way the 'build' task could do the same thing as a full build. I ran the build task as-is and got many errors (I suggest you try)

3. Could there be a quickbuild task that does the needed steps up to creating jars, but does not test?

4. I noticed the following error messagein a full build:
```
> Task :javaTestbedCompile
Execution optimizations have been disabled for task ':javaTestbedCompile' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/tcl/nobackup-tmp/CheckoutForUmplePatchTest3/umple/dist/gradle/bin'. Reason: Task ':compileJava' uses this output of task ':javaTestbedCompile' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.0.2/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```
I notice some other tasks give a similar message. Please investigate.

5. I notice that phpTestbedCompileUmple generates an xml file and dumps it to the output. A few other targets do the same thing. Please look into what happens in Ant building. We don't want the build log so cluttered.

6. Can the unit tests (I think task 'test') be done before ethe testbed tests please?

I won't require that every one of these points be solved before merging this pull request, so if you would like me to merge before we proceed to a new branch and new PR, I am OK to do that. Let me know.

PR #1771, opened 2021-06-22
Issue1743_UpdatingJQuery

Svetlana154:
Updated jQuery to v3.6.0 and jQuery UI to v1.12.1. Fixed appearance of UmpleOnline.


TimLethbridge:
A few things to please look at:

1. The font choice is not as nice as it was before. I think we go back to the oriiginal main font, with the new choice as fallback (there is a way to list alternatives in css). Even the monospaced font in the main text editing area looks not as nice. The text in the buttons at the top has a lower baseline (e.g. not centred in the buttons)
2. On Safari the 'SAVE & LOAD' menu choice is still black when not selected and white when selected (the little triangle just before it is correct). Please examine if there is any subtle difference; it may be the amersand perhaps or something else subtle.

TimLethbridge:
Regarding item 2, save and load black vs. white, the problem goes away in the fall release of Safari (I downloaded a beta). I would, however, be reluctant to have to wait that long so if we can figure it out, it would be helpful

TimLethbridge:
So I pushed a fix to the white/black issue

Please delete previous JQuery versions so we can keep the size of the repo down and not have unused stuff hanging around

Please then test thoroughly, including with URL options.

If all is good at that point you can remove the Draft status and I will do a final test before merging.

Issue #1744, opened 2021-06-11
Reduce output of generateAndCompilePhp and generateAndCompileJava to make output better - Verify in Gradle when fails


TimLethbridge:
Currently in both ant and gradle builds, the output of the two targets listed in the title are extensive. Each example in the manual and umpleonline examples is compiled. It would be better if most of the output was suppressed unless a 'verbose' flag was passed to the target. Can the 'processing' 'Success', Skipping' 'Compiled to target' be supressed? and for php the 'No Syntax errrors detected' Only errors and warnings are really useful (but if there are any the relevant file would need to be displayed too).

Maybe in fact for each compile just a dot could be output only unless something goes wrong. This would make it easier to debug and search the log for problems. A count of total examples could be printed perhaps at the end.

TimLethbridge:
I noticed that in the Gradle build there a some failures from this when compiling PhP yet the build does not fail.

TimLethbridge:
This has been completed

PR #1773, opened 2021-06-24
Updated the gradle build script

RalphNgassa:
Unit tests will now be executed before testbed tests


TimLethbridge:
I ran this, and I don't see the template tests executing, and I can't fully comprehend how the changes you have made here would cause the template tests to run right after the creation of the jar, which should be happening.

It might be that nothing has changed so the tests don't need to run. But then it would be best if no tests run.

It might be that there is a confusion about the meanings of the word 'template' which is used in several totally different ways in Umple (the template tests in ant task template.text are totally separate conceptually from the notion of UmpleTL templates

I think that, like some of the other current PRs I am going to have to study the gradle build in depth to be able to ascertain whether it is correct.

RalphNgassa:
There has been quite a good amount of changes made on the gradle build script. Starting with making sure that all testbed tests are reported. All reports are located within the dist/gradle/reports folder with java testbed test results found in testbed_java, php ones in testbed_php and ruby tests in testbed_ruby. A task to create a symbolic link to the umple.jar file was added too.  

You'll also notice that I modified Statemachinedo.txt, found in the testbed/test/cruise/runtime/expected folder, which contains expected error messages for the runtime tests. This modification was made because the error message changes with more recent java versions like the one gradle uses. The ant java compiler, on the contrary, uses an older java version and therefore outputs a different error message. I think it will be best to push the updated version of the ant build on this branch but only for the purpose of passing all CI checks.

Furthermore, a few comments were added just to remind myself (or another developer working on gradle in future) of the different lines that are scheduled to be deleted once gradle fully overtakes ant. Presently, gradle moves most generated files to dist/gradle just for the purpose of not overwriting the ones generated by ant.

Finally, the a few other tasks were created to launch the automated UI tests on either umpleonline local or public. But this branch does not contain the updated UI test files.

TimLethbridge:
This will close #1744 when finally merged because it will soon have the shorter output for the example tests

Issue #363, opened 2015-08-26
Umpleonline user interface tests


Nava2:
Originally reported on Google Code with ID 464
Owned by @CraigBryan

---

Create an enhancement to the automated build to run Capybara tests of the UmpleOnline
user interface. This would test both Umpleonline details as well as loading some examples
and testing them

---

Reported by @umple on 2013-12-02 18:06:11

Nava2:
Reported by @CraigBryan on 2014-07-02 18:21:33
- **Owner changed**: @CraigBryan

Nava2:
Reported by @CraigBryan on 2014-07-14 20:06:33
- **Status changed**: `Started`

Nava2:
Added the start of a test suite for umpleonline with revision r4328. It runs on phantomjs,
so it runs headless. A build script was also added. It requires phantomjs, and a few
rubygems (capybara 2.4+, rspec 3.0, and poltergeist 1.5+) to run.

Run "ant -f build.websitetests.xml help" from the build folder for instructions on
the build script.

The tests are broken into features, and not all website features are covered. The help
target in the build script lists the features that are covered.

There are a number of tests that are "pending" These are tests that are planned, but
not fully implemented due yet.

Some tests have been failing on the test server (ie a non-local server). These failures
are likely due to the longer response time associated with a non-local server. 

---

Reported by @CraigBryan on 2014-08-22 16:05:20
- **Status changed**: `Ongoing`

TimLethbridge:
I would like to push to get this finished. Craig Bryan was almost done with it. But it was never activated. He can help but is graduating soon. it would likely be best if this could be activated however after we move from ant to grade #751 so that we only need to run UI tests in relevant cases; this can be handled by dependencies

TimLethbridge:
I think only activate this on travis and make it optional when myenv=local

CraigBryan:
If I remember correctly largest hurdle in enabled this feature is the repeatability of the tests. When this was run on a non-local server, the speed of responses from the server would often make tests fail. I will take some time to look at these tests and figure out what we need to do to move forward.

TimLethbridge:
So the status seems to be that there is an ant script build.websitetests.xml 

and there are tests themselves at umpleonline/testsuite/spec

We need to polish this work so we can activate it in Travis and also convert it to Gradle

RalphNgassa:
[...]

PR #1783, opened 2021-07-02
Adding automated user interface tests. Progress on #363

RalphNgassa:


RalphNgassa:
The tests now run on chrome headless through the cuprite driver. This choice was made based on the fact that phantomJS has been discontinued about 5 years ago now and the test result won't reflect umpleonline behaviour on the most recent browser versions.

You might want at the comments under issue #363 which provide guidelines on how to run the tests with ant.

TimLethbridge:
Requires running the following.

 - gem install rspec capybara cuprite selenium-webdriver
 - ant -f build.websitetests.xml "-DhostName=http://cruise.local/umpleonline/" doWebsiteTests

Note that it may not yet work perfectly on all platforms, so we will continue work

Issue #1710, opened 2021-05-19
Improving ability of Umple end-users to use UmpleParser using agreementParser as a case study and test


TimLethbridge:
Within the Umple top level directory is UmpleParser. The code for this is incorporated into the umple build using the ant target UmpleParser in build/build.umple.xml starting at line 149. The parser files are independently built and then copied into the main Umple compiler build

This is intended to be a set of Umple files that can be incorporated into any program that wants to do parsing; not just the Umple compiler. However the Readme.md file says 'Currently, this can only be compiled as part of the main Umple build'. 

However, it should just be possible to include one .ump file in an Umple program, instantiate the parser, add some rules into the parser, and then call upon the parser to parse some text.

A demonstration project called AgreementParser was created to demonstrate this. The project is at 
https://github.com/umple/agreementNegotiation

The tasks of this issue is to:

* Understand the UmpleParser code and its building and refactor as needed so that the same code from the UmpleParser directory can be used in any system.

* Complete a functioning version of the agreement parser so that the .txt files in testAgreements can be parsed according to the grammar in src/Agreement.grammar.

TimLethbridge:
It has been suggested that those working on this watch the following video, which although old, might help

 https://www.site.uottawa.ca/~tcl/video/ExtendingUmpleGrammarAndPopulatingMetamodel.mp4

RalphNgassa:
The umple parser doesn't properly parse .txt files, even after adding a new set of rules such those within the 'Agreement.grammar' file. I think the problem might come from the 'parse(file)' method contained in 'RuleBasedParser.java'. The RuleBasedParser.java file can be found in the cruise.umple.parser.analysis package.

TimLethbridge:
What's needed is the ability to have a simple program with just two three core lines.

The first line would be a an umple use statement that incorporates the needed umple parser code as a single file (which may in turn use other umple parser files

The second line would be a call to a method to read in a grammar file (or, alternatively a few lines to specify grammar rules directly).

The third line would be a call to a method to parse a text file (passed as an argument) and generate a token (with subtokens). The program would then do what it wants with the parsed results.

Right now, the problem is that there seem to be glitches in this ... some needed code seems to be only in the umple compiler under cruise.umple/src, whereas none of the needed code should be there if the parser is to be usable independently.

TimLethbridge:
[...]

PR #1790, opened 2021-07-08
Extension to umple build scripts regarding #1710 draft

Emmacho:
Extension to umple build scripts that can be used to build umple parser only. #1710_Draft. Use "ant -Dmyenv=local -f build.umple.xml buildParser" to build


TimLethbridge:
Closing this because we decided on a different approach (build of a Jar in the UmpleParser directory)

PR #1810, opened 2021-07-30
Adding ant build script to build umple parser

Emmacho:


TimLethbridge:
Merging of this PR has been delayed for an extended period, since it is not well documented.

My understanding is the following: If within Umple end-user code, one wants to be able to parse, one needs to call Umple's parsing features, which are not part of the compiler, but are a separate feature writen in Umple. Therefore this PR builds a jar that can be incorporated in end-user Umple code to enable parsing.

But testing and example cases are needed.

PR #1834, opened 2021-08-11
Fixing State Machine Diagrams scaling problems.

### LINKED ISSUE #1619: Complex state machine will not render properly
The state machine in the attached file does not properly render.

[ProblemComplexStateMachine.ump.zip](https://github.com/umple/umple/files/5016535/ProblemComplexStateMachine.ump.zip)

A. In a local image (Graphviz/dot 2.44)) if guards and actions are displayed. It chops off at the bottom (See image A below) If actions are displayed it displays a lot of overlap and items that are not in their correct parents (See images B and C below). Overlap of transition labels is to be expected and can't be avoided if there are many, but overlap of boxes is the problem. It may be that complex actions need further truncation/ simplification.

B. In the Docker image (Graphviz/dot 2.38)), there are two problems. 1) It fails to render the state machine and displays the following, when it is supposed to display a special. Note: There is an intention of upgrading the Docker image version, so the symptoms may change.

```
Diagram is out of synchronization with the text due to selecting Manual Sync or an error in the text that has caused the compiler to produce no output.'. However it is supposed to render.
```

2) If another state machine is displayed before the attached state machine is loaded, the old state machine continues to be displayed. This likely means that the spec

C. On the main server (Graphviz/dot 2.38) ), it does not render at all (but that is using an older graphviz). It displays the special message (in an SVG file, that is supposed to be what appears in the Docker image, see above)

```
dot (part of GraphViz) does not appear to be installed on the server 
...
It is also possible that an error has occurred in Graphviz diagram generation. 
```

ScreenShotACutOff

ScreenShotBOverlapping

ScreenShotCOverlapping

RalphNgassa:
This pr should fix #1619 and the draft label removed once the docker part of this issue is resolved.


TimLethbridge:
I have tested the docker image and all is well with this

PR #1855, opened 2021-08-24
Traits library capability

### LINKED ISSUE #80: Explore incorporation of more patterns into Umple, including a general mechanism for creating patterns
Originally reported on Google Code with ID 178
Owned by @aforward

---

The GoF patterns will be explored, but other patterns too.

Delegation is of particular promise, along with its related patterns.

Ideally code generation for the patterns would generate Umple (at the metamodel instance
level), not base language, so there would be no need to extend the code generation
templates.

This work will also explore the potential to describe patterns abstractly so new patterns
can easily be written.

See also ticket 127

---

Reported by @umple on 2011-06-25 02:19:34

### LINKED ISSUE #1393: Consider allowing multi-level modeling: Native support for the abstraction-occurrence pattern
The abstraction-occurrence pattern is when instances of a class in a model has occurrences that are modelled as instances of another class. Examples include:

TVSeries -- Episodes
ProductCategory -- Product
BookTitle -- CopyOfBook

In programming languages, both classes in each pair need to be at the same 'level', i.e. classes. However the multi-level modeling concept suggests that in models the abstraction (e.g. BookTitle) could have instances (specific BookTitles) that are the classes for their copies. The abstraction would become a metaclass.

Beginner modelers tend to get confused as to where the 'instance-of' relationship lies, so Umple support would be useful. But we also would have to be careful to avoid adding confusion, or generating code that is overly complex.

One suggestion would be to tag the abstraction-occurrence association with a stereotype. Then we could generate delegation methods automatically such that attributes of occurrence instances 'inherit' the attributes of their related abstraction instance.

TimLethbridge:
This adds a 'reusable' directory under umplibrary to contain files that will be always available in use statements that are prefixed by lib:

These can contain useful traits, and some initial examples are given

While setting this up a bug was fixed: Previously missing files on the command line were silently ignored.

### DIFF
--- a/UmpleParser/src/GrammarParsing_Code.ump
+++ b/UmpleParser/src/GrammarParsing_Code.ump
@@ -593,7 +593,11 @@ class ParserDataPackage
       }
       else
       {
-        resourceStream = getClass().getResourceAsStream(file);
+        String resourceToLoad=file;
+        if(file.contains("lib:")) {
+          resourceToLoad = "/"+file.substring(file.indexOf("lib:")+4);
+        }
+        resourceStream = getClass().getResourceAsStream(resourceToLoad);
         reader = new BufferedReader(new InputStreamReader(resourceStream));
       }
       String line = reader.readLine();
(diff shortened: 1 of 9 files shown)


TimLethbridge:
This fixes #1393 and #80

Issue #1861, opened 2021-08-30
System saves a task with a name that contains dash ("-") but cannot retrieve it in UmpleOnline


shoevancar:
## Summary

Within the UmpleOnline, the users can create a task with a unique name. While creating a task, the hover message says that a task name should include only alphanumeric, “_” (underscore), and “ ” (space). But the users can still save a task name with “-” (dash) character. (Example task name: "shuvankar-001-001-009"). The system saved it without giving any error or warning message. But the user cannot reload that task (searching by name) in the future.

## Steps to Reproduce

1. Go to the task, then click "Create a Task".
![image](https://user-images.githubusercontent.com/82404135/131305212-397106a2-1c98-4e3b-a45a-2b216079d41f.png)
2. Give any name with a dash ("-") in the "Task Name" textbox and click on the "Submit Task". The system saves the task without any error message.
![image](https://user-images.githubusercontent.com/82404135/131305439-78cc94c1-c697-4766-8fe9-f06ad3647505.png)
3. Now refresh the umpleonline page. Go to the task, and then click on "Load a Task". Type the task name that you entered at step-2. Finally, click on "Load Task".
![image](https://user-images.githubusercontent.com/82404135/131305833-6c14468a-52ea-46be-9ba6-13d84912b832.png)
4. The system shows an error message saying the task does not exist. 
![image](https://user-images.githubusercontent.com/82404135/131306135-73924b6c-b4ed-4f71-8305-70a9245fbeb7.png)

## Expected Feature

The system should restrict the users from entering a dash ("-") within a task name (allowing a dash would require much more work due to the internal naming of files in the server)

TimLethbridge:
Fixed by #1890

PR #1890, opened 2021-10-21
+fixer issue1861. taskNname ne peut plus contenir '-'

sidikik55:
**IMPORTANT**: 

## Fixed issue1861

This PR resolves the problem described in issue # 1861, this correction will no longer accept the hyphen (-) character in the name of a task when creating a task or loading it.

## Description
In the correction of this issue I modified these two files: umple_action.js and umple_page.js
 - The correction of the umple_action.js file makes it possible to no longer accept the hyphen character (-) when loading a task. 
 - The correction of the umple_page.js file makes it possible to no longer accept the hyphen character (-) when creating a task.


TimLethbridge:
Fixes #1861

Issue #1867, opened 2021-09-22
Add natural language description in GV state machines


TimLethbridge:
In the same manner that natural language appears in GV class diagrams, as per issue #1401, the same should be true in state machines.

In a state after the name of the state (which appears now), it should say:

This is the initial state the object is in when created
(if it is the first state of a top level state machine and has an incoming arrow from a start circle)

This is the initial state when VVV is entered
(if it is the first state of a nested state machine and has an incoming arrow from a start circle)

State is entered if XXX occurs when in state YYY
(the above repeated as many times as needed, could be zero if the initial state)
(and after the word 'occurs' it might say ... 'and guard [ZZZ] is true' if there is one

On entry into this state an entry action is executed
(if there is one)

On entry to this state a do activity is started in a separate thread and continues until it ends or the state is exited
(if there is one)

State is exited if PPP occurs; the object then goes to QQQ
(repeated as needed and also with the 'and guard [RRR] is true if there is one)

On exit from this state an exit action is executed
(if there is one)

VVV and PPP would normally be an event, but could also be 'after m milliseconds

Stacie0805:
Does VVV represented the next state or the state we are currently in?

TimLethbridge:
VVV is a parent state.

TimLethbridge:
Additional requests from a question on Stack Overflow:

https://stackoverflow.com/questions/68034922/how-to-render-a-better-startchart-graph-on-umple

When hovering over a state, the entry actions, exit actions, and at least some of the do activities (limited by code length if excessive) should be shown.

When hovering over transitions the full guards and the full transition actions should be shown.

PR #1908, opened 2021-11-22
Work on issue #1867 - natural language in state machines

Stacie0805:
This PR is a work in progress on issue #1867 ... to generate natural langauge tool tips in state machines


TimLethbridge:
Here are my comments:

I will use the DigitalWatchNested example 

1. Instead of 'This is the state we get from X when we enter event Y', say 'This state is entered when event Y occurs while in state X'.

2. The phrase 'This is the initial state the object in X is when created'  should be normally 'This is the default initial state'. If it is a substate then add 'when entering state X' (where X is the parent state). And if the state is actually the first state entered on creation, add 'and is entered when the object is created' (e.g. this would be added for 'regular.time'). And we also need to list the transitions entering the parent state that go to the initial state by default, so for example for 'bothOff', it should say. This state is entered when event s2 occurs while in state time'

3. Some tests are failing. You should fix them (may require replacing one or more gv files'. However I would be OK if you fixed the above two first. I will then test more thoroughly before you do this.

TimLethbridge:
Closed because superceded by #2370

PR #1910, opened 2021-11-29
Show files in separate tabs has been disabled for Use model in UmpleOnline

### LINKED ISSUE #1905: Show files in separate tabs should not be present for Use model in UmpleOnline
In UmpleOnline, users can create a class diagram and generate its equivalent USE model. The system offers an option to display the generated code in separate tabs. The system hides the entire USE model when the user clicks on "Show Files in Separate Tabs".

**Steps to produce:**
(1) Select a class diagram such as "2DShapes".
image
(2) Navigate to the "GENERATE" section and select "USE Model" from the dropdown. Click on "Generate It."
image
(3) The system generates the equivalent USE model. Scroll a bit down and click on "Show Files in Separate Tabs".
image
(4) The system cannot display the code in separate tabs. Moreover, it hides the entire code and shows nothing.
image

**Expected:**
(1) "Show Files in Separate Tabs" option should not be given for the USE model as there are no separate classes generated.

shoevancar:
This fixes show files in separate tabs for "USE" models. This option will be hidden for this particular case.


TimLethbridge:
You have not committed the .js file that changed. You have only committed all-umple-min.js, which is a file generated from the originally committed .js file (it has to be committed too, but is not the main change).

TimLethbridge (changes requested):
It doesn't yet have the changed .js file, only the minified file. I can't review yet.

shoevancar:
The changes at .js file have been added. Would you please review that?

TimLethbridge (approved):
Tested and good. Will merge

Issue #1848, opened 2021-08-19
Large model does not generate valid gv diagram


TimLethbridge:
The model below does not generate a graphviz class diagram that displays in Umple

(When you figure out the generic cause, please find a simpler model that has the issue and add it; also please rename the issue to better describe the problem)

```
class String
{
}

class Integer
{
}


class Matlab
{
  0..1 -> * Model model;
}

class BlockDefaults
{
  String blockDefaults;
  * blockDefaults -- 1..* Model model;
}


class Model
{
  String name;
  1..* model -- * AnnotationDefaults annotationDefaults;
  1..* model -- * LineDefaults lineDefaults;
  1..* model -- * System system;
  * models -- 1..* Simulink simulink;
}

class AnnotationDefaults
{
  String annotationDefaults;
}

class LineDefaults
{
  String lineDefaults;
}

class System
{
  String name;
  1..* system -- 1..* Subsystem parentSubsystem;
  1..* mySystem -- * Block blocks;
  1..* system -- * Annotation annotations;
  1..* system -- * Line lines;
}

class Subsystem
{
  isA Block;
}

class Block
{
  isA BlockDefaults;
  String name;
  String blockType;
  String description;
  String priority;
  String tag;
  String position;
  1..* portBlock -- * Port ports;
}

class Annotation
{
  isA AnnotationDefaults;
  String position;
  String text;
}

class Line
{
  isA LineDefaults;
  String points;
  String id;
  * owningLine -- * OutputPort srcPort;
}

class DirectLine
{
  isA Line;
  * owningDirectLine -- * InputPort destPort1;
}

class InputPort
{
  isA Port;
 // * destPort -- * DirectBranch owningDirectBranch;
}

class Branch
{
  String points;
  * branchs -- 1..* BranchedLine branchedLine;
  * branchs -- 1..* NestedBranch nestedBranch;
}

class BranchedLine
{
  isA Line;
}

class Primitive
{
  isA Block;
  String id;
  * primitive -- * Port refPort;
  1..* primitive -- * Parameter parameters;
}

class Port
{
  String id;
  String number;
}

class Reference
{
  isA Block;
  String sourceType;
  String sourceBlock;
  1..* le_reference -- * Parameter parameters;
}

class Masked
{
  isA Subsystem;
}

class Normal
{
  isA Subsystem;
}

class Parameter
{
  String name;
}

class OutputPort
{
  isA Port;
}

class Type
{
  isA Parameter;
  String value;
}


class Double
{
[...]

PR #1916, opened 2021-12-21
possible solution of the issue 1848 Large model does not generate val…

sidikik55:
## Pull request description

Possible solution of the issue 1848 Large model does not generate valid gv diagram

## Pull request content

For this possible solution of the issue, the file ../umpleonline/scripts/compiler.php has been modified to put a valid width
Line 499 commented out. /// echo $ shrunksvgcode;
Lines 500 to 503, are used to replace 2147483648 or 2147483736.17 by 200


TimLethbridge:
The issue #1848 has gone away so this PR is being closed

PR #1921, opened 2022-01-31
Fix Issue 1917: Mixins do not work properly with association classes

yukangyin97:
additional logic is added in `analyzeAssociationClass` to make sure the umple compiler looks for all association class with the same name and collect their singleAssociationEnd together.


cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
```diff
@@ -1328,7 +1327,7 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
     }
     if ( classToken.is("classDefinition") && "external".equals(aClass.getModifier()) )
       aClass.setModifier(""); // Remove the external modifier if a non-external specification of this class is found.
-
+
```
TimLethbridge: Best to avoid unneeded changes in spacing ... no big deal, but something to watch out for

TimLethbridge (changes requested):
I have tested this and it works! However please add a test to prevent a regression. Take a look at how other tests for AssociationClasses are done and add code like the following example, making sure it does not raise any errors (this would fail in the previous version).

```
class A {
}
class B {
}

associationClass C {
  anAttribute;
}

associationClass C {
* A;
* B;
}
```

cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
yukangyin97: Thank you, I reverted them :)

yukangyin97:
> I have tested this and it works! However please add a test to prevent a regression. Take a look at how other tests for AssociationClasses are done and add code like the following example, making sure it does not raise any errors (this would fail in the previous version).
> 
> ```
> class A {
> }
> class B {
> }
> 
> associationClass C {
>   anAttribute;
> }
> 
> associationClass C {
> * A;
> * B;
> }
> ```

Thank you, test case is added

TimLethbridge (approved):

Issue #319, opened 2015-08-26
Ability to read in models from Papyrus, Ecore, Umlet, RSA etc.


Nava2:
Originally reported on Google Code with ID 420

---

It would promote the use of Umple to be able to process models written in various XML
formats and other formats so that they could be used in the context of Umple. 

One suggestion is that any attempt to process a file, where that file is detected as
something other than Umple (because of the presence in the first line of standard XML
tags etc.) would pass it over to a separate processor that would translate it to Umple,
which would then be parsed as normal.

Ideally at the very least, the formats that Umple produces (e.g. score) could be read.

UML constructs not recognized could be ignored. In an agile manner, we could just start
with finding classes, attributes, generalizations and associations. 

Test cases could include various files found on the Internet or in specs for UML.

Formats to input could include Papyrus XMI, Umlet, Ecore, RSA XMI, etc.

---

Reported by @umple on 2013-09-17 12:54:02

Nava2:
Reported by @umple on 2013-09-17 13:12:41
- **Owner removed**

Nava2:
Reported by @umple on 2013-09-17 13:24:20
- **Status changed**: `Accepted`

Nava2:
See issue 223

---

Reported by @umple on 2013-09-21 13:46:49

Nava2:
Reported by @umple on 2013-09-22 15:45:49
- **Owner changed**: `greg.hysen`

Nava2:
Reported by @umple on 2013-12-13 16:16:11
- **Owner removed**

yukangyin97:
Instructions for taking a Papyrus model and exporting it to Umple:

1. Change directory to "Umple"
2. Copy a Papyrus model, for example, `cp cruise.umple/test/cruise/umple/compiler/PapyrusImport_ClassAssociationManyToMany.uml.txt ~/Downloads/PapyrusImport_ClassAssociationManyToMany.uml `
3. Run a full build against Umple project
4. Copy Umple cli jar file, for example `cp dist/umple-1.31.1.5860.78bb27cc6.jar ~/Downloads`
5. Run Umple cli: `java -jar ~/Downloads/umple-1.31.1.5860.78bb27cc6.jar -i ~/Downloads/PapyrusImport_ClassAssociationManyToMany.uml`

Step 5 will generate an Umple file named `PapyrusImport_ClassAssociationManyToMany.uml.ump`, which is the processed Umple file:

```


class Teacher
{
	* -- * Student;
}

[...]

PR #1931, opened 2022-03-30
#ISSUE_319 Add one case in the import file handler

ruizi:
Add one parsing case scenario to the import file handler to deal with the raise tag in XML files 
Right now, the file import handler class can not parse XML file input correctly when we have `<raise />` tag inside the XML content.
```
else if ("raise".equals(localName)){
      ImportAction raise = new ImportAction("", "", depth, "raise");
      elemStack.peek().addImportStateMachineElement(raise);
      elemStack.push(raise);
}

```


TimLethbridge (changes requested):
You should add a test case that has "raise", which would have failed before and now will not fail.

TimLethbridge:
This PR is for #319

TimLethbridge (approved):

Issue #87, opened 2015-08-26
Proper support for Umple in Papyrus or other open source UML drawing tools


Nava2:
Originally reported on Google Code with ID 185

---

Integration with Papyrus, including capabilities such as drawing diagrams (with UmpleOnline
like capability); conversion between Papyrus UML models and Umple and vice versa.

---

Reported by @umple on 2011-06-25 02:40:35

Nava2:
Reported by @umple on 2011-12-01 18:36:52
- **Labels added**: Type-ProjectUG
- **Labels removed**: Type-Enhancement

Nava2:
Umple will get more use if people using UML modelling tools in Eclipse can generate
Umple, and vice versa - can write Umple and see the result in their diagramming tool.

---

Reported by @umple on 2012-06-15 14:22:57
- **Labels added**: Diffic-Hard

Nava2:
Reported by @umple on 2012-09-13 19:32:27
- **Labels added**: Priority-High
- **Labels removed**: Priority-Medium

Nava2:
See also issue 8 and issue 8

---

Reported by @umple on 2012-09-27 20:28:10
- **Owner removed**

Nava2:
Umple needs updating comprehensively to ensure it generates proper papyrus XML from
all Umple features (all associations, attributes, interfaces, state machines, etc.)

---

Reported by @umple on 2012-09-28 20:06:27

Nava2:
Reported by @umple on 2012-12-11 13:04:02
- **Owner changed**: `r.staughton`
- **Status changed**: `Started`

Nava2:
Patch of progress related to papyrus generation. The generator supports guard conditions,
start states, final states, deep history states, history states, events, transition
actions, entry/exit actions, do-block activities. 

note: A separate papyrus project is generated for every state machine the .ump file.

Naming convention: className_StateMachineName.

TODO- Support for nested state machines, Update JUnit tests.

---

Reported by `jzamb022` on 2013-04-09 22:54:47
- _Attachment: [patch-Zambrano-#185-PapyrusGeneration.diff](https://storage.googleapis.com/google-code-attachments/umple/issue-185/comment-7/patch-Zambrano-#185-PapyrusGeneration.diff)_

Nava2:
Some progress has been made. Ideally Umple would be incorporated as a plugin to Papyrus
and model-exchange would be two-way. Making it work with basic Eclipse Ecore models
would also be great.

---

[...]

PR #1941, opened 2022-04-15
Papyrus to Umple & Umple to Papyrus functionalities through Umple Eclipse plugin

yukangyin97:
In this pr, a "Generate Papyrus project" button is introduced in Umple Eclipse plugin, it can be seen in the "Others" section.

791650055088_ pic

Once that button is clicked, a Papyrus project has the same name with the .ump file will be generated.

image

and the generated Papyrus project can be imported and its model can be visible as well

image


yukangyin97:
This pr is for https://github.com/umple/umple/issues/87

Issue #569, opened 2015-08-26
Support for concurrent state machines in state table generation


Nava2:
Originally reported on Google Code with ID 671

---

THIS IS RELATED TO #670

If you run the following

class Student
{

  status {

```
CrossingGuard {
  power -> Off;

  Walk {
    push -> DontWalk;
  }
  DontWalk { 
    push -> DontWalk; 
  }
  ||
    Green { 
      carAppraoch -> Red;
    }
  Red { 
     carApproach -> Green;
  }
}
Off { power -> CrossingGuard; }
```

  }

}

The state table doesn't support two things that it should.  First, much like nested
state machines, it should show the events related to the parent state (which should
be solved by #670).  Second, the display of the actual tables should keep the states
together to show the concurrency

---

Reported by @aforward on 2015-01-17 23:17:50

Nava2:
I am trying to think of a way to deal with this. In the example there are two sets of
concurrent states: Walk/DontWalk and Green/Red. The statemachine can independently
switch between the concurrent states, leaving us with 4 possible positions in the state
machine (Walk-Green, Walk-Red, DontWalk-Green, DontWalk-Red). My first thought re:
"keep the states together to show the concurrency" is to have these 4 possible positions
as headers. So instead of 4 independent states, there are 4 combined states.

Here's how the example would work:
https://docs.google.com/spreadsheets/d/1ouxNcw7KutWN6E0_54euOnfF3PJyvmirMpVXb0nqiAg/edit?usp=sharing

---

Reported by @CraigBryan on 2015-03-01 16:29:53

Nava2:
Reported by @CraigBryan on 2015-03-01 16:30:11
- **Labels added**: stateMachines

TimLethbridge:
The only problem with the proposes solution is that there could be an explosion of the number of possible state combinations. I think the generator would have to detect that and only give a subset if there are going to be more than, for example 50 combinations for a given set of states

TimLethbridge:
A remaining edge case is transition between nested states, if the parent states differ.

Delta-Hat:
`class Student
{
  status{
    ParentState1
    {
      crossDown -> AwayChildState2;
      ChildState1
      {
        action -> ChildState2;
      }
      ChildState2
      {
        action -> ChildState1;
        crossUp -> ParentState2;
      }
      ||
      OtherChildState1
      {
        otherAction -> OtherChildState2;
      }
      OtherChildState2
      {
[...]

PR #1949, opened 2022-04-22
Progress on #569 Added concurrent state-table support

Delta-Hat:
Currently, state tables do not support concurrent state machines.
This PR adds limited state-state and state-event tables to the state table generation.

It does this by creating a series of CompositeStateWrapper (CSW) structures which store multiple concurrent states at once.
The CSW contains a function that, given a transition and another CSW, verifies whether or not the given transition would lead to the next CSW.

Using this info, the generator fills out the state-tables.

Currently does not support transitions between concurrent states.


TimLethbridge:
This PR mostly fixed #569 but there remains an edge case regarding teransitions betwen nested states if the parent states differ.

PR #1952, opened 2022-05-03
Docker updates

TimLethbridge:
This pull request updates the Docker setup in order to enable the UmpleOnline Docker image to work properly

The UmpleOnline image previously just encompassed UmpleOnline and ran Php to create a web server.

With the new UmpleCodeExecution feature in a recent PR, there are then two other Docker images to deal with.

So in the new UmpleOnline image, there will be

Level 1: main UmpleOnline image (creates self-contained web server). Found in UmpleOnline/Dockerfile (with its base image umpelOnline/Dockerfile-base that changes only slowly). Best created using the bdock script. Once created, can be run by udock script in dev-tools, and published on DockerHub. Running container starts a web server, default port 8000.

Level 2: Within level 1, there will be the UmpleCodeExecution container with image created from its Dockerfile. This starts an internal web server defaulting to port 4400 to receive requests from level 1 to execute Java.

Level 3: Level 2 image will create and start UmpleCodeExecution/javarunner images from that the javarunner Dockerfile. These simply execute and return the result of executing the Java code, within tight constraints.

This PR updates various files, including

  UmpleCodeExecution/dockerfile
    ... to comment it

  UmpleCodeExecution/app.js
    ... to add a small comment useful for future debugging

  UmpleCodeExecution/setup.sh
    ... to tell images to restart if the docker daemon is updated (so service doesn't halt)

  bdock
    ... to add better comments, and instructions, and to ensure UmpleCodeExecution is included in the image
  
  umpleonline/Dockerfile
    ... to better document it and exclude items no longer needed
    ... also, at least temporarily, changing userid that will run
    
  umpleonline/Dockerfile-base
    ... to use latest alpine and also to include curl and bash
    ... note that this base has *already* been pushed to Dockerhub
    
  umpleonline/counter.php
    ... to better display the sha of the git version the Docker image (and UmpleOnline in general) on the main web page
    
  umpleonline/docker_config/supervisord.conf
    ... to ensure that UmpleCodeExecution/setup.sh is run; not just nginx and php as before
  
  umpleonline/scripts/compiler.php
    .. to make the The docker service to execute code is not working message more informative
    
  umpleonline/scripts/log.
    .. to make the log page (DevOps monitor) be able to display Docker info.

This PR should have no affect on UmpleOnline as installed at https://try.umple.org, or developers who install UmpleOnline directly on their machine, since such installations do not use the Level 1 image as described above.

To test this PR
  - go to the main directory of your Umple on a Mac or Linux machine and run dev-tools/bdock
  - Set up directory /umpleonline-tmp81701 for temporary ump files
  - docker run --rm -ti --name umpleonline_local -p 8000:8000 -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/umpleonline-tmp81701:/var/www/ump docker.io/umple/umpleonline:recentbuild
  - Open a web browser at http://localhost:8000/umple.php#genArea


TimLethbridge:
This is still a work in progress as it requires testing, and because there have been problems with executing: The Level 2 Docker image has been refusing at port 4409 (which is the port we decided to use).

Issue #1973, opened 2022-09-15
Support of Umple by Debian and RPM package managers


lcai006:
Add support of Umple for distribution as Deb and RPM packages, so that umple can be install by using apt/rpm command-line tools.
Ideally Umple would run just using the 'umple' command. A manual page would be added for linux users.

The result would be similar to homebrew #756

lcai006:
The gradle task which creates deb and rpm packages is using **gradle-ospackage-plugin**
(https://github.com/nebula-plugins/gradle-ospackage-plugin)
The options of manpage are generated from umple --help message.

Package includes umple.jar, a bash script to run umple, copyright and README of umple, a manpage zip file of umple.
Related files are in dev-tools/files directory

lcai006:
Run below after build the project in umple/
```
./gradlew genPackages
```
This task creates packages in umple/dist/gradle/distributions directory

To install the package locally:
```
sudo apt install pathToUmple/dist/gradle/distributions/umple-xxx.deb
sudo yum install pathToUmple/dist/gradle/distributions/umple-xxx.rpm
```

Result:
Open manpage:
```
man umple
```

Run umple:
```
umple filename
umple -help
...
```

lcai006:
Official Doc **Package signing in Debian**:
https://www.debian.org/doc/manuals/securing-debian-manual/deb-pack-sign.en.html
Wiki for create deb repository:
https://wiki.debian.org/DebianRepository/UseThirdParty?action=show&redirect=RepositoryInstructions#Complete_example

Wiki for **Signing and Creating a Repository for RPM Packages**:
https://wiki.cdot.senecacollege.ca/wiki/Signing_and_Creating_a_Repository_for_RPM_Packages
Tutorial for createing yum repository:
https://www.thegeekdiary.com/centos-rhel-how-to-create-and-host-yum-repository-over-httpd/

Personal Package Archive (PPA) for Ubuntu users:
https://help.launchpad.net/Packaging/PPA

PR #1974, opened 2022-09-21
Issue1973 Add gradle tasks to generate Deb and RPM packages

lcai006:
Issue #1973 
Add gradle tasks to generate Deb and RPM packages.
Package includes umple.jar, a bash script to run umple, copyright and README of umple, a manpage zip file of umple.
Related files are in dev-tools/files directory

Thes gradle tasks are using gradle-ospackage-plugin.
Run below after build the project in umple/
```
./gradlew genPackages
```
 This task creates packages in umple/dist/gradle/distributions directory


TimLethbridge (changes requested):
Good start. 

Instead of using dev-tools as the directory for the three files, please use the build directory ... the dev-tools directory is for scripts that are used by developers of Umple itself.

And instead of 'files' as the subdirectory, please use a more meaningful name such as rpmfiles for the files used to build rpm.

Also there is already a license for Umple in LICENSE.md. Is there a way this could just be copied directly, rather than being duplicated>? This is am MIT license. I have been putting the names of all the authors there, so it may change over time. Also I have updated the date periodically.

Please fix these (or at least reply), then I will test.

Also in the issue, it would be good if you could tell me how I might set the resulting package up so I can test it using an apt or a yum command

lcai006:
> 

> Good start.
> 
> Instead of using dev-tools as the directory for the three files, please use the build directory ... the dev-tools directory is for scripts that are used by developers of Umple itself.
> 
> And instead of 'files' as the subdirectory, please use a more meaningful name such as rpmfiles for the files used to build rpm.
> 
> Also there is already a license for Umple in LICENSE.md. Is there a way this could just be copied directly, rather than being duplicated>? This is am MIT license. I have been putting the names of all the authors there, so it may change over time. Also I have updated the date periodically.
> 
> Please fix these (or at least reply), then I will test.
> 
> Also in the issue, it would be good if you could tell me how I might set the resulting package up so I can test it using an apt or a yum command

Above issues fixed.

To install the packages:
[#1973 ](https://github.com/umple/umple/issues/1973#issuecomment-1254200273)

TimLethbridge (approved):

Issue #1978, opened 2022-10-24
The --path option does not work when generating Cpp files from umple


lcai006:
When I run umple:
`
java -jar umple.jar .\test.ump -g Cpp --path ./Cpp
`
It generates files for Cpp in current path, instead of creating a new directory.
This happens when the language is Cpp. It is expected to geneated files in the target directory.

If the language is exact "RTCpp", the path option works, but only *_Main.cpp, *_Model.h, CMakeLists.txt in the target directory.

Also, for now, the option string needs to be exact "Cpp" for generating c++ files. It is better to make it case insensitive.

TimLethbridge:
The last sentence 'Exact Cpp' is a generic issue for all generators. I do agree that making this case sensitive would be a good idea (both on the command line and also when the generate clause is used in the code); however, it should be done for all generators.

lcai006:
> The last sentence 'Exact Cpp' is a generic issue for all generators. I do agree that making this case sensitive would be a good idea (both on the command line and also when the generate clause is used in the code); however, it should be done for all generators.

When I generate files from umple, only "Cpp" or "SimpleCpp" is case sensitive.
In umple/cruise.umple/src/Umple_Code.ump, it uses equalsIgnoreCase() for languages, but it uses equals() to change "Cpp" to "RTCpp".

Umple_Code.ump:
line233:    if (language.equals("Cpp")) realLanguage="RTCpp";
line234:    else if (language.equals("SimpleCpp")) realLanguage="RTCpp";
But this does not affect --path option.

TimLethbridge:
@lcai006 OK. So you could easily fix the case sensitivity sub-issue yourself ... make the change and do a full build to ensure tests pass ... then submit a pull request with just that little change.

TimLethbridge:
This seems a duplicate of #507 ... when fixing one, please deal with the other

PR #1979, opened 2022-10-25
Make cpp and simplecpp as a case insensitive option for generating files from umple

lcai006:
When generating c++ files from umple, the option of language is case-insensitive. The option can be cpp, simplecpp and rtcpp.


lcai006:
#1978

Issue #1963, opened 2022-07-27
Allow collaborative text editing like in Google Docs


TimLethbridge:
This would be challenging, but it would be super-useful if an UmpleOnline mode could be created that would allow multiple people to edit the model simultaneously like in Google Docs.

Best if it could be done with multiple people but at least 2,

Currently text editing uses CodeMirror. It would be necessary to transmit every keystroke in both directions without disrupting what the other is typing.

Since we don't want to require 'accounts' an invitee would have to be given a special code, stored in the session directory, to allow them to collaborate with the original author.

TimLethbridge:
Some constraints and ideas on this:

Umple is login-free and we want to keep it that way.  Access to a repo is by a code (URL) that gives access to the data in a relevant directory in umpleonline/ump ... so in this feature, the person who initiates sharing would be granting other users access to that same directory.

This would not work on 'tmp' session directories, so it would imply making a 'bookmarkable URL' if that has not already been done.

To keep it simple, I suggest a button be added that says 'collaborate ' . That would make available a link that could be sent to the others (the user would copy the URL and send it by whatever means they prefer ... text, email. etc.) 

Collaboration should be symmetric. Once it is initiated, all Umpleonline sessions operating on the same directory (including the initiator) would be the same.

Although UmpleOnline is login-less, during the process of setting up a collaboration, there should be a field where the initiator can enter an ID (this should be non-modal). That would make their presence visible to others and would help control the data flow. This ID could be a field in the app that would be always visible and editable even, and likely transmitted with each request. One option would be that a random short id could be set up initially.

The 'collaborating' status would have to be recorded in the shared directory, so that any new access to that directory would require an id before allowing editing.

Some kind of peer to peer arrangement would be needed: Each collaborating session would need to watch for edits from the others. I suggest these be sents as 'diffs' of each ump file (there can be multiple active ump files) that could be applied to the text on the screen if the same ump file is being edited by the other collaborator. Ideally this would be applied as unobtrusively as possible, so if both are editing the text of both displays. Maybe recent text by others appears in a different colour, or a message appears for a few seconds.

If a diagram is edited, this would also update the text, so the text should be primary,.

Different collaborators may be showing different diagram types.

If different collaborators are editing different files, then it is still possible that a diagram update may be needed if there is a use statement between the files.

This needs to be done without adding excess dependencies to Umple.

TimLethbridge:
[...]

PR #2019, opened 2023-01-30
issue #1963 - Upgrade CodeMirror version from 2.13 to 5.65

deepankkartikey:
This pull request updates CodeMirror from a version dating from 2015 to the latest version 5.  This is intended to have no noticable functionality differencve, except to make it up to date. We will immediately move to version 6 in a later PR.


TimLethbridge (changes requested):
There is a fundamental critical problem! When run, UmpleOnline does not load this new version!

There are two problems:

1. The previous version was including foldcode.js and searchcursor.js
These moved from lib to addon subdirectories

2. The _load.js is loading allumple-min.js which still has the old version! There is a script to make the new version (unix script pumple will run this,  or you can run the ant command
ant -DshouldPackageUmpleOnline=true -Dmyenv=local -f build.umple.xml packageUmpleonline 
But currently that build script fails because of issue 1

If I do compress the scripts I get a total failure.

You need to do three things:
a) Fix the compressAllScripts target to ensure it includes the right files
b) Change the commented-out part of _load.js to also be consistent
c) Then start testing/debugging from scratch!

TimLethbridge (approved):

Issue #1994, opened 2022-11-16
Continued improvement of the Python generator


TimLethbridge:
Issue #1384 involved creating a generator for Python. However as of the end of 2022, when it was/will be released in beta mode, certain aspects remain incomplete. This issue is for continued work to bring Python generation closer to the level of the Java generator.

In particular the following needs completion:

- Concurrency: State machine do activities in particular (including the active keyword)
- Timing in state machines (e.g. the 'after' keyword)
- Execution of Python within UmpleOnline using the same mechanism as is used for executing Java (i.e. within the Docker image)
- Specific Java classes need to be converted and stored in the template files.
- Adding comments back into the generated files
- See https://cruise.umple.org/umple/Python.html for all other unsupported features

Tlapr231:
For the specific Java classes, a possible solution would be to create a new template file (copy the java ones) and manually translate these classes. This translation can either be a full translation or a half translation where the TXL will be able to properly interpret the rest. 

Note that research may have to be made to find a proper Python replacements for the Java specific classes. Also, the after keyword is linked with this issue as the after keyword generates the specific TimerTask java class.

SamuelLabonte:
Here is some context as to why the features ‘active’ and ‘after’ were left as out of scope for the initial beta. When these keywords are included in Umple code, they will generate Java files that contain nested classes (a class within a class). The nested classes (ex: TimerTask when you use ‘after’) are full of java specific code. These classes are also independent of the Umple code written by the user. Since these classes are always the same, it is simply not worthwhile to update the TXL to parse and translate the code in those classes. Since the focus of our approach was TXL, we decided to leave these classes (and the Umple keywords that generates them) as out of scope given the time constraints we had. We suggest that these classes be translated by hand and inserted in the java intermediate code before TXL translation. These can be wrapped around tags so the TXL can ignore them and not need to parse or translate them like regular code. We followed a similar approach for user functions (functions with user code which can be anything) and it worked quite well. The tags are then removed after the TXL translation, so they do not stay in the final output.


TXL has a feature that allows you to specify what you want it to consider comments. TXL then completely removes these and ignores what’s inside. Ideally, these comments would still be in the output as they help orient users in the generated class and specify from what relations certain methods originate. To translate these comments, the comments will have to parsed and translated along with the regular code. This will take new TXL rules to translate them and changes to the grammar to accept comments wherever they might appear (at a quick glance, making comments [class_body_element] and [statement] might be enough). 


[...]

PR #2034, opened 2023-04-27
Java to Python: nested classes, timing, concurrency improvements.

FredWei-Git:
## Pull request description

[Issue 1994](https://github.com/umple/umple/issues/1994)

Improvements to nested classes, timing, and concurrency of Java to Python. Nested classes, after keyword, active block, do activity now supported.

## Pull request content

- TXL changes needed for nested classes, after keyword, active block, do activities
- Changes made to Python generator to allow for Python blocks and code within do activities to not be translated
- Python state machine implementation tests included


build/_wlocal.xml
FredWei-Git, line 1: having it set to rake.cmd does not pass the build but having rake.bat passes the test. it was originally rake.cmd.

TimLethbridge (changes requested):
Here are some notes with needed fixes. I have focused for now on testing for the active keyword. Once I see this working, I will test equivalent do activity state machine code that should give the same result. I wll also test the other stuff you have done with main programs.

### Note A:

I tried compiling the attached Umple code PtActive.ump for the active keyword, with some embedded python. This is similar to the code testing active for Java in the user manual page https://cruise.umple.org/umple/ActiveObjects.html .

Note that it won't allow me to attach here filenames ending in ump, so I have added .txt.

[PtActive.ump.txt](https://github.com/umple/umple/files/11344203/PtActive.ump.txt)

There are several problems in the result.

- It seems to join together several lines of Python without indentation, resulting in bad Python. This may be the result of injection of a line end character that is incorrect. Perhaps it works in Windows, I don't know.
- It is for some reason adding a strange extra time.sleep call that I had not specified.

The bad part of the generated code is around line 75
```
        try :
            print("Object " + self.getName() + "is now active")
            for i in range(1, 9):
                print(" " + self._name + " was here") print(" " + self._name + " was also here")
                time.sleep(1) print(" " + self._name + " was here too")
                time.sleep(0.001)
```

### Note B:
I notice you have put comments in build.exampletests.xml that say

`April 2023, no longer excluded
`

However the set of files where the exclude has been deleted equals the set of files where the exclude has been added back (Dishwasher, Ball_Game and Runway)

### Note C:
I see implementation tests for inner classes but no tests for the other improvements.

It is really only by testing that I can be confident things work properly.

### Note D

Once you have this working, you should also update this user manual page:
https://cruise.umple.org/umple/Python.html
The source for that is at build/reference/14030Python.txt

### Note E
UmpleOnline still says, for the active keyword
#An Unsupported feature for Python generation was used. File could not be generated.
That would need fixing, once the active keyword is working

### Note F
The following may be a problem left over from last year, but if I put the keyword 'Python' before the starting curly bracket code block of the main method, no code is output.

TimLethbridge (approved):

PR #2065, opened 2023-12-12
Solving requirements crash #2051 and convering print to proper warning #2052

### LINKED ISSUE #2051: Requirements parsing crashes in certain cases when a requirement indsicated in an implementsReq statement does not have a matching req statement
This is a sub-issue of #1865 which is the main issue for the requirements feature.

The following code causes a crash during analysis following parsing. It seems to be related to when there is an implementsReq that does not match a req
```
req RT3 {
}

class X {
  implementsReq RT4;
  a;
}
```

It also occurs if there are no req statements at all (this should be another test). Or if a matching req statement is in a mixset that is not activated with a use statement.,

### LINKED ISSUE #2052: Requirements problems should raise warnings rather than printing to standard output
This is a sub-issue of #1865 which is the main issue for the requirements feature.

Currently there are System.out.println statements at lines 207 and 234 of class/UmpleInternalParser_CodeClass.ump

These are statements pointing out problems with requirements. All requirements issues should instead raise warnings.

New relebvant warning messages should be added to en.error (with relevant parameters). and then the parsing code should call addErrorMessage to raise the warning ... the warning should occur at the appropriate line number. See how other warning messages are raised in the same file.

And while at it ... please comment the code of the relvant methods so they are easier to understand.

See also related requirements issues as you work on this.

TimLethbridge:
This fixes #2051 where there was a crash case in the requirements feature: It was trying to analyse a requirements implementation as an attribute.

It also turns the print statement into a proper Umple warning message (#2052)

### DIFF
--- a/cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
@@ -223,37 +223,27 @@ class UmpleInternalParser
 
   private void analyzeImplementReq(Token t){
     int counter=0;
+    String badReq="";
 
     if (t.hasSubTokens()){
       for (Token sub : t.getSubTokens()){
         if (!sub.getValue().equals("STATIC") && model.getAllRequirements().containsKey(sub.getValue())){
             lastRequirements.add(model.getAllRequirements().get(sub.getValue()));
         }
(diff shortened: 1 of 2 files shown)


TimLethbridge:
Fixes #2051   Fixes #2052

PR #2067, opened 2023-12-17
Added suboption to increase separation

### LINKED ISSUE #2015: Add ability to fine tune layout of graphviz class diagrams
Graphviz class diagrams are currently set with default node separateion. Sometimes when there are many associations this makes them look messy.

Graphviz has nodesep and ranksep settings that can be applied right after the digraph directive. The default setting is 0.5 but I have experimented, and sometimes being able to increase this to about 3 can be good. 

I suggest a 'suboption' be added to the generator (and related generators) that would inject a nodesep and ranksep value. 

In the UI, these settings could perhaps be adjusted by a separate menu item in Umpleonline (eg. Graphivz widely separated). Or a special command could be applied in the line number field e.g. sep=1.0 or sep=2.0 that would trigger regeneration of relevant diagrams (and persist the value)

syedahmed1674:
# Description
I added a suboptions to increase separation for ranksep and nodesep. 
- Example use : java -jar umple.jar -g GvStateDiagram gd.ump -s gvseparator=2

The ability for suboptions to have numeric arguments is also available to be used in other options

### DIFF
--- a/cruise.umple/src/Generator.ump
+++ b/cruise.umple/src/Generator.ump
@@ -19,11 +19,10 @@ interface CodeGenerator
   boolean setOutput(String aString);
   boolean setSuboption(String subopt);
   boolean hasSuboption(String subopt);
-  double getSuboptionValue();
+  double getSuboptionValue(String subopt, Double defaultVal);
   void generate();
   void prepare();
   void postpare();
-
 }
 
 // Default implementation for generators with suboptions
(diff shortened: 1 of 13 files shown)


AxelTWC:
Reference: Issue: #2015

TimLethbridge:
I made changes to fix the following: 

- 'seperator' should be 'separator' everywhere. I changed it
- getSuboptionValue should be generic (so it can work for any suboption). I changed it so it took a string argument (so not hardcoded and a default value so it can be used in other contexts).
- indentation was not consistent (was pretty bad in fact in many lines). I fixed
- You forgot to add the new tests to the relevant java classes in the dv test directory so they were not being run
- The class diagram gv file did not even have the new separator lines

PR #2082, opened 2024-01-19
requirements with same identifier should be combined

vybhavigururaj:
This pull request solves the issue 2053 where requirements with the same identifier should be able to concatenate the content provided that the language is the same. If the language is different a warning message is displayed.


TimLethbridge:
This was solved instead with PR #2084 so this one is being closed

Issue #2081, opened 2024-01-15
Improve UmpleOnline editing of classes and state machines


TimLethbridge:
In UmpleOnline's G mode (graphviz class diagram) one can right-click on a class name, bringing up a menu that says 'add attribute', rename class, etc.

Suggested improvements

1) It would be good if double-clicking would have the same effect, as right-clicking might not be intuitive.
2) Enable clicking on an attribute, with it bringing up a menu to delete the attribute, change its type and rename it
3) If I right-click on a transition, it would be good if it would enable me to edit it.
4) It would be good if users somehow would know that they can double-click or right click! ... e.g. if the hover message would say 'right click to edit' ... note that we shold also ensure that PR #1908 for issue #1867 is either redone or finished in conjunction with this

TimLethbridge:
I am not sure why I didn't list associations as well as transitions. But yes the idea is that if the user clicks on one it would seem that one should be able to. delete it, alter the event name and the destination state. For an association one should be able to alter the multiplicity at either end and the role name at either end, and delete it (but maybe not change the classes ... for that I think deleting and adding new would work better perhaps

TimLethbridge:
Work on this is underway

PR #2106, opened 2024-03-01
Issue #2081 Part 3: Enhanced transition interactions in UmpleOnline Graphviz diagrams

ah110:
This PR enhances user interactions with transitions in UmpleOnline's Graphviz diagrams. Users can now directly modify transition properties such as event names and destination states through a clickable context menu.


TimLethbridge (changes requested):
Excellent progress.


I note the following bugs

If I do the following
    i. Load a state diagram
    
    ii. double-click on a transition
        (the menu appears correctly)
        ISSUE a: For the other options, right-clicking (control-clicking on a mac) has the same effect, so this should be consistent.

    iii. Select 'Change Event Name'
        (a popup apears correctly most of the time)

       ISSUE b: The name of the original event should appear in the field, making it easier to edit (I notice that even naming states puts this value textually prior to the field, but it would be better if it was in the field itself, hence editable) .. this is OK for editing attributes in class diagrams. Please check this for everything that is now editable.

       ISSUE c: If I click somewhere else (outside the box) the box does not go away ... and then it never goes away! The same problem occurs for rename attribute in a class diagram ... please check this for all menus

       ISSUE d: For Canal Lock example double-clicking on boatRequestsToEnterAndGoUp and then select Change Event Name ... the popup does not appear ... same for some other transitions. Please ckeck this for many examples.

TimLethbridge (approved):

TimLethbridge (changes requested):
I have tested this thoroughly. It is much better. There are still some glitches:

a) It is great now that if I do 'Change Event Name' that the previous event name is now filled in.

   TODO 1: I noticed that everything prior to the -> is treated as the event name and everything after is treated as the destination state. However in Umple Syntax there also can be a guard (marked by square brackets), and optional block of code, marked by / {...}. We should really separate these 4 elements and have Change Event Name, Change or Add Guard (that would allow changing the contents of the item in square brackets, and if blank is entered the result would remove the square brackets, Change or Add Transition Action (that would change the contents of the curly brackets, and if none would remove the / {...), and Change Destination State. Note that some events have arguments (so the event can have brackets and arguments) ... so I suggest Change Event Name be Change Event
   
    ... although I note that if there is also a transition event it sometimes allows editing of it and soemtimes not.
   TODO 3: If I select 'Rename State' it says "New name for 'X'", but it should put the X in the editable field.
   TODO 4: Same for Rename Class in a class diagram
   
b) In a class diagram, I can double-click on a class or attribute. Great
  TOTO 5: If I control-click (right click) on an attribute) it doesn't do the same thing as double-clicking ... it should.

TimLethbridge:
Merged into #2117, so closing

PR #2107, opened 2024-03-01
Draft-Wenxin - Replacedby2109 To Close

Wxin-Duan:
Now it allows the use of suboption -s genJson to generate a toJson() java function, by using the command "test1.ump -g Java -s genJson"
**If were to modify the code and run test, make sure compile the code everytime before testing using "test1.ump -c-"; otherwise the changes would not be reflected on the .java code.


TimLethbridge:
See PR #2109 instead

Issue #26, opened 2015-08-26
Complete generation of abstract in php


Nava2:
Originally reported on Google Code with ID 76

---

class X
{
  abstract;
}

Or,

abstract class X
{

}

The first syntax is suggested even though it is against the _convention_ in other languages.
We already can declare a class as 'singleton'.

Doing this would require also changing UmpleOnline to allow marking a class as abstract.
There are metamodel and code generation implications.

---

Reported by @umple on 2011-06-24 13:41:07

Nava2:
Reported by @umple on 2011-06-24 15:46:05
- **Labels added**: contribSought

Nava2:
I'm looking into this currently.  My plan is to go for the first method for the Umple
syntax.  That way it follows how "singleton" and "immutable" are done.  Consistency
is important I think here.

---

Reported by @EmperorJordan on 2012-03-16 01:50:53

Nava2:
The abstract keyword is in place and working correctly.  Simply doing the following:

class Student
{
     abstract;
}

Will make the class "student" an abstract class such that if you were to generate Java
code the class would be "public abstract Student".

The only thing left right now is two fold, as I see it:

1)  This only works for Java code generation currently, other languages will need the
implementation which should be a jet-template change (the jumpjet)

2)  Checks to ensure instantiation of an abstract class are impossible.  It was mentioned
by Andrew that this would be solved if the native language compilers (javac for Java,
or gcc for C++) would bubble up the errors.

If time permits I will have this solved before the end of term (next week), if not
I'll simply solve it afterwards as I'm more than comfortable continuing onwards.

---

Reported by @EmperorJordan on 2012-04-01 03:28:28

Nava2:
Reported by @umple on 2012-06-14 21:47:38
- **Status changed**: `Started`

Nava2:
Reported by @umple on 2012-06-15 15:12:19
- **Status changed**: `MostlyDone`
- **Labels added**: Diffic-Easy

Nava2:
Reported by @umple on 2012-09-27 19:41:51
- **Owner changed**: @EmperorJordan

Nava2:
Reported by @umple on 2013-06-19 12:51:17
- **Owner removed**
- **Labels added**: Priority-VHigh
- **Labels removed**: Priority-High

Nava2:
Reported by @umple on 2013-09-12 15:08:50
- **Labels added**: ucosp

Nava2:
[...]

Issue #25, opened 2015-08-26
Put in a disclaimer in Ruby indicating certain aspects not fully implemented


Nava2:
Originally reported on Google Code with ID 72
Owned by @aforward

---

A disclaimer needs putting into the Ruby code indicating certain areas where there are
weaknesses. The Java and PHP generators are more developed.

---

Reported by @umple on 2011-06-24 13:35:53

Nava2:
Updated the Ruby disclaimer to:

# PLEASE DO NOT EDIT THIS CODE

# This code was generated using the UMPLE @UMPLE_VERSION@ modeling language!

# NOTE: Ruby generator is experimental and is missing some features available in

# in other Umple generated languages like Java or PHP

Once the latest build passes - it will be available on UmpleOnline

---

Reported by @aforward on 2011-06-30 19:57:50
- **Status changed**: `Fixed`

PR #2108, opened 2024-03-01
Dependabot alerts #25 #26 - semver vulnerable to Regular Expression Denial of Service

stevewjb:
This PR should fix Dependabot alerts 25 and 26, which are related to semver vulnerable to Regular Expression Denial of Service. However, since the version upgrade of some packages is a breaking change, there might be some potential side effects after applying these changes.


stevewjb:
I am going to do more tests on the system to make sure it is safe to upgrade.

Issue #2101, opened 2024-02-22
Enums Not Generated Correctly in Ecore


erica-d:
## Specify a Brief Description

When converting an Umple file with an enum to an ecore file, the enum gets generated as an EData Type rather than as an EEnum.

## Give the minimum Steps to Reproduce

1. On Umple online, create a class diagram with an enum in one of the classes
    Possible Example:
    
```
class RootClass {
  1 <@>- * Class1 classes1;
}

class Class1 {
  unique Integer number;
  enum ExampleEnum { Option1, Option2 }
  ;
  ExampleEnum exampleEnum;
  
}
```
2. Press "Generate" after selecting "Ecore" in the "Generate" tab

## Describe the expected result that didn't happen

Expected a section in the ecore file with ExampleEnum as an EEnum

`<eClassifiers xsi:type="ecore:EEnum" name="ExampleEnum">
    <eLiterals name="Option1"/>
    <eLiterals name="Option2" value="1"/>
  </eClassifiers>`

## Actual Result
Instead, an EData Type is created:

`<eClassifiers xsi:type="ecore:EDataType" name="ExampleEnum" instanceClassName="ExampleEnum" />`

## Labelling and Assigning

'ecore', 'Enum' 'EEnum'

TimLethbridge:
The relevant code for the ecore generator can be found in directory

cruise.umple/src/generators/ecoregenerator

The class Generator_CodeEcore_Model defines class EcoreEnum at line 468 and is set up to output the correct code for Ecore enums.

The bug appears to be where EcoreEnum is instantiated ... currently only in line 222. In Umple enums were originally just degenerate state machines from a grammatical perspective without the keyword 'enum' and this notation still works. The Ecore generation code, which was written long ago, assumed this notation. 

```
class X {
 Status { FullTime, PartTime }
 Status s; 
}
```

If this notation is used the enum Status is output correctly .. but ALSO  there is an incorrect output of an ecore:EDataType Status.

Part of the problem is that the Umple Ecore generator wasn't updated to use enums that use the enum keyword. The generator would have to be modified to identify these as in the following:

```
class X {
 enum Status { FullTime, PartTime }
 Status s; 
}
```

or in the following

```
enum Status { FullTime, PartTime }

class X {
 Status s; 
}
```
The Ecore for the last simple example should (I think) be:

Correct code (probably)

```
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #2116, opened 2024-03-20
Issue#2101 Enums Not Generated Correctly in Ecore

stevewjb:
This PR aim to fix issue #2101, currently it did not pass all the test. I will continue working on it and make new commit in this PR.


stevewjb:
For now, the generator should be able to identify the "enum" keyword, however, it only works for one type of syntax. The new test case for this syntax was added. I need to further polish the code to make it work for other scenarios.

stevewjb:
All the issues are fixed, the PR is ready to be merged.

PR #2117, opened 2024-03-20
Menus for G mode editing (graphviz class diagram) and S mode state diagrams

ah110:
Pull Request Description:
This PR introduces  association interaction within UmpleOnline's class diagrams, enabling users to efficiently manage associations through a context menu. With this feature, users can now click on an association to access options for modifying multiplicity, changing or adding role names, and deleting the association directly from the diagram. 

Key Features:

Clickable associations in class diagrams to trigger context menus and highlight the code.
Options to modify multiplicity and role names, or delete associations.
Simplifies association management within UmpleOnline diagrams.

this menu is well tested , it working with all type of association(sorted, association class, composition, Independently Defined Associations...)

for association it wont allowed user to modify the end multiplicity since its not allowed in umple.
for  Symmetric association the start and end will be same.

Fixes for state diagrams are also incorporated.

UX improvements include
  Escape key to dismiss edit panels
  Previous names displayed, ready to edit
  Syntax errors displayed in red right next to edit panels


ah110:
merge transition menu into this branch, since its easier as the transition menu branch is out of date ,

ah110:
for association when click it should highlight the code, and when right click or double click it will display a menu for user to modify role name and multiplicity, for transition, when click it should highlight the code, currently not working with negative guard, and for it should display a transition menu for user to modify the guard and action and name.

Issue #1067, opened 2017-07-13
Allow option objects of any Umple model to be serialized/deserialized to/from Json


TimLethbridge:
It would be good if any object had a `String toJson()` method that would generate (serialize) a Json representation, and a `fromJson(String)` static method that would instantiate (deserialize) the object.

Since this would not be needed in every single model, my suggestion is to only generate this if -s genJson is specified as a suboption on a generator.

A complication is that Umple has dates and times as primitives. We would need to use ISO8601 format. See [https://www.hanselman.com/blog/OnTheNightmareThatIsJSONDatesPlusJSONNETAndASPNETWebAPI.aspx](https://www.hanselman.com/blog/OnTheNightmareThatIsJSONDatesPlusJSONNETAndASPNETWebAPI.aspx)
and
[http://blog.plataformatec.com.br/2014/11/how-to-serialize-date-and-datetime-without-losing-information/](http://blog.plataformatec.com.br/2014/11/how-to-serialize-date-and-datetime-without-losing-information/)

Note that Umple already has  a Json generator, but it is for a representation of the model, not the instances of the model. See Generator_CodeJson.ump

Also Umple has a parser for Json. See Json_Code.ump.

For serialization, associations should probably become arrays. But some mechanism for dealing with cycles would be needed.

For the deserialization, it would seem necessary to be able to call the deserialization of associated objects, and again how to deal with cycles would seem necessary.

In both the above cases, it would seem that perhaps an objectID ought to be stored in the Json, so the same object encountered twice would result in any other attributes being stored.

Nava2:
Java 7 (8?) added the [javax.json](http://www.oracle.com/technetwork/articles/java/json-1973242.html) packages. If each generator provides it's own solutions, that'd probably be best. 

Solving recursive models requires some form of "cached" identifier, the `objectId` approach is how [Jackson](http://www.baeldung.com/jackson-bidirectional-relationships-and-infinite-recursion) solves this. 

Just saw this issue in my github feed, thought I'd add some info.

TimLethbridge:
Generators for json-ld https://json-ld.org  or Turtle (.ttl) https://www.w3.org/TR/turtle/ should also be considered, but for outputing instances, and also outputting class diagram schemas, perhaps compatible with Schema.org https://github.com/schemaorg/schemaorg / https://schema.org

TimLethbridge:
In addition to Json, other formats should be considered. 

[...]

PR #2128, opened 2024-04-16
#1067-part2-Deserialize json string when suboption '-s genJson' is specified

Wxin-Duan:
When user specify '-s genJson', umple should not only produce toJson() to serialize Json string, but also fromJson() to deserialize Json string and instantiate validate objects.


UmpleToJava/UmpleTLTemplates/toJson_declare.ump
```diff
@@ -180,17 +190,7 @@ class UmpleToJava {
...
-  * Generate Json for this object and connected objects visited objects to enable avoidance of infinite loops
-  *
-  * @param toJsonOutput  Output is aded to this as the network of objects is traversed
-  * @param visitedList  Every concrete object visited is added so we don't re-outpu
-  * @param nestLevel    As we output deeper objects, indent them more
-  * @param atConcreteClass false when we are recursing to a superclass
-  *     so we get the superclass data
-  * @return whether or not anything was output (so we can tell whether we need to output a comma)
-  */  
+  toJson_declare <<!<</*toJson_declare*/>>
```
TimLethbridge, line 209: Why did you remove this block of comments for the toJson?

Wxin-Duan: I moved this comment to the very top of the class, because the @return statement explains what toJsonHelper returns.

TimLethbridge: Ah I see. It was in the wrong place! ... but there should be a comment generated before toJson (although it would not have any param descriuptions)

Wxin-Duan: For sure, I just added a comment before toJson describe it would return a string in Json format

PR #2134, opened 2024-04-21
update Generator_CodeRTCpp.ump to fix zip file problem

JunpengWen:
## Pull request description and content
This pull request resolves an issue where the zip file button failed to package all files into a ZIP archive. Previously, some files were omitted during the compression process, leading to incomplete archives. Now all output files, including RTCpp code generated by the command line, are stored in the folder 'RTCpp'.

### DIFF
--- a/cruise.umple/src/generators/Generator_CodeRTCpp.ump
+++ b/cruise.umple/src/generators/Generator_CodeRTCpp.ump
@@ -51,10 +51,9 @@ class RTCppGenerator
   
   protected void generateContents(UmpleModel model ,StringBuilder model_code, String filename, String content, String owingFolder) throws IOException {		
 		 String path = model.getUmpleFile().getPath() + "/";
-		 if(owingFolder.equals("RTCpp")){
-                      owingFolder = "";
-                 }
-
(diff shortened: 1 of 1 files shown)


TimLethbridge:
This doesn't seem to work properly, although it is in the right direction. If I use the following Umple Code

```
class X {
  a;
}
  
class Y {
  b;
  1-- * X;
}
```

and generate a zip file in UmpleOnline I get a tree as follows, with a directory RTCpp containing an internal directory also called RTCpp that has three files, and the class files being within the outer RTCpp

|-- RTCpp
|   |-- RTCpp
|   |   |-- CMakeLists.txt
|   |   |-- model_Main.cpp
|   |   `-- model_Model.h
|   |-- X.cpp
|   |-- X.h
|   |-- Y.cpp
|   `-- Y.h

I tested this both on my Mac, and also temporarily on the UmpleOnline test server  https://umple.org/test   ... the 7 files should all be at the same level in the outer RTCpp, and there shouldn't be an inner one.

TimLethbridge (changes requested):
This does not work as it creates an inner RTCpp directory. See may comment at the top level of this PR

TimLethbridge (approved):

PR #2139, opened 2024-04-29
Alternative way for automated Cpp testbed

JunpengWen:
## Pull request description and content
This pull request is an alternative way to create an automated test for integrating the C++ testbed into the Gradle and Ant build processes. If this does not work please check Pull request #2136.


TimLethbridge:
I will work on getting Jenkins and Appveyor to be able to resolve cpp, gcc and g++ ... but it is risky and time consuming so I will do other PRs first

PR #2188, opened 2024-10-23
Fixed copy to url issue on master

payamd:
Fixed copy URL to the clipboard when users click on save URL several times.


TimLethbridge:
This was a temporary branch. Now the full functionality is merged to master, we are closing this

Issue #1966, opened 2022-08-11
Re-layout of editable Class diagrams using Graphviz as the cue


TimLethbridge:
See comments further down, since this is done, with some bugs and improvements

There needs to me a 'relayout' option for our editable diagrams.

One possibility could work as follows.

It would first in a hidden way generate a GvClassDiagram .gv file, then scan this for the locations of each class, then move each class to the location it would be in the .gv diagram. (some mathematical transformation would need to be applied as the coordinate systems are different).

Existing association layouts would have to be removed and defaulted (the layout code would be replaced for this to work).

It would need to be subject to 'undo' so the previous layout could be restored if the user does not like it.

It is uncertain whether this would give good results, but might be worth a try.

The original editable diagrams still work adequately for many purposes; the gv layouts often work well, however. We considered bringing in other third-party tools (See #600), such as joint.js, but this was a big project that had too many downfalls.

TimLethbridge:
Work on this is underway in other issues

TimLethbridge:
This feature has been partly done, but there are bugs, so it needs continued work to improve its UX

PR #2198, opened 2024-11-13
Adding graph options for gv

### LINKED ISSUE #2081: Improve UmpleOnline editing of classes and state machines

### LINKED ISSUE #2146: Try an alternative of using orthogonal arcs in UmpleOnline graphviz diagrams
As an option in the graphviz generators we should experiment to see whether creating orthogonal arcs would be useful. This should be easy.]
These can be generated using

splines=ortho

in the graphviz.

We already have the option of making the elements separated more widely (available in the command line 
 -s gvseparator=2 for example
(the above was done in #2015 and pr #2067  but does not appear in UmpleOnline ... should be in the preferences menu along with other suboptions.

TimLethbridge:
This PR aims to do the following:

 - put options for gv graphs at the top of the graph pane
 - Enable filtering (a text field for the classes to filter, and a numeric field for hops)
 - Allowing different graph algorithms to be used, including neato, circo, and orthogonal links
 - Enabling separattion to be changed wtih the separation soption.


Teleportingsqirl:
if you are looking at model "Software Arch IEEE" in extra examples q-z, and turn on gvmanual, then gvortho, then gvcirco, the diagram cannot be loaded and will permanantly hang.

TimLethbridge:
Will close #1966 (relayout editable using gv ... and vice versa) when this is done, and #2146 (orthogonality), and #2081 in part

PR #2201, opened 2024-11-28
Clean up debug statements for collaboration

payamd:
Cleaning up the debug statements for collaboration both in client and server.


umpleonline/scripts/umpleCollab.js
```diff
@@ -7,24 +7,29 @@
...
+let baseclientID = null;  // Variable to store the client ID
+let attempt = 0;   // Variable to store the number of attempts
+let checktemp = false;  // Variable to check if the model is temporary
+let isConnected = false; // Variable to store the connection status
+let inactivityDisabled = false; // Variable to store the inactivity status
 
 // the main process of Idle timer
-let inactivityTime = 600000; // 10 minutes
-let inactivitywarningTime = 540000; // 9 minutes
+let defaultInactivityTime = 600000; // 10 minutes
```
deepankkartikey: @payamd all constants can be declared using **const** keyword instead of let

umpleonline/scripts/umpleCollab.js
```diff
@@ -22,8 +22,8 @@ let inactivitywarningTime = 540000; // 9 minutes
 let diff = (inactivityTime - inactivitywarningTime)/1000; // 60 seconds
 let inactivityTimer; // Variable to store the timer
 let inactivitywarningTimer;
-let debugFlag = false; 
-let achknowledgedTimer = 0; 
+let collabDebugFlag = false;
```
deepankkartikey: We can name this variable **collabClientDebugFlag** - just to signify it is part of client side

umpleonline/scripts/codemirror6/editor.mjs
```diff
@@ -102,41 +109,17 @@ const changeListenerPlugin = ViewPlugin.fromClass(class {
...
-       console.warn('Content changed');
-
-        //console.warn('new content:', newContent); 
-        if (debuggerFlag)
-        console.log('new content lenght:', newContent.length);
 
-        // console.warn('old content:', this.lastContent);
-        if (debuggerFlag)
-        console.log('old content lenght:', this.lastContent.length);
+        if (debuggerFlag) {
```
deepankkartikey: We can name this as **codemirrorDebuggerFlag**

umpleonline/scripts/umpleCollab.js
payamd: Sure, I will do that. We can also read these variables from the config file if you want, as you mentioned, but I think reading some variables from the config file can make it more complex.

umpleonline/scripts/codemirror6/editor.mjs
payamd: sure, I will do these changes today

UmpleCollabServer/src/server.ts
```diff
@@ -153,14 +145,10 @@ io.on('connection', (socket: Socket) =>{
   // when the client requests for them to be pulled
   // fileKey parameter should come from the client - umpdir_filename
   socket.on('pullUpdates', (fileKey: string, version: number) => {
-    // DEBUG
-    // console.log("inside pullUpdates with filekey: ", fileKey)
 
     if(serverDebugFlag){
```
deepankkartikey: We can name this variable **collabServerDebugFlag** - just to signify it is part of server side - specifically for collab server

payamd: Sure, I will change it to that.

Issue #2231, opened 2025-06-15
Umple `unique` Attribute Does Not Work with Python Generator


mcgill-sfeng:
Hello!

First of all, thank you for maintaining this wonderful initiative.

Recently I've been using the Umple -> Python generator and found out that unique attributes do not work due to an unwanted `remove` operation on a `dict` object.

Given the following Umple code as example:

```umple
class Foo {
  unique String bar;
}
```

The generated Python code looks like the following (with my selection and omission):

```python
class Foo():
    foosByBar = dict()

# my omission

    def setBar(self, aBar):

        # my omission

        if not (anOldBar is None) :
            Foo.foosByBar.remove(anOldBar)
        Foo.foosByBar[aBar] = self
        return wasSet
```

However, running this model will yield the following error:

```
AttributeError: 'dict' object has no attribute 'remove'
```

This is because `remove` in Python is not a valid operation on a `dict` object. The corresponding operation should be `pop`.

TimLethbridge:
Fixed by #2237

PR #2237, opened 2025-09-08
Issue #2231 Replaced invalid remove() calls with pop(,None) for Python dictionaries

UO-BS:
This is in relation to Issue #2231 

When converting a Java map to a Python dictionary, the remove(arg) method of maps must be converted to its equivalent pop(arg,None) for dictionaries.

## Changes

Appended a new rule to the TXL files to do the above. 
Repaired a bad test case which used the incorrect remove(arg) method.


TimLethbridge:
We note that this works for command line work: the UmpleToPython/txl/umpleJavaToPython.ctxl is being built. The command line versions work

But the copy of that file in umpleonline/scripts/txl/umpleJavaToPython.ctxl is not being made.

Actually it is a very stinky smell that there is a dupliucate file.

The solution is as follows: take umpleonline/scripts/txl/umpleJavaToPython.ctxl  out of version control. The modify the build script that does packageUmpleOnline to copy the file every time it is run. (that script in ant is build.umple.xml line 461 (look for the related place in Gradle)

TimLethbridge:
I notice that after testing this PR the following Umple code generates Python with cases of 'remove' in it for the association unlinking situations (as well as 'pop' for the unique case as highlighted in this PR). Please verify that the 'remove' for the associations is valid Python (I am not a Python expert. 
```
class X {
  unique a;
}

class Y {
  b;
  1 -- * X;
}
```

TimLethbridge (commented):
See my comments in the mainline comments of this PR. I would understand it if the build issue might be addressed in a separate PR (to copy the compiled txl in build rather than during development ... and that could also address the ideal situation of avoiding dumping the txl into the output). But I am not sure about the additional remove cases ... whether they need addressing in this PR

Issue #2242, opened 2025-09-19
Many python tests pass but with incorrect Python code according to pylint


TimLethbridge:
The list below is the result of running pylint in the directory

cruise.umple/test/cruise/umple/implementation/py

The command run is:

pylint *.py.txt | grep ': E'
(hint had to instally pipx using homebrew on my mac and then use /Users/tcl/.local/bin/pylint *.py.txt | grep ': E')

```
2:BasicPostcondition1.py.txt:41:37: E0001: Parsing failed: 'unterminated string literal (detected at line 41) (BasicPostcondition1.py, line 41)' (syntax-error)
4:BasicPrecondition1.py.txt:34:37: E0001: Parsing failed: 'unterminated string literal (detected at line 34) (BasicPrecondition1.py, line 34)' (syntax-error)
6:ClassTemplateTest_AbstractClassAvoidingInstantiation2.py.txt:40:37: E0001: Parsing failed: 'invalid syntax (ClassTemplateTest_AbstractClassAvoidingInstantiation2.py, line 40)' (syntax-error)
8:ClassTemplateTest_ExtraCode.py.txt:24:1: E0001: Parsing failed: 'expected an indented block after function definition on line 21 (ClassTemplateTest_ExtraCode.py, line 24)' (syntax-error)
10:ClassTemplateTest_Import.py.txt:54:50: E0001: Parsing failed: 'invalid syntax (ClassTemplateTest_Import.py, line 54)' (syntax-error)
12:ClassTemplateTest_MethodCommentWithEmptyLines.py.txt:50:9: E0001: Parsing failed: 'invalid syntax (ClassTemplateTest_MethodCommentWithEmptyLines.py, line 50)' (syntax-error)
14:ClassTemplateTest_MethodInlineComment.py.txt:28:9: E0001: Parsing failed: 'invalid syntax (ClassTemplateTest_MethodInlineComment.py, line 28)' (syntax-error)
16:ClassTemplateTest_MethodMultilineComment.py.txt:28:9: E0001: Parsing failed: 'invalid syntax (ClassTemplateTest_MethodMultilineComment.py, line 28)' (syntax-error)
18:ClassTemplateTest_MultipleMethodComments.py.txt:28:9: E0001: Parsing failed: 'invalid syntax (ClassTemplateTest_MultipleMethodComments.py, line 28)' (syntax-error)
[...]

PR #2255, opened 2025-10-04
#Issue 2243 Fix for Python tests not being run

UO-BS:
## Overview of changes

There was a bunch of Python tests in the /implementation/py/ folder that were not running. Usually, this was because they were copied over from other languages, but never actually verified.

I have gone through and made various changes:
- Some tests are now "global" (they run for Java, Python, Ruby, Php)
- Some tests are Python exclusives (and now actually run)
- Some tests were deleted (they were created as Java exclusives and incorrectly copied)

## Remaining Issues

There are a few tests that remain not running in the /implementation/py/ folder:
- Some tests have outputs for all languages but aren't run anywhere (BooleanIsApim, DependInterface, NoNamespaceinterface)

Many of the unit test outputs fail a python linter... See Issue #2242


TimLethbridge (approved):
This looks good. I will merge

PR #2261, opened 2025-10-23
Issue 2249

jacquesjack:
## Pull request description

modified the TXL in MethodTranslation.txl to make a new TXL function that scans java methods and store all local variables in a list. In StatementTranslation.txl there is a new where condition in one of the rules that checks to see if the variable is local or not before changing the variable to be self_.varname.


jacquesjack:
I was intending to add a new test case to test a class having an instance variable called wasSet. Since wasSet is used as a local variable in the code that umple generates this would test if umple can generate code correctly with a local variable and an instance variable with the same name.

Sadly umple is not able to correctly do this. The issue does not lie with the Java -> Python TXL. In this particular case it is the java that gets generated incorrectly. 

If the Java code is generated correctly I do believe that the Java -> Python TXL would be able to handle it since there is a specific TXL rule that translates all cases of this.foo -> self._foo. I have not tested this though.

This umple code:

```
class Student
{
  wasSet;
}
```

Generates this Java code:
```
public boolean setWasSet(String aWasSet)
{
boolean wasSet = false;
wasSet = aWasSet;
wasSet = true;
return wasSet;
}
```

Instead of the correct Java code:
```
public boolean setWasSet(String aWasSet)
{
boolean wasSet = false;
this.wasSet = aWasSet;
wasSet = true;
return wasSet;
}

**Comment by jacquesjack** [comment]
Also I am not sure why the checks fail here? When I do a bumple on my branch everything is correct. the CI tests say that there is a test in the PythonClassTemplateTest that fails but this test does not fail on my branch so I'm not sure what is happening.

**Comment by TimLethbridge** [comment]
bumple reports errors! Just because it says BUILD SUCCESSFUL doesn't mean it is. You have to look for failed tests. Here is what I get on my machine with this branch

```
BUILD SUCCESSFUL
Total time: 5 minutes 22 seconds
************************
Script bumple ended ABNORMALLY at /Users/tcl/umple2
The word FAILED or ERROR was found in the above. Build was NOT SUCCESSFUL.
Build log is at /tmp/umplebuildlog14622.txt
Test log can be opened in a web browser at /Users/tcl/umple2/dist/qa/index.php
The following is a snippet of the log with the problem
269-    [junit] WARNING: multiple versions of ant detected in path for junit
270-    [junit]          file:/Users/tcl/unix/github/umple/cruise.umple/bin/org/apache/tools/ant/Project.class
271-    [junit]      and jar:file:/Users/tcl/unix/github/umple/dist/libs/core/ant.jar!/org/apache/tools/ant/Project.class
272:    [junit] TEST cruise.umple.implementation.py.PythonClassTemplateTest FAILED
273-[junitreport] Processing /Users/tcl/unix/github/umple/dist/qa/cruise.umple/TESTS-TestSuites.xml to /var/folders/31/n59x86ys06xgjj5fzh07tg340000gn/T/null1960231372
274-[junitreport] Loading stylesheet ile:/opt/homebrew/Cellar/ant/1.10.15_1/libexec/lib/ant-junit.jar!/org/apache/tools/ant/taskdefs/optional/junit/xsl/junit-frames.xsl
275-[junitreport] Transform time: 1344ms
```

Issue #510, opened 2015-08-26
Allow generation of instance diagrams


Nava2:
Originally reported on Google Code with ID 612

---

As a new generation target, allow generation of a 'random' instance diagram from any
class model. Essentially the generator would walk the class diagram and generate an
instance diagram that complies with the class diagram (with random numbers and strings
for attributes) and links that conform to the multiplicity (Objects with 0, 1 and 2
where there is a \* association for example)

The reason for this is to facilitate teaching. Each run of the generator would generate
a different diagram.

See also issue 611, which proposes doing something similar with state machines.

---

Reported by @umple on 2014-07-24 14:27:19

TimLethbridge:
This would require walking the class graph to determine possible starting points for creating instances, and generating random numbers of instances with random values for attributes that are consistent with association multiplicity (but perhaps with a little intelligence, e.g. if an attribute looks like a name, then generate human names)

TimLethbridge:
The following is the first draft of a rough algorithm for creating randomized numbers of objects of each class.

You will have to consider many special cases:
 * making sure it works with reflexive associations
 * there may be runaway cases; it may be necessary to have a counter that stops adding new instances after, say 3 times the number of classes in the model.
 
Except for the two classes, this would all work within the generate method for the gvInstanceDiagram. However, the following blocks likely would be their own methods that would have to be passed necessary arguments.

Create a class UmpleInstance that you will use for each node, and which would have the following associations 
  * -> 1 UmpleClass;
  String [] attributeStrings;
  
Create a class UmpleLink that you will use for each arc
  * -> 1 UmpleAssociation;
  * -- 1 UmpleInstance fromInstance;
  * -- 1 UmpleInstance toInstance;

Create the following data structures as a local variable of the generate method

  // Objects that don't have all their associations yet
  ArrayList<UmpleInstance> partiallyCreatedObjectsQueue;
  
  // a structure to allow us to quickly look up the list of instances of each class
  HashMap instanceMap
    key UmpleClass
    value List<UmpleInstance>

A. Loop through the classes in the model
  Add each to instanceMap with an initially empty list of UmpleInstance
  
B. Iterate through all keys of InstanceMap (i.e. through all classes)
  If the value is empty (no instances yet) then we need to find the next starting class from the ones that don't have instances:
     Tentatively mark the first such class as the starting class, but keep looking
     If any class is encountered with a 1-- association then that has priority, and pick the first of these, but keep looking;
[...]

PR #2265, opened 2025-11-21
Issue510 Instance Diagram Generator

UO-BS:
## Adding an Instance Diagram Generator (Issue 510)

This PR is to resolve Issue #510 , which involves the creation of a new generator for Umple. The goal is to generate a "random" instance diagram from a class model for teaching purposes.

## Required Changes

This will involve a new file "InstanceDiagramGenerator" to be created with all the code for this generator.
It will involve some minor changes to the grammar, compiler, and UmpleOnline to allow the user to actually call and generate an InstanceDiagram
There should also be some new tests created to ensure that the generator is working properly.


jacquesjack:
I did this last commit through vs code and I think it just resused a commit message from way before. Disregard this commit message. The true commit message is  "I did not push all of the changes that I wanted to push in my previous commit."

jacquesjack:
Added a new condition on the while loop that ensure no infinite looping

TimLethbridge:
I am planning to do the final test and merge this this morning.

PR #2287, opened 2026-01-23
[2274] starts umple online in G mode

kyle-labelle:
Umple online now starts in G mode instead of E mode. This is a small fix in the umple.php file in the umpleonline directory. It changes the default state of diagram type from class to GvClass and makes the "G" button active by default instead of the "E" button on the UI side.


kyle-labelle:
For issue 2274, not 2214

PR #2313, opened 2026-02-04
Fixes #2263: Remove ghost markers and implement fallback for null injections

### LINKED ISSUE #2263: Code Injection functions for different languages will put comments in any language
When using functions with code injections which are specified in a language, Umple will attempt to inject the code into any place it should be found but then remove the code injected instead of just skipping the injection altogether.

For example when generating the following code in either Java or Php
```
class Student {
  String foo(int a) Java {
    return 1
  }

  String foo(int a) Php {
    return 1
  }

  before custom foo Java {
    System.out.println("Testing Injection");
  }

  before custom foo Php {
     print "Testing Injection";
  }
}
```

We get the function contents as follows, with both injection comments one of which is empty:
```
// line 12 "model.ump"
System.out.println("Testing Injection");
// END OF UMPLE BEFORE INJECTION
// line 15 "model.ump"
 
// END OF UMPLE BEFORE INJECTION
return 1
```

The result should only show the comments from injections in the valid language:
```
// line 12 "model.ump"
System.out.println("Testing Injection");
// END OF UMPLE BEFORE INJECTION
return 1
```

Fadiedy:
**Description** 
This pull request addresses the generation of unwanted "ghost" markers (placeholders such as //Php or #Ruby) during code injections. The logic within GeneratorHelper.ump has been refined to prioritize constraint-based code generation while ensuring stability for existing PHP and Ruby test cases.

**How it works**
The updated logic implements a more precise filtering and fallback mechanism:

1. Constraint Fallback: When a generator and constraint tree are present, the system first attempts to get code via getConstraintCode. If this returns null or an empty string, it now falls back to inject.getCode(). This ensures that valid code meant for specific platforms is not lost if the constraint evaluator returns an empty result.

2.  Refined Skip Condition: The loop now only executes a continue (skipping the injection and its associated line markers/headers) if the injectedCode is truly null.

3. Ghost Elimination: By preventing the generation of line markers and footers when no actual code content exists, it removes the placeholder comments that were previously cluttering the output, resolving the "ghost marker" issue.

**Automated Tests**
The changes have been verified against the existing Umple test suite to ensure no regressions. I have resolved 9 specific test failures where the generator was previously inserting ghost markers or cross-language code. The following test cases now pass successfully after updating their baseline expectations:

1. JavaPhpRubyNoNullMethodInjectionTest (Java, PHP, Ruby): Fixed 3 failures where the tests expected ghost comments (//Java, //Php, and #Ruby). These are now correctly skipped when the injection is null.

2. ClassTemplateTest (PHP & Ruby): Resolved 4 failures (ClassCodeInjections_SingleLine, _Basic, and _Comments). These fixes ensure that Java-specific code no longer leaks into PHP/Ruby files and that unnecessary whitespace created by ghost markers is removed.

3. CodeInjectionTest (StateMachines for Java & PHP): Fixed 2 failures where Java-specific state machine logic (e.g., light = aLight;) was being incorrectly injected into other language outputs.

**Verification**
A full build was performed, and the build transcript was verified to ensure that no "error" or "failure" messages appear. The generated code now correctly aligns with the expected baseline .txt files without including unnecessary placeholder markers.

### DIFF
--- a/cruise.umple/src/generators/GeneratorHelper_CodeClass.ump
+++ b/cruise.umple/src/generators/GeneratorHelper_CodeClass.ump
@@ -54,64 +54,84 @@ class GeneratorHelper
     }  
   }  
   
-  public static String toCode(List<CodeInjection> allCodeInjections)
-  {
+  public static String toCode(List<CodeInjection> allCodeInjections) {
     String asCode = null;
-    if (allCodeInjections != null)
-    {
-      for (CodeInjection inject : allCodeInjections)
-      {
-        mixset Mixset {
-	      if(inject.hasCodeLabel())
-          continue;// handle the case when labels are used. // Do nothing  
-        }
-        String comment = "//";
-        mixset RubyGeneratorIntMixset {
-          comment = RubyGenerator.class.isInstance(generator)?"#":"//";
-        }
-        String positionString = "";
-        Position p = inject.getPosition();
-        Position codeP = inject.getCodePosition();
-        String injectFooter = "";
-        String injectType = inject.getType();
-        if (codeP != null) {
-            positionString =  comment + " line " + codeP.getLineNumber() + " \"" + inject.getUmpleClassifier().getRelativePath( p.getFilename(), inject.getUmpleClassifier().getSourceModel().getDefaultGenerate() ) + "\"\n";
-            if (injectType != null) {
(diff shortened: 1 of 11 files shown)


TimLethbridge:
Fixes #2263

TimLethbridge:
Excellent work! When the merge CI check finishes, I will merge this. I have tested it manually

PR #2333, opened 2026-02-19
2224 drawing in g mode

kyle-labelle:
Added the drawing palette from E mode to G mode on umpleonline. Created new actions in umple_action.js to support deleting, adding and doing associations and generalizations between classes while in G mode from the "Draw" palette.


kyle-labelle:
I have added the fixes for the association and generalizations to be coloured based on the theme, for them to start in the right spot when clicked, and for the new class names to not be reused when adding classes in G mode.

TimLethbridge:
@kyle-labelle I am merging this as we discussed in the meeting as it is an important improvement. There are some details that can be improved 

*  enabling mouse-down - mouse-up operation 
* seeing if there can be cursor shapes enabled for various states (initial selection of the tool, and while rubber banding)
* some kind of rule that says if the focus is lost on the canvas then the operation is abandoned

PR #2335, opened 2026-02-20
Fixes issue 2290 enabling more options on UmpleOnlin RHS

TimLethbridge:
It will be useful to be able to display many of the generation targets such as ERDs and CRUD UI on the right hand side. This adds a menu that will enable the user to choose more than E, G and S mode to appear aon that side, live when any edit to the model occurs.

This being done in small commits.


TimLethbridge:
I have tested this and it is working nicely. In the interests of being agile I am going to merge it. Then in a new branch (with new PR) we can go on to add the instance diagram and CRUD UI and so on.

PR #2338, opened 2026-02-22
Issue #2273: Fixed double redraw when loading examples by ignoring non user codemirror doc updates

### LINKED ISSUE #2273: Umple redraws G mode diagrams twice
This is a recurrent issue. Following recent work, Umple now seems to redraw class and state diagrams twice when they are displayed. This issue happened in the past and now seems to be happening again. Be careful not to break other things when fixing this.

EDIT Jan 24: This seems to be the case only when an example is loaded.

### LINKED ISSUE #2336: UmpleOnline: Example triggered diagram will be refreshed twice, hiding the generated output
## Brief Description

UmpleOnline has two related refresh bugs in the editor/diagram flow:

1. Generated output (code/image/Javadoc/etc. in the bottom panel) can be hidden after diagram refresh even when the Umple model did not change.
2. Loading an example triggers two diagram refreshes: one immediate refresh (expected) and a second delayed refresh (~1.5–2s, unnecessary), caused by programmatic editor updates being treated as user typing.

### Output visibility path
1. Open UmpleOnline.
2. Load any example.
3. Immediately, click a generate action (e.g., Generate Java) so bottom output is visible.
4. Observe output visibility and refresh behavior.

### Double-refresh path
1. Open UmpleOnline.
2. Select any example.
3. Observe diagram refresh timing (immediate + delayed second refresh).

## Expected Behavior

- Generated output remains visible when the model is unchanged since last generation.
- System (programmatic) input in CodeMirror should not be treated as user input; it shouldn't refresh immediately after changes are made.

## Actual Behavior

- Generated output may be hidden by a later refresh callback even when model content is unchanged.
- Example selection causes two refreshes:
  - immediate refresh from example load,
  - delayed second refresh (~1.5–2s) from debounced typing listener in CM6.

https://github.com/user-attachments/assets/f33b9c4d-2a29-4268-9f6b-0332614ed0c0

## Root Cause Summary

- Output hide logic was tied to refresh callback timing rather than model-change detection.
- CM6 change listener debounced typing pipeline processed programmatic text replacement as user input.

Aliloraf:
Edit: Fixes #2273 and Fixes #2336
When loading examples, codemirror performs doc updates that were treated like user typing. This triggered debouncedProcessTyping causing Action.processTyping to run redundantly and resulting in a double redraw.
I have put a gate on the debouncedProcessTyping call so it only runs for real user edit transactions. 

Files changed: 
scripts/codemirror6/editor.mjs 
scripts/codemirror6/editor.bundle.js

### DIFF
--- a/umpleonline/scripts/codemirror6/editor.bundle.js
+++ b/umpleonline/scripts/codemirror6/editor.bundle.js
@@ -25218,18 +25218,34 @@ var cm6 = (function (exports) {
      update(update) {
        
        if (update.docChanged) {
-           
-           const newContent = update.state.doc.toString();
-           
-           if (newContent !== this.lastContent) {
-           const currentPositionofCursor = this.view.state.selection.main.head;
+            // Debug: Issue 2273: only react to user edits to avoid triggering processTyping
+            let userEdit = false;
+            if (update.transactions) {
+                userEdit = update.transactions.some(tr =>
+                (typeof Transaction !== "undefined" && tr.annotation && tr.annotation(Transaction.userEvent)) ||
 
-           this.lastContent = newContent;
+                // fallback 
+                tr.isUserEvent("input") || tr.isUserEvent("delete") || tr.isUserEvent("paste") ||
+                tr.isUserEvent("drop") || tr.isUserEvent("cut") || tr.isUserEvent("undo") ||
+                tr.isUserEvent("redo"));
+            }
 
-           debouncedProcessTyping("newEditor", false ,currentPositionofCursor); // call the debounced function
-         }
-       }
-       Action.updateLineNumberDisplay();
+            if (!userEdit) {
+                Action.updateLineNumberDisplay();
+                return;
+            }
+
+            const newContent = update.state.doc.toString();
+
+            if (newContent !== this.lastContent) {
+                const currentPositionofCursor = this.view.state.selection.main.head;
+
+                this.lastContent = newContent;
+
+                debouncedProcessTyping("newEditor", false, currentPositionofCursor);
+            }
+        }
+        Action.updateLineNumberDisplay();
      }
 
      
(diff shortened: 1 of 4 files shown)


ElwinLiu:
hello, nice attempt to avoid extra processTyping calls, but this guard is too broad. I see that you're filtering out the non-user edits, but it also drops:

1. Valid docChanged transactions from UI flows that use dispatch({changes: ...}) without a userEvent annotation. 
2. Any edit done through Page.setCodeMirror6Text(...). 
3. Collaborative incoming edits that are from receiveUpdates dispatch.

Try to edit through UI flow, then you should see it won't be refreshed. Check the image below:

image


My suggestion is that, instead of allowing user-edits only, we allow all edits except some programmatic edits. Any edit that should be skipped must not rely on codemirror for diagram refreshing.

I've submitted a fix, maybe it can help you:
[fix(umpleonline): avoid duplicate refresh on example load](https://github.com/umple/umple/commit/afd9adb77f5664a899273119487d3c7d5bbfce2e)

TimLethbridge:
@Aliloraf  I am moving this to Draft, pending using the fix as pointed out by @ElwinLiu, and further extensive testing. There is a high risk of breakage of other functionality if this is not treated carefully.

Aliloraf:
Integrated Elwin’s codemirror6 annotation based approach to skip debouncedProcessTyping instead of gating on isUserEvent.

The fix now uses a skipDebouncedTypingAnnotation to suppress debouncedProcessTyping only for programmatic example load updates. It also ensures generated output is only hidden if the model actually changes which fixes issue #2336.

Tests:

- Example load: single refresh (no delayed redraw)
- Typing / paste(including large paste) / delete / undo / redo (multiple sequential operations)
- UI context-menu edits (rename / change type / delete attribute)
- Diagram and text sync
- Generate output stays visible unless stale (issue 2336)
- Browser console shows no new errors
- Rapid switching between examples
- Selection replace

ElwinLiu (approved):
LGTM

PR #2340, opened 2026-02-23
[WIP]FIx #2279 wrong line number returned from parser

### LINKED ISSUE #2279: Wrong line number for errors and warning while they are at the last line of the file
## Environment
- Apple M1 Pro (Thoe 26.2)
- Umple Version 1.35.0

## Description
Umple compiler (umple.jar and umplesync.jar) will wrongly report the line number where the errors and warnings are if they are at the last line of the ump file. And it will work properly if simply add a newline after the last line.

## Reproduce the Issue

1. Write a simple Umple file 
2. Put any errors or warnings at the last line without any further newline characters
3. Use umple.jar or umplesync.jar to compile it 

Image
Image

## Solutions 

An easy solution can be letting compiler add newlines after it's compiled anyway to avoid the error line becoming the last line. Otherwise, some logic in compiler might need to be modified to recognize this kind of scenarios correctly

skyfackr:
Fixes #2279

The underlying bug for this issue was originally addressed in commit `8e98550c2527a97952f58e34074aafe729aeedac` (2014). However, that change only mitigated the problem by incrementing the parsed line number by 1, which left the edge case reported in #2279 unresolved.

The root cause is that the line-number parsing logic depends on the offset of the first character of the **next line** relative to the `lookfor` offset, instead of the first character offset of the **current line**.

This causes two problems:

1. The parsed line number can be incorrect.
2. When the `lookfor` offset points to the last line, there is no "next line", so the parser effectively skips the last line.

As a result, the parser fails to return the correct line number.

The original parser also contains a fallback that sets the line number to `1` if the parsed result remains `0` after the parsing logic runs. This is intended as a safeguard when parsing fails. For `lookfor` offsets on the last line, the code always falls back to this path, which is why it always returns line `1` for the last line.

This PR fixes the issue by reworking the line-number parser into a two-stage process:

- **Stage 1:** Build a line map from the input, where each line is mapped to its start and end offsets.
  - The range format is `[start, end)`.
  - `start` is the offset of the first character of the current line.
  - `end` is the offset of the first character of the next line, or the total input length for the last line.
- **Stage 2:** Determine which line contains the `lookfor` offset by checking which range it falls into.

The fallback that returns `line = 1` when parsing yields `line = 0` is intentionally kept to preserve existing behavior, although it should no longer be reached in normal execution.

This PR also includes some newly ignored Git config files. See the separate inline comment for details.

---

<sub><em>Disclaimer: Some non-code text in this PR (such as comments or the PR description) may be proofread by AI. The code changes themselves were designed and implemented by a human developer. AI may be used for bug checking and formatting improvements, but not for implementation.</em></sub>


.gitignore
```diff
@@ -97,4 +97,11 @@ hs_err_pid*
 # Gradle
 .gradle
 
-__pycache__/
\ No newline at end of file
+__pycache__/
+
+# Python venv
+.venv/*
+venv/*
```
skyfackr, line 104: To ignore the python virtual env folder. This can help create a isolated python environment from the host machine's main env, to avoid some outside effect

.gitignore
```diff
@@ -97,4 +97,11 @@ hs_err_pid*
...
+
+# Direnv local SHELL ENV settings
+/.envrc
```
skyfackr, line 107: For the tool `DirEnv`, which can set specific env value/path value when enter the project path. Used for setting `UMPLEROOT`, `OSTYPE`, and add the `dev-tools` to the `PATH`

TimLethbridge:
I have come up with a vastly simpler solution to this .. @skyfackr please contact me in Teams to arrange a meeting

TimLethbridge:
Fixes #2279

PR #2342, opened 2026-02-24
Add Change Color option to State Diagram context menu (Fixes #1958)

### LINKED ISSUE #1958: Ability to colour items through a menu, not just textually. Add to state diagrams, and fix in clkass diagrams.
Currently classes can be coloured using an attribute color. There is an example of this in Manufacturing Plant Controller example. A feature was added to the menu of G mode class diagrams to change the colour, but it currently only changes to black! (new bug)

State can also be coloured; A similar menu needs to be added for that.

ORIGINAL:
It should be possible, however, to specify the colour in a menu in editable view. Even in Graphviz view, this could be considered as one of several menu editing items that could be added.

Samoumine:
**IMPORTANT:** I could not fully test this locally due to environment issues. Please verify functionality before merging.

I have made the necessary code changes to add the "Change Color" option to the state diagram context menu, but I recommend testing it thoroughly before merging.

This PR adds a "Change Color" option to the state diagram context menu.

Changes:
- Added menu entry in Action.drawStateMenu
- Implemented colorState case in Action.drawInputState
- Added Action.setColorState to insert displayColor into state block
- Triggered diagram refresh and saved edit history

The state color is updated using the displayColor attribute and refreshes correctly.


TimLethbridge:
As with changing colours on classes, this does not currently work on Safari ... it needs investigation on that platform ... I am sure there is an easy fix (adjusting the way the colour picker activation is captured I think).

Also, I notice that if the colour already present, this adds a second one, when it should be replacing the original.

And if possible, can it line up the text more nicely (same indentation as other items in the state, indented 2 spaces from the braces?

TimLethbridge (changes requested):
Please change this so that if there is already a colour statement in the state, then it will change the colour, rather than adding a new colour statement in that state.

PR #2351, opened 2026-02-26
Fixes #2350 remove stale states check for generated code

### LINKED ISSUE #2350: Urgent regression: After PR  #2338 major failure in error reporting
After commit  3eab93f6debddac13741981d99561d074cbb5293 for PR #2338  fix(umpleonline): avoid duplicate refresh on example load

There has been a major problem:

  1. If the server returns no error, it does not erase the previous one at the screen bottom
EDIT: There second problem was an error and so there is only the above. 

https://github.com/umple/umple/pull/2338

ElwinLiu:
## Pull request description

After commit `3eab93f`, UmpleOnline sometimes kept showing an old compiler error at the bottom even after the next compile succeeded.

We introduced logic to avoid hiding generated output when it doesn't need to be updated.  
However, the call to `Page.hideGeneratedCode()` was also responsible for hiding the previous error output area.

So once we guarded that call with a stale-check condition, this side effect happened:

- generated output considered "not stale" -> `hideGeneratedCode()` not called
- old error message remained visible after a successful compile

## Pull request content

This PR is intentionally small and focused on issue #2350 only. It reverts the stale check logic from commit 3eab93f **only**.

### DIFF
--- a/umpleonline/scripts/codemirror6/editor.bundle.js
+++ b/umpleonline/scripts/codemirror6/editor.bundle.js
@@ -25258,7 +25258,6 @@ var cm6 = (function (exports) {
    exports.getSyncedVersion = getSyncedVersion;
    exports.onDarkModePreferenceChange = onDarkModePreferenceChange;
    exports.prefersDarkMode = prefersDarkMode;
-   exports.programmaticChangeAnnotation = skipDebouncedTypingAnnotation;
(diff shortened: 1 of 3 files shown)


ElwinLiu:
@codex review the commit and confirm this change is scoped. It should only revert stale-check logic introduced in commit 3eab93f

PR #2360, opened 2026-02-27
Issue #2301: Cross-file error filename display in UmpleOnline

### LINKED ISSUE #2301: Cross-file error messages in Umple Online should also mention the file name where the errors are raised
## Environment
- Apple M1 Pro (Thoe 26.2)
- Umple Version 1.35.0

## Description
When there are multiple ump files in the workspace, the errors in imported files are reported with only the information of line number and error type . For example, **there are A.ump uses B.ump in the same workspace in Umple Online, there is an error at line 4 of B.ump, when A.ump is compiled, we will only see the error message mentioning there is an error  at line 4 without mentioning the file name B.ump. This can be confusing because the error doesn't come from A.ump so the line number is invalid**

## Reproduce the Issue

1. Write two Umple files and make one import another one as a reference
2. Put an error in the imported one and go back to another file
3. The error message will show without file name information

Image

Image

## Solutions 

The 'filename' field in the error message given by Umple compiler indicates where does the error come from, so the only problem is which line should the error be put at:

1. **Easy solution:** Put the error at the first line and mention the file name where the error actually comes from.
2. **More accurate solution:** Construct an import map, and do a search algorithm (e.g. DFS, BFS) on each "use" statement in the file. And we put the error at the line of the "use" statement where it can search the file name that has the error.

Aliloraf:
This PR is for issue #2301 (Cross-file error messages in Umple Online should also mention the file name where the errors are raised).

The compiler already returns structured JSON errors including the filename field.
However, UmpleOnline did not display this context properly, leading to confusion when errors originated in imported files.

I updated "getErrorHtml()" in compiler.php.
What it does:
   - Ignores internal file model.ump to only display filename if the error is from another tab.
   - Uses JS to:
       - Detect active file tab (line 1070)
       - Count open file tabs
       - Conditionally hide filename when:
           - only one tab exists
            - or error belongs to currently active tab
       - Show filename only when error originates in a different file.
   
   Still need in the future to automatically switch tabs when clicking an error belonging to another file. Which I will be working on.

After tests (see images bellow):
- Single file gives simple error message  
- Multi-file shows in B.ump only when relevant  
- Internal model.ump never displayed  

image
image


TimLethbridge:
I am testing and merging branches one at a time. This is next in the queue

TimLethbridge (changes requested):
This is not sufficient yet.

If I make an mistake on line 6 in  one tab (I call it File1), then in another tab I say

```use File1.ump;```

The message that appears is

```Warning in File1.ump on line 5 : .....```

a) There are two problems: It is getting the line wrong (off by one).
b) The hyperlink under 'line 5' points to line 5 in the current file. It instead should run some Javascript that flips to the correct tab and then highlights the correct line there (the link could be under the entirety of the text 'File1.ump on line 6'). There may need to be some kind of delay in the javascript function that does this to give the other tab time to load.

TimLethbridge:
We talked about improvements in the meeting today ... I have brought this up to date with other PRs merged into master

Aliloraf:
Addressed the remaining review comments for issue #2301.

- Updated compiler.php:
  - Fixed off-by-one issue by adjusting line number calculation based on whether the error is cross-file or same-file
  - Modified error link generation to distinguish between same-file and cross-file behavior

- Updated umple_action.js:
  - Added Action.goToCrossFileError to handle navigation to the correct tab and line

Cross-file errors now switch to the appropriate tab and highlight the correct line, while existing same-file behavior remains unchanged.

Let me know if any adjustments are needed.

TimLethbridge:
Tested and it works!

PR #2366, opened 2026-03-06
Generate Random Data in CRUD, Store and Load JSON

### LINKED ISSUE #2334: Sending randomly generated instance models to CRUD interface, and sending CRUD interface data to instance diagram generator
Currently we have two generators that we would like to get to work together

*  InstanceDiagram will randomly create a set of objects corresponding to the model and then generate a gv instance diagram. The code for this is in cruise.umple/src/generators/instanceDiagramGenerator/Generator_CodeInstanceDiagram.ump
* CRUD interface (issue #535)

What we want to achieve is the following:
1. When an instance diagram in gv format is generated, the json format (see below) if this can also be obtained, and can be downloaded, or loaded into the CRUD interface
2. When the crud interface is generated, the json format (same as in point 1, see below) can also be obtained, and can be downloaded, or used to generate  an instance diagram.

STEP 1:

I suggest we start by having the CRUD be able to save and load Umple JSON data format (to/from a file in the backend most likely)

To get some sample data, the suboption "genJson"; can be added to any model. Then in a main program we can call toJson() in a main program. For example if this is done in the Airkine example, we can add System.out.println(a.toJson()); to the last line of the main program.

The Umple Json format has lines like ```"umpleObjectID" : "1442407170",``` that allow readers of the json to not recurse indefinitely when they encounter the same object.

STEP 2:
Currently the instance diagram generator uses a two step process with most of the code in Generator_CodeInstanceDiagram.ump

Some refactoring will be needed so the output of gv can be done independently of the generation of the objects. Some key code for generating objects is in lkines 112 to 124 of  Generator_CodeInstanceDiagram.ump where it initiates an instance of a class ClassInstanceCountSet and generates instances in this.  That class itself starts at line 525.

I suggest the first thing to do would be to pull out that class and its initialization into a separate file that does the necessary generation of this and is then called by the original gv generation code, but can now be called independently when we want to randomly generate a CRUD model too.

STEP 3

We should alsoprovide a way to initialize the data structure used in the gv generator from Json (new method) and a way to output json from this.

STEP 4

Now in the CRUD generator we can arrange to generate Json, and then call the instance diagram generator specifying to read the json, rather than random. Similarly in the instance diagram generator, we should be able to write out the json (perhaps always writing it out along with the .gv file?) so it could be loaded into the CRUD whenever we want.

Then the CRUD interface could load data that has been generated from the instance diagram and vice versa,

The UX would have to be set up so this is quick and easy to use

thamil-vani:
This PR fixes issue #2334 . It adds end‑to‑end JSON support for the CRUD UI and plugs it into the Instance Diagram generator. The generator now has a crudRandomJson suboption that produces CRUD‑compatible JSON, exposed via a new crud_random_data.php, and the CRUD UI gets “Generate JSON”, “Generate Random Data”, and “Load JSON” buttons that all use the same import/export code so the code can save, reload, or auto‑generate instance data.

### DIFF
--- a/build/build.umple.xml
+++ b/build/build.umple.xml
@@ -385,7 +385,8 @@
       <fileset file="${scriptPath}/umple_tooltips.js"/>
       <fileset file="${scriptPath}/umple_tab_control.js"/>
       <!-- CRUD UI -->
-      <fileset file="${scriptPath}/umple_crudui.js"/>      
+      <fileset file="${scriptPath}/crud/umple_crudui.js"/>
+      <fileset file="${scriptPath}/crud/umple_crud_json_persistence.js"/>        
       <!-- Core AI modules - must be in dependency order -->
       <fileset dir="${scriptPath}/ai/utils" includes="**/*.js"/>
       <fileset file="${scriptPath}/ai/config/umple_ai_config.js"/>
--- a/cruise.umple/src/generators/instanceDiagramGenerator/Generator_CodeInstanceDiagram.ump
+++ b/cruise.umple/src/generators/instanceDiagramGenerator/Generator_CodeInstanceDiagram.ump
@@ -105,55 +105,520 @@ class InstanceDiagramGenerator
   {
     // Update meta variables based on suboptions
     useSuboptions();
-
-    // Create class instance object
-    ClassInstanceCountSet classInstanceResult = new ClassInstanceCountSet(model);
-
-    // Initialize class counts
-    if (suboptionInitialClassCountFlag == -1) {
-      Map<Integer, Double> minimumInstanceProbabilities = new HashMap<>();
-      minimumInstanceProbabilities.put(0, 0.25);
-      minimumInstanceProbabilities.put(1, 0.50);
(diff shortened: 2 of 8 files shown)


TimLethbridge:
There seems to be a glitch in your PR. I built it in both the test server and my own machine, and tried it on multiple browsers. If load an example and then generate CRUD UI, it hangs,

On investigation I see the following

```
​Uncaught TypeError: Page.showCrudFromJson is not a function
   at Page.showGeneratedCode (allumple-min.js:292:9047)
   at Action.generateCodeCallback (allumple-min.js:346:8311)
```

This may be due to the sequencing of js loading. I have such a large number of PRs to look at that I will have to move on,  and try to merge others ... and that may mean I end up merging master into yours at some point too.

TimLethbridge (dismissed):
Tested on 2 browsers and 2 machines, and fails after loading an example and generating CRUD with
```
Uncaught TypeError: Page.showCrudFromJson is not a function
    at Page.showGeneratedCode (allumple-min.js:292:9047)
    at Action.generateCodeCallback (allumple-min.js:346:8311)
'''

PR #2367, opened 2026-03-06
Validate association role names as identifiers

### LINKED ISSUE #2341: Role names can have special characters which results in bad generated code
Role names in Umple should be checked to make sure they are valid identifiers (just like attributes).

The following code will generate bad code because there is no checking for the syntax of the role name Df-dfd (it is consumed as a single token! A new error message should be created for this (with a number as close as possible to the other association error messages)

```
class X {
  implementsReq R1%R3;
  agt;
}

req R1%R3 {test}

class Y {
  1 -- * X Df-dfd;
  public static void main() {
    System.out.println ("Aha!")
  }
}
```

We also need to assess whether we should allow the requirement above to have special characters in its name too!

Ilyas-Ait-Ali:
Fixes #2341

Summary
This change validates association role names so they must be valid identifiers.

Changes
- Added identifier validation for association role names using Token.isValidIdentifier in createPreliminaryAssociationEnd(...)
- Enabled error messages 120 and 121 in en.error
- Prevented misleading cascade error E008 when an association end already fails validation and the association is not an association class
- Added parser regression test:
  cruise.umple/test/cruise/umple/compiler/2341_invalidAssociationRoleName.ump
  TestInvalidAssociationRoleNameError120()

Testing
- Tested using the issue example in UmpleOnline
- Invalid role name "Df-dfd" now correctly reports error 120 only
- Added parser regression test verifying the parser fails with error 120

Result
Association role names containing invalid characters are now correctly rejected by the parser.

### DIFF
--- a/cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
@@ -4537,7 +4537,9 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
     AssociationEnd firstEnd = createPreliminaryAssociationEnd(myEndToken, defaultMyType);
     AssociationEnd secondEnd = createPreliminaryAssociationEnd(yourEndToken, defaultMyType);
     if(firstEnd == null || secondEnd == null){
-        setFailedPosition(associationToken.getPosition(), 8, name);
+        if(isAssociationClass){
+          setFailedPosition(associationToken.getPosition(), 8, name);
+        }
         return null;
     }
     
@@ -4650,6 +4652,10 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
       if (!mult.isValid()){
         String invalidBound = bound == null ? lowerBound + ".." + upperBound : bound;
         setFailedPosition(associationEndToken.getPosition(), 4, invalidBound);
+        return null;
+      }
+      if(roleName != null && !"".equals(roleName) && !Token.isValidIdentifier(roleName)){
+        setFailedPosition(associationEndToken.getPosition("roleName"), 120, roleName);
         return null;
       }
       if(type == null){
--- a/cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
+++ b/cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
@@ -4536,6 +4536,16 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
     }
     AssociationEnd firstEnd = createPreliminaryAssociationEnd(myEndToken, defaultMyType);
     AssociationEnd secondEnd = createPreliminaryAssociationEnd(yourEndToken, defaultMyType);
+    // If createPreliminaryAssociationEnd() fails validation (e.g., an invalid role name),
+    // it reports the error and returns null for that association end.
+    //
+    // Without the guard below, analyzeAssociation() would also report error 8
+    // ("Association class is missing an end") whenever one end is null.
+    // This would create a misleading cascading error, since the real issue
+    // (such as error 120 for an invalid identifier) has already been reported.
+    //
+    // Error 8 should only apply to AssociationClasses that actually lack an end,
+    // so we restrict it to those cases using the isAssociationClass check.
     if(firstEnd == null || secondEnd == null){
         if(isAssociationClass){
           setFailedPosition(associationToken.getPosition(), 8, name);
@@ -4657,6 +4667,11 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
       if(roleName != null && !"".equals(roleName) && !Token.isValidIdentifier(roleName)){
         setFailedPosition(associationEndToken.getPosition("roleName"), 120, roleName);
         return null;
+      }
+      if(roleName != null && roleName.length() > 0 && Character.isUpperCase(roleName.charAt(0))){
+          getParseResult().addErrorMessage(
+              new ErrorMessage(121, associationEndToken.getPosition("roleName"), roleName)
+          );
       }
       if(type == null){
           type = defaultType;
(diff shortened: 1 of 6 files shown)


cruise.umple/test/cruise/umple/compiler/2341_invalidAssociationRoleName.ump
```diff
@@ -0,0 +1,13 @@
+class X {
```
TimLethbridge: Tests should be minimal. All we need in this file are the lines 8, 9, 12 and 13

cruise.umple/src/en.error
```diff
@@ -118,8 +118,8 @@
 111: 4, "https://manual.umple.org/?WE1xxIdentifierInvalid.html", Interface name '{0}' should start with a capital letter ;
 112: 2, "https://manual.umple.org/?E112DuplicateConstantsinInterface.html", Interface '{0}' has duplicate constant name {1} ;
 
-# [TODO - issue 337] 120: 2, "https://manual.umple.org/?WE1xxIdentifierInvalid.html", Association role name '{0}' must be alphanumeric and start with a letter ;
-# [TODO - issue 337] 121: 4, "https://manual.umple.org/?WE1xxIdentifierInvalid.html", Association role name '{0}' should start with a lower-case letter ;
+120: 2, "https://manual.umple.org/?WE1xxIdentifierInvalid.html", Association role name '{0}' must be alphanumeric and start with a letter ;
+121: 4, "https://manual.umple.org/?WE1xxIdentifierInvalid.html", Association role name '{0}' should start with a lower-case letter ;
```
TimLethbridge, line 122: I see code that triggers error 120, but not code that triggers error 121

cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
```diff
@@ -4537,7 +4537,9 @@ private Boolean checkIsDistributed(UmpleInterface uInterface)
     AssociationEnd firstEnd = createPreliminaryAssociationEnd(myEndToken, defaultMyType);
     AssociationEnd secondEnd = createPreliminaryAssociationEnd(yourEndToken, defaultMyType);
     if(firstEnd == null || secondEnd == null){
-        setFailedPosition(associationToken.getPosition(), 8, name);
+        if(isAssociationClass){
```
TimLethbridge, line 4550: The rationale for this change is not explained. It seems to have nothing to do with the purpose of this PR.

cruise.umple/src/en.error
TimLethbridge, line 122: This error is not being checked for

cruise.umple/test/cruise/umple/compiler/2341_invalidAssociationRoleName.ump
TimLethbridge: Tests should me minimal. only include the code of lines 8-9 and 12-13

cruise.umple/test/cruise/umple/compiler/UmpleParserTest.java
```diff
@@ -1862,6 +1862,11 @@ public void TestReservedRoleNameError32() {
 	  assertFailedParse("009_ReflexiveReservedRoleNameError32.ump",32);
   }
 
+  @Test
```
TimLethbridge, line 1865: Add another test for message 121

cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
TimLethbridge, line 4550: I see you have a comment in the PR message explaining this. Add a comment before this I think mentioning that this is to catch association ends in AssociationClasses only ... but it is puzzling why it was not needed before. Perhaps in the comment state that without this check X happens, where X is the bad thing that you need to explain

TimLethbridge (changes requested):
Some changes are needed. Please see the comments

cruise.umple/test/cruise/umple/compiler/2341_invalidAssociationRoleName.ump
Ilyas-Ait-Ali: Updated. I reduced the test to the minimal association example needed to trigger the invalid role-name error.

cruise.umple/src/en.error
Ilyas-Ait-Ali: Addressed. I added handling for warning 121 and a separate regression test for an uppercase role name.

cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
Ilyas-Ait-Ali: Added a code comment explaining the rationale. Without the guard, a failed association-end validation would also trigger error 8, which creates a misleading cascading error unrelated to the root cause.

cruise.umple/src/en.error
Ilyas-Ait-Ali: Fixed. I added parser logic that emits warning 121 and added a regression test covering that case.

cruise.umple/test/cruise/umple/compiler/2341_invalidAssociationRoleName.ump
Ilyas-Ait-Ali: Updated. I reduced the test to the minimal association example needed to trigger the invalid role-name error.

cruise.umple/test/cruise/umple/compiler/UmpleParserTest.java
Ilyas-Ait-Ali: Added. There is now a separate parser test for warning 121 using an uppercase role name.

cruise.umple/src/class/UmpleInternalParser_CodeClass.ump
Ilyas-Ait-Ali: Added a code comment before the guard explaining that it limits error 8 to actual AssociationClass missing-end cases and prevents a misleading secondary error after prior validation has already failed.

TimLethbridge (approved):
The changes I requested have been made. I will update to bring in changes from master, then merge.

PR #2368, opened 2026-03-06
Fix gradle build issue #2358 and remove the plugin unnecessary

### LINKED ISSUE #2358: Gradle.build cannot run due to grgit:3.0.0 EOL
Running gradle.build directly will be fail due to the plugin `grgit:3.0.0` problem.

This version was end of life because it was only published on JCenter, which was shutdown. This version is not available in maven.

Although we can include third party repository, this may effect other plugins and bring supply chain risk.

We should upgrade to at least 4.x.x. May need to evaluate any breaking change and do necessary change, but the `grgit` has only few place using in the project, this wont be hard.

skyfackr:
This PR fixes #2358.

I tested both `grgit:4.1.1` and `grgit:5.3.3`, and neither introduced breaking changes. However, after reviewing how `grgit` is used here and comparing it with the Ant build, I concluded that removing `grgit` is the better long-term solution.

`grgit` is only used to generate the Umple version number, while the Ant build already does this via the `git` command. More importantly, `grgit` and Ant produce different version numbers for the same codebase. Since version numbers are part of the build artifact names, this inconsistency could lead to conflicts between Gradle and Ant outputs.

In addition, `grgit` is no longer actively supported by its author and may become unmaintained in the future:
`https://git.sr.ht/~ajoberstar/grgit`

The version string does not require any special semantics; it only needs to be consistent for everyone building from the same commit. Based on that, this PR removes `grgit` and replaces it with a small Gradle function that generates the version number in the same way as Ant, using the `git` command directly and without any plugin dependency.

After multiple tests and builds, the new approach works as expected and remains consistent with Ant. The only remaining failure is the `umpleOnline` build (like `pumple` did), which appears unrelated to version generation and is more likely caused by missing build artifacts during the build process.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
 }
 
 plugins {
-  id 'org.ajoberstar.grgit' version '3.0.0'
+  id 'org.ajoberstar.grgit' version '5.3.3'
   id 'nebula.ospackage' version '9.1.1'
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -10,8 +10,21 @@ buildscript {
 }
 
 plugins {
-  id 'org.ajoberstar.grgit' version '5.3.3'
   id 'nebula.ospackage' version '9.1.1'
+}
+
+String runGit(List<String> args, String fallback) {
+    // The Ant script uses the git command to get version info data. Use the same behavior here.
+    def output = new ByteArrayOutputStream()
+    def result = project.exec {
+        commandLine = ['git'] + args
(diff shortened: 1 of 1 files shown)


TimLethbridge (approved):
I am unable to test this personally since I have Gradle 9 and there are parts of the gradle build that require gradle 7 (still used by the Jenkins C)I. I will ask if you can do the Gradle tagged issues including new one #2363 

I am merging this though since it passed Jenkins

PR #2370, opened 2026-03-08
Fixes #1867 Add natural language descriptions to GV state diagram tooltips

### LINKED ISSUE #1867: Add natural language description in GV state machines

ElwinLiu:
## Summary

This PR adds automatically generated **natural language (plain-English) descriptions** to the tooltips of states and transitions in Graphviz state machine diagrams. When users hover over a state or transition in UmpleOnline, they now see descriptions in natural language.

A new **"Natural Language" toggle checkbox** is added to UmpleOnline's state diagram view options. It is **enabled by default**

image

This is a **scoped, non-breaking change** — it only adds new content to tooltips and a new UI toggle. All existing functionality remains intact.

### State tooltips

| Situation | Example tooltip text |
|---|---|
| **Initial state** | `This is the default initial state and is entered when the object is created.` |
| **Initial substate (nested)** | `This is the default initial state when entering state s2.s2a.` |
| **Incoming event transition** | `This state is entered when event e1 occurs while in state s1.` |
| **Incoming timer transition** | `This state is entered after 5 second(s) while in state s1 when guard [isReady()] is true.` |
| **Entry action present** | `On entry into this state an entry action is executed.` |
| **Do activity present** | `A do activity runs in a separate thread until completion or state exit.` |
| **Outgoing event transition** | `On e1, transitions to s2.` |
| **Outgoing timer transition** | `After 5 second(s) if guard [isReady()], transitions to s2.` |
| **Self-transition** | `On e1, loops back to self.` |
| **Exit action present** | `On exit from this state an exit action is executed.` |

### Transition tooltips

| Situation | Example tooltip text |
|---|---|
| **Event transition** | `When e1 occurs in state s1, transitions to s2.` |
| **Timer transition** | `After 5 second(s) in state s1, transitions to s2 if guard [isReady()].` |
| **Auto-transition** | `Automatically transitions from s1 to s2.` |
| **Self-transition** | `When e1 occurs in state s1, loops back to self.` |
| **Guarded transition** | `When e1 occurs in state s1, transitions to s2 if guard [x > 0].` |

### Deeply nested states

For deeply nested sm, the natural language uses full state paths.
image

### Toggle (show/hide)

- **Default**: Natural language descriptions are **shown** (`Page.showNaturalLanguage = true`)
- Users can **uncheck** the "Natural Language" checkbox
- Only applies to the **state machine diagram** 

## Test plan

- [x] Run `GvSdGeneratorTest` — all 5 new NL tests and all 27 existing tests should pass
- [x] Open UmpleOnline, create a state machine, and verify tooltips show natural language descriptions on hover
- [x] Uncheck the "Natural Language" checkbox and verify tooltips revert to technical-only content
- [x] Verify nested state machines display correct full state paths in NL text
- [x] Verify no regressions in other diagram types (class diagrams, etc.)

Fixes #1867

### DIFF
--- a/cruise.umple/src/generators/statemachineDiagramGenerator/Generator_CodeGvStateDiagram.ump
+++ b/cruise.umple/src/generators/statemachineDiagramGenerator/Generator_CodeGvStateDiagram.ump
@@ -27,6 +27,7 @@ class GvStateDiagramGenerator
 {
   Boolean hideActions = false;
   Boolean hideGuards = false;
+  Boolean hideNaturalLanguage = false;
   Boolean showTransitionLabels = false;
   Boolean showGuardLabels = false;
   internal String display_language = "";
@@ -71,6 +72,7 @@ digraph "<<=filename>>" {
     
     hideActions = hasSuboption("hideactions");
     hideGuards = hasSuboption("hideguards");
+    hideNaturalLanguage = hasSuboption("hidenaturallanguage");
     showTransitionLabels = hasSuboption("showtransitionlabels");
     showGuardLabels = hasSuboption("showguardlabels");
     display_language = getDisplayLanguage();
@@ -458,26 +460,9 @@ node [ratio="auto" shape = circle, fixedsize = true, width=.3];
            transitionLabel = "   t"+getObjectIdentity( root, t )+": ";
         }       
         
-        if(event.isAutoTransition()) 
+        if(!event.isAutoTransition())
         {
-          transitionLabel += "";
-        } 
-        else {
-          if (event.getIsTimer()) 
-          {
-            transitionLabel += "after("+event.getTimerInSeconds()+")";
-          }
-          else 
-          {
-            if(event.getArgs() == null || event.getArgs() == "") 
-            {
-              transitionLabel += event.getName();
-            }
-            else 
-            {
-              transitionLabel += event.getName()+"("+event.getArgs()+")";
-            }
-          }
+          transitionLabel += getTriggerDescription(event);
         }
         
         if (action == null || getAvailableActionCode(action) == "" || hideActions) 
@@ -508,7 +493,7 @@ node [ratio="auto" shape = circle, fixedsize = true, width=.3];
           else guardString = " [";
 
           //guardString += guard.getCondition(gen).replaceAll("\"","&quot;") + "]";
-	      guardString += guard.getExpression().replaceAll("\"","&quot;") + "]"+guardLabel;
+	      guardString += getEscapedGuardExpression(guard) + "]"+guardLabel;
         }
         
         String orig = getTransitionNameForState(t.getFromState(),uClass,true);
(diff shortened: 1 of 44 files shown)


cruise.umple/src/generators/statemachineDiagramGenerator/Generator_CodeGvStateDiagram.ump
```diff
@@ -632,39 +617,89 @@ node [ratio="auto" shape = circle, fixedsize = true, width=.3];
...
+          nlTransition = "Automatically loops back to " + tooltipFromState;
+        }
+        else
+        {
+          nlTransition = "Automatically transitions from " + tooltipFromState + " to " + tooltipToState;
+        }
+      }
+      else if (event.getIsTimer())
+      {
+        nlTransition = "After " + event.getTimerInSeconds() + " second(s) in state " + tooltipFromState;
```
TimLethbridge: The intention was that this is supposed to be millisecond(s) ... needs fixing wherever 'second(s)' appears here and in the relevant tests!

TimLethbridge (changes requested):
I marked only one comment that I think is important ('second(s)' should be changed to 'millisecond(s)') in relevant places. This is the only change to make please.

The only other thing is that It would be excellent (for the future) would be if the comments preceding states and transitions could be output as they are for classes and attributes. But I looked at the code and the metamodel parser doesn't even deal with them yet, so that needs to be set up as a separate issue. Even comments on associations are not yet output in class diagrams. The following is a sample of code:

```
// This is the comment for class X
// It is indeed output to the natural language
class X {
  // This is the comment for attribute a1
  // it is correctly output to the NL
  a1;

  sm1 {
    // This is the comment for state s1
    // not yet output to the NL ... there is not
    // even a comment association in sm
    s1 {
      // This is the comment for the e1 trans to s2
      // Not output to the NL
      e1 -> s2;

      // this is the comment for the next trans
      // Not output to the NL
      
      e2 -> / {doSomething();} s2;
    }
    // This is the comment for state s2
    s2 {
      
    }
  }
}

class Y {
  // Comment for association from Y to X
  // Not output to the NL for the association
  1 -- * X;
}
```

cruise.umple/src/generators/statemachineDiagramGenerator/Generator_CodeGvStateDiagram.ump
ElwinLiu: Converted to millisecond(s) with commit: 0f07c212f8daa4e83a53e288de9783d8281fdb2a

Since the original number is considered as second(s), to convert to the intended millisecond(s), I simply multiplied it by 1000. Please let me know if I got it wrong.

ElwinLiu:
> I marked only one comment that I think is important ('second(s)' should be changed to 'millisecond(s)') in relevant places. This is the only change to make please.
> 
> The only other thing is that It would be excellent (for the future) would be if the comments preceding states and transitions could be output as they are for classes and attributes. But I looked at the code and the metamodel parser doesn't even deal with them yet, so that needs to be set up as a separate issue. Even comments on associations are not yet output in class diagrams. The following is a sample of code:
> 
> ```
> // This is the comment for class X
> // It is indeed output to the natural language
> class X {
>   // This is the comment for attribute a1
>   // it is correctly output to the NL
>   a1;
> 
>   sm1 {
>     // This is the comment for state s1
>     // not yet output to the NL ... there is not
>     // even a comment association in sm
>     s1 {
>       // This is the comment for the e1 trans to s2
>       // Not output to the NL
>       e1 -> s2;
> 
>       // this is the comment for the next trans
>       // Not output to the NL
>       
>       e2 -> / {doSomething();} s2;
>     }
>     // This is the comment for state s2
>     s2 {
>       
>     }
>   }
> }
> 
> class Y {
>   // Comment for association from Y to X
>   // Not output to the NL for the association
>   1 -- * X;
> }
> ```

Just drafted an issue for this: #2374, and a PR that's in draft for now: #2375. Since it's touching the metamodel and the parser, I'd like to spend more time testing it.

TimLethbridge:
@elwinLiu thanks! I will merge this one in the morning (after CI passes). Then your #2375 can build on this one.

PR #2384, opened 2026-03-13
fix #2383 set up main page jump

### LINKED ISSUE #2383: UmpleOnline docker shows 404 without /umple.php
Because there is no main page jump when no path is given

skyfackr:
set up a 302 transfer to `/umple.php` when accessing without any path. 

This will not affect 404 page when accessing non-exist path


skyfackr:
website test failed. need extra testing

TimLethbridge:
Please confirm that this works if the umpleonline in the docker image is invoked with arguments ?example=2DShapes ... when you confirm this works, then take this out of draft mode and I will merge

PR #2389, opened 2026-03-16
Fix state diagram context menu not appearing on right-click

Samoumine:
The popup menu failed because drawStateMenu() tried to extract the state identifier using indexOf/substring logic expecting stateClicked(&quot;...&quot;), while the actual SVG uses normal quotes stateClicked("..."). This caused the identifier parsing to fail and elemText to be undefined, preventing the menu from being created.

Replaced the brittle extraction with a regex that correctly captures the stateClicked("...") identifier and split it safely. Also restored the correct class/state-machine casing using the xlink:title attribute so selectStateInClassCM6() can locate the state in the source model.


TimLethbridge:
@samoumine 

You have re-used branch issue1958-fix-clean for this PR. That is an absolute no-no! Once a PR is merged that branch should never be touched again

I remember notiing several weeks ago that there was a problem right-clicking on a state, but in master there is no problem now, so it is puzzling what this is fixing. It is not issue 1958 as that does not mention this problem. Maybe you already fixed this in PR #2342 that I merged from this branch previously?

Do you notice some specific remaining states in the master branch where right-click is not working so I can test? I tried several examples including 'Auction'.

I have noticed a totally separate issue, which is that in the Auction example, some of the transitions are are not, it seems getting any menu at all, and some are not even selecting the correct issue. That should likely be raised as a new issue.

Other general comments:L There is no issue number in your PR title (which always in include one). The branch name says issue1958-fix-clean, but issue 1958 is closed and 'fix-clean' was never informative ... your branch name should include 2 words that help people remember what the branch is about such as state-rightclick. I am not asking you to change the branch name now, but please remember for the future.

I am marking this back to 'DRAFT' since this has used up a lot of my time, and I need invormative answers to the above.

TimLethbridge:
Closing as replaced by #2433

Issue #2377, opened 2026-03-10
Add support for user stories as a requirements type


TimLethbridge:
Add a userStory tag in the req statement.

Purpose: user stories are a common approach to requirements, Being able to tag these will allow both humans and AIs to better understand them. See also #2378 which will be very similar but for use cases.

This will suggest to any requirements analysis tool that the requirement is a user story.

Within the body of the requirement the text could be free-text, or could also have any of the following:

who { arbitrary text)
  defines actor / role / who

when { arbitrary text }
  defines (condition / when /  where),

what ( arbitrary text ... core of the story }
  defines  (goal  / what / capability)

why { arbitrary text }
  defines (benefit / why)

Ilyas-Ait-Ali:
I’m working on this issue incrementally across multiple PRs.

### Breakdown

- [x] **Grammar support for `userStory`**
  - Extend the requirements grammar so the `userStory` requirement type is recognized syntactically
  - Preserve existing requirement syntax
  - Support structured `userStory` body:
    - `who { ... }`
    - `when { ... }`
    - `what { ... }`
    - `why { ... }`
  - Add parser tests for the new syntax

- [x] **Metamodel updates**
  - Update the `Requirement` metamodel to store structured user story elements

- [x] **Parsing and validation**
  - Update the internal parser to extract and store structured fields (`who`, `when`, `what`, `why`)
  - Ensure backward compatibility with existing requirements

- [x] **Output updates**
  - Update requirement output/documentation generation to include structured user story data


- [x] **Grammar alias support**
  - Accept userstory as a valid alternative to userStory
  - Add grammar coverage tests for both spellings

- [x] **User manual documentation**
  - Add a user manual page section with realistic userStory examples
  - Include both plain-text and structured forms
  - Show examples using:
    - `who { ... }`
    - `when { ... }`
    - `what { ... }`
    - `why { ... }`

PR #2390, opened 2026-03-18
Issue #2377: Extend Requirement metamodel to support structured user stories

Ilyas-Ait-Ali:
This PR implements the metamodel step of Issue #2377.

### Scope
- Extend the `Requirement` class to support structured user story elements:
  - `who`
  - `when`
  - `what`
  - `why`
- Update the deep-copy constructor to include the new fields
- Preserve backward compatibility by adding a 3-argument constructor used by existing code
- Add unit tests to verify:
  - structured fields are stored correctly
  - deep copies preserve structured data

### Rationale
The grammar now supports `userStory` requirements with optional structured sections. This PR introduces the necessary data model changes so these elements can be stored internally.

### Notes
- This PR does not yet populate these fields during parsing
- This PR does not modify output generation

These will be handled in subsequent PRs.

### DIFF
--- a/cruise.umple/src/Umple_Code.ump
+++ b/cruise.umple/src/Umple_Code.ump
@@ -923,12 +923,20 @@ class Comment
 /*
  */
 class Requirement{
+  public Requirement(String aIdentifier, String aStatement, String aLanguage)
+  {
+    this(aIdentifier, aStatement, null, null, null, null, aLanguage);
+  }
 
   // deep copy constructor
   public Requirement(Requirement aReq)
   {
     identifier = aReq.getIdentifier();
     statement = aReq.getStatement();
+    who = aReq.getWho();
+    when = aReq.getWhen();
+    what = aReq.getWhat();
+    why = aReq.getWhy();
     language = aReq.getLanguage();
   }
   public static List<Comment> convertToComment(List<ReqImplementation> reqSelected,UmpleModel aModel){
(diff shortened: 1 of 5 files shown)


cruise.umple/src/Umple_Code.ump
```diff
@@ -923,12 +923,20 @@ class Comment
 /*
  */
 class Requirement{
+  public Requirement(String aIdentifier, String aStatement, String aLanguage)
```
TimLethbridge: Umple already generates constructors with all the attributes so this should not be needed. You should be able to generate the Javadoc from your branch to see the API it creates. I am not sure why the deep copy constructor is needed, below; but I understand extending that one. However creating an entirely new constructor here I think will be duplicative.

TimLethbridge:
@Ilyas-Ait-Ali Please respond to the above comment ... if you can justify it I can merge but I think the generated constructor should be enough (the generated one would also create an empty association to tokens as it has that association)

Ilyas-Ait-Ali:
Thanks, that makes sense. I had added the 3-argument constructor only to preserve the existing call sites after extending Requirement, but I agree that using the generated constructor is the better approach here. I’ve now removed the custom constructor, kept the deep copy constructor updated, and changed the call sites to use the generated constructor instead. I reran bumple and the build passes successfully.

PR #2411, opened 2026-03-26
Issue #2377: Add userStory parsing support (who/when/what/why)

### LINKED ISSUE #2377: Add support for user stories as a requirements type

Ilyas-Ait-Ali:
Fixes #2377

## Summary
This PR enhances the requirements parsing to support structured `userStory` definitions.

- Extracts structured sections: `who`, `when`, `what`, and `why`
- Stores extracted values in the `Requirement` model
- Cleans the `statement` field by removing structured blocks
- Preserves existing behavior for non-`userStory` requirements
- Ensures consistent parsing across both trait and class parser implementations

## Implementation
- Updated:
  - `UmpleInternalParser_CodeTrait.ump`
  - `UmpleInternalParser_CodeClass.ump`
- Added parsing logic to:
  - detect structured `userStory` blocks
  - extract section values
  - retain only non-structured content in `statement`

## Behavior
- Fully structured user stories:
  - `statement` is empty
  - `who/when/what/why` are populated

- Partially structured user stories:
  - only provided fields are populated
  - missing fields remain `null`

- Plain user stories:
  - remain stored entirely in `statement`

- Duplicate requirement IDs:
  - existing merge behavior is preserved
  - structured fields are updated when present

## Tests
Updated tests for:
- Fully structured user story
- Partially structured user story

## Notes
This PR focuses only on parsing and model behavior.  
Output/rendering updates will be handled in a separate PR.


TimLethbridge (changes requested):
Please redo without the massive changes that just affect indentation!

PR #2423, opened 2026-03-29
Fixes #2054 - Improved requirements generator formatting, sorting, and suboptions

### LINKED ISSUE #2054: Fix and improve the default requirements generator
This is a sub-issue of https://github.com/umple/umple/issues/1865 which is the main issue for the requirements feature.

Do the following to the Generator_CodePlainRequirementsDoc.ump

You can make these changes in multiple smaller pull requests.

Some of them may be done. When done, check the box.

- [x] By default requirements should all be outputted (not just implemented ones). But only the ones in selected mixsets if requirements are in any mixsets. This may be done. Needs verification
- [x] Currently the implemented requirements are being output multiple times; they should only be output once.
- [x] Requirements not implemented should be followed by 'NOT IMPLEMENTED'.
- [x] Requirements that are implemented should be followed by. Implemented by ...' with the name of the class(es) (if at class level) or name of the attributes (and in which class) or states (and in which state machine and class etc.). Also include the file and lines of code. It would be good to use a style for this output so that it can be more easily distinguished from the requirements themselves.
- [x] Requirements should be sorted by default in alphabetical order in the output.
- [ ] Add generator suboptions (using -s) to limit output to only implemented or not-implemented requirements ... and to hide the implemented and not implemented statements. This may be done. Needs verification.

ShmRayan:
## Summary

Addresses issue #2054 by improving the PlainRequirementsDoc generator.

**Depends on #2430 being merged first** (adds association regression tests).

## Changes

**Bug fixes:**
- Fix string equality: changed `==` to `.equals()` for identifier comparisons so implementations are correctly matched to requirements
- Fix `reqSortStat` comparator: include identifier as tie-breaker so sort is stable when two statements are identical

**New features:**
- Add deduplication: use `HashSet` to prevent the same implementation line from appearing twice in the output
- Add `reqHideImpl` suboption: when set, show only requirements that are NOT yet implemented
- Add `reqHideStatements` suboption: when set, show only the requirement identifier without the statement text

## Tests

- `ReqHideImpl`: regression test for the `reqHideImpl` suboption
- `ReqHideStatements`: regression test for the `reqHideStatements` suboption
- `ReqSortStat`: regression test for the stable `reqSortStat` sort order

Association output is covered by the tests in #2430.


TimLethbridge:
This PR has major problems as it deletes critical functionality. That functionality must be put back and relevant tests added to ensure there are no future regressions.

The following example code shows that good information about associations is missing in the simple requirements output:

```
req R1 {blah}

class X {
  a;
}

class Y {
  b;
  implementsReq R1;
  0..1 -> * X;
}
```

For this the current master says:

```
R1: blah

  IMPLEMENTED BY:

    association:  0..1 y -> * X xs;  in class Y model.ump line 10 ;
```

Your version currently says:

```
R1: blah

  Implemented by:

    - association in class Y in model.ump (line 10)
 ```

What you have done is eliminated the representation of the association. That was carefully written code so that people can understand the actual association when they read the report. It is only short, so should not have been deleted. Just stating that it is an association is much less informative.

To rectify:

 - [x] Read the code in master and understand the details being presented in the requirements document and how they are being generated. It may not just be associations ... read the code carefully.
 - [x] Add tests that test the missing cases. I suggest a different PR in a different branch that will add more test cases to test associations and anything else that is was not tested before. That PR can then be merged before this one.
 - [x] Update this PR to maintain that information about the associations (generating good the missing information and updating the tests mentioned in point 2 for new format)

I am putting this back into draft until this PR is thoroughly fixed.

PR #2429, opened 2026-03-31
Progress on #2369: Add Gradle 8 support

MysteryBlokHed:
Related issue: #2369
(Doesn't close the issue)

Adds support for building Umple with Gradle 8. Gradle 8 support can be achieved using a very small diff, whereas Gradle 9 support requires a bit more work. I figured it would still be valuable for Umple to be buildable using a non-EOL version of Gradle until Gradle 9 support is added, especially since the changes are so minimal.

The project also still builds properly with Gradle 7.

---

> PRs should always be made from a branch you have created on umple/umple, not from a fork: This ensures that a Docker image of your PR can be automatically generated for manual testing.

I don't have write access to the repository, so I had to use a fork. If this is a problem, the branch could be copied to the upstream repo and then used as the head of this PR instead.

Issue #2369, opened 2026-03-07
Gradle build scripts need updating to be compatible with Gradle 8 and 9 (but ideally still running in 7)


TimLethbridge:
When running ./gradlew build with the most recent version of Gradle 9.4.0, there are numerous issues

The code below shows at the end that there are deprecations that need fixing such as SourceDirectorySet.setOutputDir, AbstractArchiveTask.version property,  AbstractArchiveTask.version property, and JavaExec.main

Also the trace below is showing 'dupicated java classes'. This might be as a result of having done an ant build (bumple) first, or it could be something to do with directory structure. The gradle build needs fixing so the ant build and gradle build both work and do the same thing.

Please also see other issues tagged gradle as there may be overlaps

```
Gradle 9.4.0
------------------------------------------------------------

Build time:    2026-03-04 10:36:00 UTC
Revision:      b631911858264c0b6e4d6603d677ff5218766cee

Kotlin:        2.3.0
Groovy:        4.0.29
Ant:           Apache Ant(TM) version 1.10.15 compiled on August 25 2024
Launcher JVM:  17.0.12 (Oracle Corporation 17.0.12+8-LTS-286)
Daemon JVM:    /Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home (no Daemon JVM specified, using current Java home)
OS:            Mac OS X 26.3.1 aarch64

[MacBookPro:~/umple] tcl% ./gradlew build --warning-mode all

/gradlew build --warning-mode all

> Configure project :
The SourceDirectorySet.setOutputDir(File) method has been deprecated. This is scheduled to be removed in Gradle 8.0. Please use the destinationDirectory property instead. See https://docs.gradle.org/7.3/dsl/org.gradle.api.file.SourceDirectorySet.html#org.gradle.api.file.SourceDirectorySet:destinationDirectory for more details.
        at build_9v440bwisy8jt4wkxrmgpdqw8$_run_closure21$_closure129.doCall(/Users/tcl/unix/umple/build.gradle:379)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
The AbstractArchiveTask.version property has been deprecated. This is scheduled to be removed in Gradle 8.0. Please use the archiveVersion property instead. See https://docs.gradle.org/7.3/dsl/org.gradle.api.tasks.bundling.AbstractArchiveTask.html#org.gradle.api.tasks.bundling.AbstractArchiveTask:version for more details.
[...]

PR #2431, opened 2026-04-01
Progress on #2369: Add Gradle 8 support (V2)

TimLethbridge:
Clone of PR #2429 from @MysteryBlokHed since he did not have write access

Related issue: #2369
(Doesn't close the issue)

Adds support for building Umple with Gradle 8. Gradle 8 support can be achieved using a very small diff, whereas Gradle 9 support requires a bit more work. I figured it would still be valuable for Umple to be buildable using a non-EOL version of Gradle until Gradle 9 support is added, especially since the changes are so minimal.

The project also still builds properly with Gradle 7.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
 }
 
 plugins {
-  id 'com.netflix.nebula.ospackage' version '12.3.0'
+  id 'com.netflix.nebula.ospackage' version '11.5.0'
 }
 
 String runGit(List<String> args, String fallback) {
--- a/build.gradle
+++ b/build.gradle
@@ -426,7 +426,7 @@ tasks.named("compileTestJava") {
 }
 
 // ===============
-//  Jar packaging
+//  Jar packaging 
 // ===============
 
 // Grammar and error files also need to be included in the Umple jar
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
 }
 
 plugins {
-  id 'nebula.ospackage' version '9.1.1'
+  id 'com.netflix.nebula.ospackage' version '12.3.0'
 }
 
 String runGit(List<String> args, String fallback) {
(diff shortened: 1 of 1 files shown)


TimLethbridge:
First try resulted in a concurrent modification exception, so will tweak and try to get Jenkins to process it again

build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
 }
 
 plugins {
-  id 'nebula.ospackage' version '9.1.1'
+  id 'com.netflix.nebula.ospackage' version '12.3.0'
```
MysteryBlokHed: I can also reproduce the concurrent modification exception locally. I think that when I tested the PR, I used the latest Gradle 7 and 8 instead of using the Gradle 7 via the project's wrapper. For whatever reason, that version built just fine.

From my testing, it seems to be caused by this plugin. I found v11.5.0 to be the newest version that builds properly with the Gradle 7 wrapper.

```suggestion
  id 'com.netflix.nebula.ospackage' version '11.5.0'
```

TimLethbridge:
So Jenkins still failing. I am going to see if I can update it a little (although racing to do too many things)!

TimLethbridge:
This PR will have to wait until Jenkins is updated. Which in turn will have to wait until we have a period of quiescence in incoming PRs ... I can't have Jenkins downtime (or risk over-comuming my time) while students are rapidly rushing in contributions as the semester comes to an end. In the latter part of April I will have time to update Jenkins and all its plugins, and will get back to this PR at that time!

MysteryBlokHed:
> This PR will have to wait until Jenkins is updated. Which in turn will have to wait until we have a period of quiescence in incoming PRs

Yeah, that's understandable.

I wonder if it would be worth just skipping this change and going directly to Gradle 9, whenever a contributor has time to work on that. I had originally opened my PR because it seemed like Gradle 8 compatibility could be accomplished with minimal effort, but it might be more effort than it's worth just to get to a version of Gradle that will also likely go end-of-life in the near future.

TimLethbridge:
Closing as superceded by #2517

PR #2433, opened 2026-04-02
Fix local state popup menu in state diagrams

Samoumine:
## Summary
This PR fixes the state popup menu parsing in UmpleOnline for local setups.

The issue was that `Action.drawStateMenu()` assumed a specific `stateClicked(...)` encoding in the SVG/outerHTML, which caused the popup menu code to fail locally with:

`Uncaught TypeError: can't access property "split", t[2] is undefined`

## Changes
- Made state popup menu parsing more robust in `Action.drawStateMenu`
- Added safe ancestor traversal before reading the graph node
- Added a guard when the expected state identifier cannot be extracted
- Preserved the existing state menu behavior once the identifier is found

## Testing
- Rebuilt UmpleOnline locally
- Opened the local site
- Right-clicked or Double-Click states in state diagrams
- Confirmed the popup menu no longer crashes locally
- Confirmed the official web behavior was preserved

### DIFF
--- a/umpleonline/scripts/umple_action.js
+++ b/umpleonline/scripts/umple_action.js
@@ -1653,27 +1653,16 @@ Action.drawStateMenu = function(){
     targ = targ.parentNode;
   }
   //grabs state name
-  var match = targ.outerHTML.match(/stateClicked\("([^"]+)"\)/);
-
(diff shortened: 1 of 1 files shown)


TimLethbridge:
@Samoumine This PR does NOT work. In fact it seems to break the very functionality it  purports to be fixing! Moving to Draft

Issue #2378, opened 2026-03-10
Add support for use cases as a requirements type


TimLethbridge:
Add a useCase reqLanguage in the req statement, so 

This will suggest to any requirements analysis tool that the requirement is a use case.

example

```
req addItemToBill useCase {
   who {salesPerson};
   userStep 1 {select product}
   systemResponse 1 {display price}
}
```

Purpose: use cases are a common approach to requirements, Being able to tag these will allow both humans and AIs to better understand them. See also #2377 which is very similar but for user stories. To avoid code duplication, the processing of use cases and user stories should likely common code, just extended for use cases. Use case diagrams could also be drawn from this

Within the body of the requirement the text could be free-text, or could also have any of the following:

The who/when/what/why from user stories could be the same. See issue #2377 

Use cases would support the following, where ids can be numbers or alphanumberic (the following is only a rough draft)

```
userStep [[useCaseId]] ([ [[useCaseCondition]] ])? { [[arbitrary text]] };
systemResponse [[useCaseId]] ([ [[useCaseCondition]] ])? { [[arbitrary text]] };
include usecase;
isA usecase
```

Ilyas-Ait-Ali:
I’m working on this issue incrementally across multiple PRs.

### Breakdown

* [x] **Grammar support for `useCase`**

  * Extend the requirements grammar so the `useCase` requirement type is recognized syntactically
  * Preserve existing requirement syntax
  * Support structured `useCase` body:

    * `who { ... }`
    * `when { ... }`
    * `what { ... }`
    * `why { ... }`
    * `userStep <id> { ... }`
    * `systemResponse <id> { ... }`
  * Add parser tests for the new syntax
  * Add lowercase alias grammar support for `usecase`

* [x] **Metamodel updates**

  * Update the `Requirement` metamodel to store structured use case elements
  * Add support for storing:

    * user steps
    * system responses
    * step identifiers

* [x] **Parsing and validation**

  * Update the internal parser to extract and store structured fields
  * Normalize `usecase` → `useCase`
  * Parse step sequences for:

    * `userStep`
    * `systemResponse`
  * Ensure backward compatibility with existing requirements

* [x] **Output updates**

  * Update requirement output/documentation generation to include structured use case data
  * Preserve step ordering in rendered output

* [ ] **Extended use case syntax**

  * Add support for optional future use case extensions:

    * `include usecase`
    * `isA usecase`
    * optional step conditions
  * Add grammar and parser coverage tests

[...]

PR #2440, opened 2026-04-04
Issue #2378: Extend Requirement metamodel to support structured use cases

Ilyas-Ait-Ali:
This PR implements the metamodel step of Issue #2378.

## Scope

Extend the `Requirement` class to support structured use case elements.

This PR introduces a `UseCaseStep` model element and associates it with `Requirement` so ordered use case steps can be stored internally.

Supported structured step storage:

* `userStep`
* `systemResponse`
* step identifiers
* step ordering

Additional changes:

* Update the deep-copy constructor to preserve associated use case steps
* Preserve backward compatibility with existing `Requirement` construction and formatting flows
* Reuse existing structured requirement fields already introduced for `userStory`:

  * `who`
  * `when`
  * `what`
  * `why`

## Tests

Add unit tests to verify:

* structured use case steps are stored correctly
* step order is preserved
* deep copies preserve structured use case step data

## Rationale

The grammar now supports `useCase` requirements with structured step syntax. This PR introduces the necessary metamodel changes so these elements can be stored internally before parser extraction and output generation are added.

## Notes

* This PR does not yet populate these fields during parsing
* This PR does not yet modify output generation
* These will be handled in subsequent PRs

### DIFF
--- a/cruise.umple/src/Umple.ump
+++ b/cruise.umple/src/Umple.ump
@@ -1327,13 +1327,19 @@ class Requirement{
   * -> * Token reqToken;
 
   // Ordered structured use case steps such as userStep/systemResponse
-  * -- * UseCaseStep useCaseStep;
+  1 -- * UseCaseStep;
 }
 
 class UseCaseStep
 {
+  enum UseCaseStepType
+  {
+    userStep,
+    systemResponse
+  }
+
   id;
-  stepType;
+  UseCaseStepType stepType;
   content;
 }
 
--- a/cruise.umple/src/Umple.ump
+++ b/cruise.umple/src/Umple.ump
@@ -1325,6 +1325,16 @@ class Requirement{
   // Used when reporting errors regarding missing implementations
   // and when reporting where a requirement is located in the code
   * -> * Token reqToken;
+
+  // Ordered structured use case steps such as userStep/systemResponse
+  * -- * UseCaseStep useCaseStep;
+}
+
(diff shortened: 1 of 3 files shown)


cruise.umple/src/Umple.ump
```diff
@@ -1325,6 +1325,16 @@ class Requirement{
   // Used when reporting errors regarding missing implementations
   // and when reporting where a requirement is located in the code
   * -> * Token reqToken;
+
+  // Ordered structured use case steps such as userStep/systemResponse
+  * -- * UseCaseStep useCaseStep;
```
TimLethbridge:
```suggestion
  1 -- * UseCaseStep;
```

TimLethbridge: It seems to me that a UseCaseStep should only belong to one use case. Also it would be best practice not to add a role name. Without the role name you will get a nice pluralized API getUseCaseSteps, addUseCaseStep and so on (this will force you to modify code elsewhere)

cruise.umple/src/Umple.ump
```diff
@@ -1325,6 +1325,16 @@ class Requirement{
...
+}
+
+class UseCaseStep
+{
+  id;
+  stepType;
```
TimLethbridge: I think you should add an enum with values userStep and systemResponse and use that type for this

TimLethbridge (changes requested):

PR #2451, opened 2026-04-07
[Draft] Attempt to propagate full file paths in parser and error hand…

Fadiedy:
**Summary**

This PR attempts to propagate full file paths instead of file names in error reporting.

**What I changed**

* Updated parser (`TextParser`, `UmpleInternalParser`) to store absolute file paths
* Updated some position and utility handling to use full paths

**Current behaviour**

* Full paths are now used internally in some parts
* However, error messages still display only the file name (e.g., `test.ump`)

**Notes**

This PR is a draft for feedback. The issue appears to remain in the error message formatting layer.


Fadiedy:
Digging deeper on this bug when we are back in September.

PR #2457, opened 2026-04-09
[ci skip]Update agents.md for agent instruction on using bumple and file operation tools

skyfackr:
This PR updates `AGENTS.md` to make agent workflow rules more explicit.

- It adds a rule requiring file reads/writes/searches/edits to go through the system-provided tools instead of being bypassed with shell or scripting commands.
- It clarifies that `tumple` should only be run after `qbumple` or `qfbumple` when any changes may have happened since the last compile (or when that state is unknown).
- It documents that `bumple` must be given at least a 15-minute timeout.
- For testing that does not require the full pre-PR validation run, the guidance now prefers `qbumple`/`qfbumple` + `tumple` whenever that is sufficient.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,12 +14,25 @@ Umple uses Ant (`build/` directory). Use the dev-tool shortcuts in `dev-tools/`:
 
 See `dev-tools/README.md` for the full list.
 
-- When running `bumple`, use at least 15 minutes for the tool execution timeout(This is not a param of `bumple`. This is a param of tool execution, and param name may be variable based on platform).
-- For testing purposes other than the full pre-PR validation run, prefer the combination of `qbumple` or `qfbumple` followed by `tumple` whenever that combination is sufficient to achieve the goal.
+### Compile and Test Guidelines
+
+#### Builds
+- For build and test tasks, agents MUST prefer scripts under `dev-tools/` over direct `Ant` or `Gradle` commands whenever a script covers the task.
+- During development, agents SHOULD use the narrowest `dev-tools/` script that covers the change.
+  - The scripts listed here are examples, not an exhaustive list.
+  - If the right script is not obvious, agents MUST check `dev-tools/README.md` before choosing a compile or package step, and MUST NOT default to `bumple` first.
+  - Typical cases: `qbumple` for changes limited to Java in the main `umple.jar`; `uminify` for JS-only changes; `bumple` only when a full build with tests is required.
+- See `dev-tools/README.md` for the full script list. Additional guidance is available at `https://github.com/umple/umple/wiki/CheatSheet` and `https://github.com/umple/umple/wiki/DevelopmentSetUp`.
+
+#### Tests
+- Before running `tumple`, if any file changed after the last compile step, agents MUST run the appropriate compile step first, even for test-only changes.
+- If the correct compile step is unclear, agents MUST determine it from `dev-tools/README.md`, local context, and recent changes; if still unclear, they MUST ask the user rather than default to `bumple`.
+- When running `bumple`, set the execution-tool timeout to at least 15 minutes. `bumple` already includes `tumple`; do not run it again. Except for full-repo testing or pre-PR validation, prefer a sufficient partial compile followed by `tumple`.
 
 ## Commits and PRs
 
 - Commit format: `type(scope): summary` with issue references (e.g. `fix(parser): handle nested state machines, Fixes #1234`)
+  - Common types: `feat`, `fix`, `docs`, `refactor`, `test`, `chore`
 - Branch naming: `issue#<number>_<short-description>`
 - When making a PR, read `.github/PULL_REQUEST_TEMPLATE.md` for detailed guidelines on PR titles, descriptions, and content requirements.
 - When making an issue, read `.github/ISSUE_TEMPLATE.md` for detailed guidelines.
@@ -30,8 +43,6 @@ See `dev-tools/README.md` for the full list.
 - Run `bumple` before opening a PR
 - Show diff before committing
 - Update user manual in `build/reference/` when changing Umple syntax or semantics
-- For file operations such as reading, writing, creating, deleting, searching, and editing, use the system-provided tools such as `read`, `grep`, `search`, `edit`, and `apply_patch`. The tools available may be different based on the model, platform, and agent application. Unless a developer explicitly approves or requires it, do not use general script execution tools including `bash`, `pwsh`, `python`, or similar mechanisms to bypass those tools for file operations. If a provided file tool rejects a request, follow its error information to determine the next step. If the rejection is clearly caused by a system mechanism such as permissions, do not try to bypass it with script execution tools; instead, reassess whether the action is compliant and necessary, or ask the user for a decision.
-- Before running `tumple`, if any file has been modified since the last compile step such as `qfbumple`, `qbumple`, or `bumple`, run `qfbumple` or `qbumple` first, regardless of whether the changes are in project code or tests. If the agent does not know when the last compile happened, or whether any files changed after it, default to running `qfbumple` or `qbumple` before `tumple`.
 
 ## Post-Work
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,7 +14,7 @@ Umple uses Ant (`build/` directory). Use the dev-tool shortcuts in `dev-tools/`:
 
 See `dev-tools/README.md` for the full list.
 
-- When running `bumple`, use a timeout of at least 15 minutes. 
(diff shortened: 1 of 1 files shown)


ElwinLiu:
Hi @skyfackr, thank you for this PR. I can see that you're trying to optimize the harness's workflow's efficiency. 

First of all. we expect AGENTS.md to be as high level and workflow agnostic as possible, so that it benefits all developers' workflow. 
Prompts like `bumple` timeout is a very good example that should be added in to AGENTS.md.

However, others like:
- tool call instructions, it's codex specific problem, other harness might not have similar tool set, e.g. apply_patch is a codex specific tool. Also, codex is the only harness that abuse Python script. So a better practice should be adding instruction on your machine, but not global AGENTS.md. You can achieve this by editing it locally, but exclude the AGENTS.md in .git/info/exclude; or put it in ~/.codex/AGENTS.md. Either way, they won't affect remote AGENTS.md
- Command order: again, different workflows have different order preference, some students focus solely on JS files, they only need `uminify`, some might touch compiler... So they also look personal workflows to me.

Overall, the `bumple` timeout is a beneficial universal change. The others are more suited to personal workflows and shouldn't be merged into origin/master. Please let me know if you have a different opinion.

skyfackr:
> Hi @skyfackr, thank you for this PR. I can see that you're trying to optimize the harness's workflow's efficiency.
> 
> First of all. we expect AGENTS.md to be as high level and workflow agnostic as possible, so that it benefits all developers' workflow. Prompts like `bumple` timeout is a very good example that should be added in to AGENTS.md.
> 
> However, others like:
> 
> * tool call instructions, it's codex specific problem, other harness might not have similar tool set, e.g. apply_patch is a codex specific tool. Also, codex is the only harness that abuse Python script. So a better practice should be adding instruction on your machine, but not global AGENTS.md. You can achieve this by editing it locally, but exclude the AGENTS.md in .git/info/exclude; or put it in ~/.codex/AGENTS.md. Either way, they won't affect remote AGENTS.md
> * Command order: again, different workflows have different order preference, some students focus solely on JS files, they only need `uminify`, some might touch compiler... So they also look personal workflows to me.
> 
> Overall, the `bumple` timeout is a beneficial universal change. The others are more suited to personal workflows and shouldn't be merged into origin/master. Please let me know if you have a different opinion.

Hello,

Thanks for feedback. I can improve the workflow order instructions.

But for the tool call instructions: I use all of codex, claude, and copilot. I found such issue among all three agent apps, that the agent tries to use bash or pwsh to bypass the file modify tracking and permissions. I believe this is not a codex specific issue but a general issue for AI because the AI model will keep focus on "solve the goal" during their internal generation. This is a general instruction in my own agent global instruction. But since the umple is too big, I think it is worth to add again in the repo's instruction so other devs can get benefit

ElwinLiu:
> Hello,
> 
> Thanks for feedback. I can improve the workflow order instructions.
> 
> But for the tool call instructions: I use all of codex, claude, and copilot. I found such issue that the agent tries to use bash or pwsh to bypass the file modify tracking and permissions. I believe this is not a codex specific issue but a general issue for AI because the AI model will keep focus on "solve the goal" during their internal generation.

Personally, I haven't experienced AI modifying files through bash. Which models are you using? 

Tools prompting is already injected into the system prompt, which has the highest priority/weight, LLM usually wouldn't use bash to edit unless it's necessary. Here I quote Claude Code's system prompt:

```
Executes a given bash command and returns its output.

The working directory persists between commands, but shell state does not. The shell environment is initialized from the user's profile (bash or zsh).

IMPORTANT: Avoid using this tool to run `find`, `grep`, `cat`, `head`, `tail`, `sed`, `awk`, or `echo` commands, unless explicitly instructed or after you have verified that a dedicated tool cannot accomplish your task. Instead, use the appropriate dedicated tool as this will provide a much better experience for the user:

 - File search: Use Glob (NOT find or ls)
 - Content search: Use Grep (NOT grep or rg)
 - Read files: Use Read (NOT cat/head/tail)
 - Edit files: Use Edit (NOT sed/awk)
 - Write files: Use Write (NOT echo >/cat <<EOF)
 - Communication: Output text directly (NOT echo/printf)
While the Bash tool can do similar things, it’s better to use the built-in tools as they provide a better user experience and make it easier to review tool calls and give permission.
```

skyfackr:
> > Hello,
> > Thanks for feedback. I can improve the workflow order instructions.
> > But for the tool call instructions: I use all of codex, claude, and copilot. I found such issue that the agent tries to use bash or pwsh to bypass the file modify tracking and permissions. I believe this is not a codex specific issue but a general issue for AI because the AI model will keep focus on "solve the goal" during their internal generation.
> 
> Personally, I haven't experienced AI modifying files through bash. Which models are you using?
> 
> Tools prompting is already injected into the system prompt, which has the highest priority/weight, LLM usually wouldn't use bash to edit unless it's necessary. Here I quote Claude Code's system prompt:
> 
> ```
> Executes a given bash command and returns its output.
> 
> The working directory persists between commands, but shell state does not. The shell environment is initialized from the user's profile (bash or zsh).
> 
> IMPORTANT: Avoid using this tool to run `find`, `grep`, `cat`, `head`, `tail`, `sed`, `awk`, or `echo` commands, unless explicitly instructed or after you have verified that a dedicated tool cannot accomplish your task. Instead, use the appropriate dedicated tool as this will provide a much better experience for the user:
> 
>  - File search: Use Glob (NOT find or ls)
>  - Content search: Use Grep (NOT grep or rg)
>  - Read files: Use Read (NOT cat/head/tail)
>  - Edit files: Use Edit (NOT sed/awk)
>  - Write files: Use Write (NOT echo >/cat <<EOF)
>  - Communication: Output text directly (NOT echo/printf)
> While the Bash tool can do similar things, it’s better to use the built-in tools as they provide a better user experience and make it easier to review tool calls and give permission.
> ```

I am not sure the root issue. But I am sure I found the issue before on gpt 5.4 xhigh/opus4.6 max. And I found opus breaches the permission recently(in recent one week). Maybe this is a new prompt by claude? 

If you believe the additional safeguard instruction is not necessary, I can remove them. But just based on my personal experience I still recommond adding them. 

Thanks to the opus, my claude permission setting file now already have at least 300 lines due to it continue trying to access all tools via bash.

ElwinLiu:
> > > Hello,
> > > Thanks for feedback. I can improve the workflow order instructions.
> > > But for the tool call instructions: I use all of codex, claude, and copilot. I found such issue that the agent tries to use bash or pwsh to bypass the file modify tracking and permissions. I believe this is not a codex specific issue but a general issue for AI because the AI model will keep focus on "solve the goal" during their internal generation.
> > 
> > 
> > Personally, I haven't experienced AI modifying files through bash. Which models are you using?
> > Tools prompting is already injected into the system prompt, which has the highest priority/weight, LLM usually wouldn't use bash to edit unless it's necessary. Here I quote Claude Code's system prompt:
> > ```
> > Executes a given bash command and returns its output.
> > 
> > The working directory persists between commands, but shell state does not. The shell environment is initialized from the user's profile (bash or zsh).
> > 
> > IMPORTANT: Avoid using this tool to run `find`, `grep`, `cat`, `head`, `tail`, `sed`, `awk`, or `echo` commands, unless explicitly instructed or after you have verified that a dedicated tool cannot accomplish your task. Instead, use the appropriate dedicated tool as this will provide a much better experience for the user:
> > 
> >  - File search: Use Glob (NOT find or ls)
> >  - Content search: Use Grep (NOT grep or rg)
> >  - Read files: Use Read (NOT cat/head/tail)
> >  - Edit files: Use Edit (NOT sed/awk)
> >  - Write files: Use Write (NOT echo >/cat <<EOF)
> >  - Communication: Output text directly (NOT echo/printf)
> > While the Bash tool can do similar things, it’s better to use the built-in tools as they provide a better user experience and make it easier to review tool calls and give permission.
> > ```
> 
> I am not sure the root issue. But I am sure I found the issue before on gpt 5.4 xhigh/opus4.6 max. And I found opus breaches the permission recently(in recent one week). Maybe this is a new prompt by claude?
> 
> If you believe the additional safeguard instruction is not necessary, I can remove them. But just based on my personal experience I still recommond adding them.
> 
> Thanks to the opus, my claude permission setting file now already have at least 300 lines due to it continue trying to access all tools via bash.

I suggest we remove it, because they're considered external bugs, which should be decoupled from Umpleonline repo. By the way, you might have already known this: make sure the context window is always under 30%(around) to prevent hallucination, and debug the tool call with the agents. 

Usually, the system prompt's constraint on tool-use is the most powerful, as the prompt is amplified through RL post training (the model was trained specifically to use them), and it's injected before every turn (as the model remember recent instructions better).

skyfackr:
@ElwinLiu done. Can you please review if change is ok? thanks

ElwinLiu (approved):
LGTM

Issue #2144, opened 2024-05-01
Design-decision assistance in Umple


TimLethbridge:
Core to engineering design is the creation of design alternatives, which can be compared using various criteria, and then a final design selected (or perhaps more than one for different versions).

Umple already has mixsets that allow description and generation of alternative designs.

It would be good to research how this capability could be leveraged so that alternative designs (alternative configurations of mixsets) could be separately tested or benchmarked, and have separate documentation ascribed to them.

One capability that might help is requirements embedded in Umple (see issue #1865 and its related issues). The intention is that requirements can be connected implemented by mixsets. So perhaps one way of using the Umple requirements feature would be to tag different mixsets corresponding to alternative designs according to the same set of criteria, and then being able to generate a table showing the comparisons.

In UmpleOnline it would be good to be able to switch on different designs (activating mixsets) by clicking buttons that would instantly switch between then.

TimLethbridge:
One capability would be to have a panel that would appear in UmpleOnline with a list of mixsets that could be activated; the user could click on the ones they want to activate. The set of mixsets could be found by scanning the code.

The second capability perhaps would be to be able to specify quality requirements (non functional) that could be tagged as  'quality' in req statements. Then in mixsets one could somehow tag a mixset with the same quality and give it a value

Suggestion:

```
req Speed quality { High {}; Medium {}; Low {} }
req Security quality { Perfect {}; SlightRisky {}; Weak {}; }

implementsReq Speed(High);
implementsReq Security(SlightlyRisky);
mixset Mixset1 {

  class X {}
}

implementsReq Speed(Medium);
implementsReq Security(Perfect);
mixset Mixset1 {
  class X {a;}
}

```

Output generated in requirements document:

|  | Mixset1 | Mixset2 |
| :--- | :---: | ---: |
| Speed | High | Medium |
| Security | Slightly risky | Perfect |

To do the above, follow similar approach to how #2377 (user stories, severak PRs including #2411) and #2378 (use cases) are being implemented

Step 1: Change grammar to enable
Step 2: Change metamodel to enable
Step 3: Change parsing to recognize and populate (see how done in #2411 )
Step 4: Change req doc to output table

PR #2468, opened 2026-04-20
Impl #2144: Feature model selector

skyfackr:
This PR implements the new mixset selector and replaces the old mix checkbox selector.

This also includes new generator `FeatureModelJson` to provide the feature model structure as json so the website can render the selector.

This also fixes a bug that may cause the requirement parser falls in the infinte dead loop.

As per previous meeting discussion, the umpleonline E2E tests are down. I tried but failed to deploy the tests on my local env. So the E2E test cases are deferred and not included. Java side unit tests are included


skyfackr:
Copilot found a bug that requires deep investigation and maybe change on AJAX component. mark as draft before done

PR #2493, opened 2026-07-05
Fixes #2466 sync Live View selector on example load

### LINKED ISSUE #2466: "Live View" selection not changed when switching the example files
## Bug Summary

When switching to a different example file, the backend updates the **Live View** state correctly, but the UI still shows the previous selection.

## Steps to Reproduce

1. Open the UmpleOnline website.
2. In **Examples**, select **Feature Diagram -> Feature Dependencies...**
3. After the example loads, check the **Live View** dropdown in the top-left corner.

## Expected Behavior

The **Live View** should display **Feature Diagram**, since the example file changes the diagram type and the backend state has already been updated accordingly.

## Actual Behavior

The **Live View** still shows the previous selection, even though the backend state has already changed.

## Impact

This causes the UI to become out of sync with the actual **Live View** state in the background.

## Possible Cause

The UI is not properly reflecting the updated **Live View** state after the example file is switched.

faizan9233:
Bug -
"Live View" dropdown selection doesn't update when switching to a different example that changes the diagram type. The backend correctly switches diagram type, but the UI dropdown still shows the previous selection.

Steps to Reproduce -

Open the UmpleOnline website.
In Examples, select Feature Diagram → Feature Dependencies...
After the example loads, check the Live View dropdown in the top-left corner.

Root Cause -
Action.loadExample only called Action.setLiveView on the default GvClass branch. The other seven diagram-type branches (state, feature, ERD, instance, state tables, event sequence, structure) set the backend diagramType correctly but never synced the #liveViewSelector dropdown, so it kept showing the previous selection after an example loaded a different diagram type.

Fix -
Added the matching Action.setLiveView(...) call to each branch, using the option codes already defined in umple.php's #liveViewSelector.

Testing -
Verified manually in-browser: loaded examples across Feature/State/ERD/Instance/State Tables/Event Sequence/Structure categories and confirmed the Live View dropdown now updates correctly in each case. The existing default GvClass path was unaffected.

### DIFF
--- a/umpleonline/scripts/umple_action.js
+++ b/umpleonline/scripts/umple_action.js
@@ -4667,29 +4667,36 @@ Action.loadExample = function loadExample()
   var diagramType="";
   if(Page.useGvStateDiagram) {
     diagramType="&diagramtype=state";
+    Action.setLiveView("sd");
   }
  else if(Page.useGvFeatureDiagram) {
     diagramType="&diagramtype=GvFeature";
+    Action.setLiveView("gfd");
   }
 
   else if(Page.useGvEntityRelationshipDiagram) {
     diagramType="&diagramtype=entityRelationshipDiagram";
+    Action.setLiveView("erd");
   }
 
   else if(Page.useInstanceDiagram) {
     diagramType="&diagramtype=instanceDiagram";
+    Action.setLiveView("instanceDiagram");
   }
 
 
   else if(Page.useStateTables) {
     diagramType="&diagramtype=StateTables";
+    Action.setLiveView("stateTables");
   }
 
   else if(Page.useEventSequence) {
     diagramType="&diagramtype=eventSequence";
+    Action.setLiveView("eventSequence");
   }
   else if(Page.useStructureDiagram) {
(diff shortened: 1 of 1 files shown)


TimLethbridge:
@faizan9233 Thanks for this. I am slightly challenged, since our current CI pipeline requires PRs to come from a clone rather than a fork ... The Jenkins server is not being triggered by a PR from a fork. Since this is a change simply to a file that is not affected by CI, I might simply bypass CI temporarily to let this PR be merged.

faizan9233:
> @faizan9233 Thanks for this. I am slightly challenged, since our current CI pipeline requires PRs to come from a clone rather than a fork ... The Jenkins server is not being triggered by a PR from a fork. Since this is a change simply to a file that is not affected by CI, I might simply bypass CI temporarily to let this PR be merged.

Thanks for the context - sounds good, happy to have you bypass CI for this one. Let me know if there's anything else needed on my end.

TimLethbridge:
I will bypass Jenkins and merge after 2496 ... please don't try updating base branch just yet; I will do that after 2496 passes

PR #2496, opened 2026-07-20
Prevent arbitrary file creation through filename validation

DeveshKhanna7:
This change prevents arbitrary file creation by validating user-supplied filenames before saving uploaded files.

Only supported file extensions ('.ump' and '.svg') are accepted. Requests with unsupported extensions are rejected with an HTTP 400 response.

Tested:
- Successfully saved a valid `.svg` file.
- Attempted to save `shell.php` and received `HTTP 400 Bad Request`.


TimLethbridge:
I have tested this and will merge in the morning of Aug 2

PR #2505, opened 2026-09-02
Fixes #2454 warn when use statements do not satisfy a require statement

### LINKED ISSUE #2454: Feature model constraint not enforced when conflicting use statements found
When two or more `use` statements activate mixsets that violate feature model constraints (for example, `require [not X]`), no error or warning is emitted.

The `satisfyFeatureModel()` method correctly evaluates the constraints and returns `false`, but its return value is silently discarded in `analyzeFeatureModel()`.

---

## Steps to Reproduce

```umple
mixset DesignA {
  require [not DesignB];
  isFeature;
  class X { name; }
}

mixset DesignB {
  require [not DesignA];
  isFeature;
  class X { age; }
}

use DesignA, DesignB;
```

---

## Expected Result

A warning or error should be emitted indicating that the active `use` statements violate the feature model constraints.

Specifically, `DesignA` and `DesignB` are mutually exclusive, but both are activated.

---

## Actual Result

No warning or error is produced.

In `UmpleInternalParser_CodeRequireStatement.ump` (lines 397–403), `analyzeFeatureModel()` calls `satisfyFeatureModel()` but discards its return value:

```java
public void analyzeFeatureModel() {
    parseMixsetsNotUsed();
    if(model.getFeatureModel() != null)
        model.getFeatureModel().satisfyFeatureModel();
}
```

The only consumer of this return value appears to be the `GvFeatureDiagram` generator, which displays `"Valid"` / `"Not Valid"` visually.

As a result, users who do not generate a feature diagram receive no feedback at all.

---

## Suggested Fix

When `satisfyFeatureModel()` returns `false`, emit a new warning with a message such as:

> Active use statements violate feature model constraints.

Ideally, the warning should identify which specific constraint(s) failed, but even a general warning would be a significant improvement over the current silent acceptance.

jatingartan:
The verdict of `satisfyFeatureModel()` was computed during analysis and then discarded, so `use` statements that violate a `require` statement compiled silently; only the feature diagram showed "Not Valid".

**Changes**

- New warning 1514, raised once per unsatisfied require statement, positioned at that statement and naming the mixset or file that contains it.
- Each `FeatureLink` records the position of the require statement that created it.
- `FeatureModel.getUnsatisfiedFeatureLinks()` checks every applying require statement on its own; `satisfyFeatureModel()` is now `isEmpty()` on that list, so the diagram status and the warning always agree. A require statement at the top level of a file always applies; one inside a mixset applies only when that mixset is used, as the manual describes. This replaces the recursion into used features (and its cycle guard), since those requirements are now checked as constraints of their own.
- Two diagram-status changes follow from that rule: a require statement in an unused mixset no longer makes the configuration invalid, and a used branch of an `xor` whose own require statement fails is now reported (the recursion masked it).
- Manual: page `W1514RequireStatementNotSatisfied` with an example, and the rule stated on the require statement page.
- `BerkeleyDB_SPL.ump` and `MixedFeatureWithOrphanRoots.ump` now carry a valid configuration so they stay warning-free.
- The uncompiled duplicate `cruise.umple/src/feature/UmpleInternalParser_CodeRequireStatement.ump` is removed in a separate commit.

**Tests**

- New tests: warnings in used mixsets and at file level, no warning for unused mixsets, one warning for a statement that creates several links, and exact attribution in the diamond and xor fixtures.
- Tests that deliberately use invalid configurations parse with `parseWarnings` instead of `run()`, which throws on any warning, and assert that 1514 is the only message. The two syntax fixtures now declare and use the mixsets they require.
- Full build passes locally.

**Follow-ups noticed while working on this**

- `require [not M2 and m4]` is parsed as `[M2 and m4]`: a leading `not` followed by a linking operator is dropped. The syntax fixture states the two constraints separately for now.
- A file that is `use`d only from an unused mixset is partially parsed to discover dependencies, so its top-level require statements are treated as applying and can warn.


Ilyas-Ait-Ali (approved):

Ilyas-Ait-Ali (approved):
I built and ran it locally, all green

One thing: you flag `not M2 and m4` as a follow-up, but it bites harder now. 
- use `m4` gets a false warning
- use `M2, m4` passes silently

where before it only skewed a diagram label. Worth an issue plus a test pinning current behaviour.

TimLethbridge:
I am going to merge this now. I raised issue #2521 regarding the case sensitivity issue mentioned in the review

PR #2511, opened 2026-09-13
Fix inconsistent filename format in brace diagnostics

### LINKED ISSUE #2401: Inconsistent format of the field "filename" in messages from umplesync.jar
## Description

umplesync.jar can give full  file path or only basename depending on the error code.

## Repro

Image

Image

## Solution

Need to unify the output format of the field "filename" in umplesync.jar. This can affect the behavior of diagnostics feature of the LSP

Two places to look
Parser_Code.ump:1079
UmpleInternalParser_CodeClass.ump

When testing check UmpleOnline to ensure it works OK,. particularly in the scenario in #2301 as fixed in PR #2360  where UmpleOnline has been changed to display the filename

Fadiedy:
## Summary

Fixes #2401.

This PR fixes inconsistent values in the `filename` field returned by
`umplesync.jar` for parser diagnostics.

Previously, a normal parsing error returned:

`"filename": "old.ump"`

while a brace mismatch error (1505) could return an absolute path such as:

`"filename": "/private/tmp/umple2401-check/old.ump"`

The fix normalizes the filename used by the brace mismatch diagnostic to the
basename, keeping the output consistent with other parser diagnostics.

## Changes

- Normalize the filename when creating the `Position` for brace mismatch errors.
- Add regression test `braceMismatch_filenameUsesBasename`.

## Testing

- Error 1503 returns `"filename": "old.ump"`
- Error 1505 returns `"filename": "old.ump"`
- `braceMismatch_filenameUsesBasename` passed successfully
- `UmpleParserTest` completed successfully with no failures or errors
- Full build completed successfully

## Related

- Fixes #2401
- Related to #2301 and PR #2360


TimLethbridge (changes requested):
You have made a massive number of changes to UmpleInternalParser_CodeClass.ump that consist of just deletion of extraneous spaces. Please could you replace with just the actual functional lines that have changed.

It is acceptable to 'clean up' files as a separate refactoring task to remove technical debt. But a PR to fix an issue should focus on just the lines changed relevant to that issue only.

Fadiedy:
Thank you for pointing this out. The unrelated whitespace changes have now been removed. I restored the file and kept only the functional changes required for issue #2401, along with the regression test. The PR has been updated and is ready for another review.

PR #2513, opened 2026-09-14
Fix collaboration edits not reaching other browsers

jatingartan:
Collaboration on new.umple.org: the collaborator count worked, but edits never showed up in the other browsers.

umple.php called Collab.connectCollabServer() before the model text had loaded, so the collab server got an empty document and rejected every later change. The connection is now made after the model file and the tab list have loaded. A browser joining an existing session takes the server's document, even if it is empty. The server only seeds a session that has no edits yet, and drops a push it can't apply instead of forwarding it.

Tested on new.umple.org with 2 and 3 browsers: typing both ways, joining late, reload, concurrent edits, paste, empty model, two tabs, disconnect and reconnect, fork.


jatingartan:
Continued in #2514 after renaming the branch.

PR #2518, opened 2026-09-16
Issue #2378: Extend useCase syntax with include, isA and step conditions

Ilyas-Ait-Ali:
This PR implements the extended syntax step of Issue #2378, covering the work deferred in #2442.

## Scope

Add the optional use case extensions to the structured `useCase` body.

This PR adds support for:

* use case relationships:
  * `include`: incorporate the steps of other use cases
  * `isA`: specialize another use case
* optional step conditions on:
  * `userStep`
  * `systemResponse`

Conditions are written like a state machine guard, and `isA` follows the existing comma-separated `isA Name;` convention:

```
req UC5 useCase {
  isA UC1;
  include UC2, UC3;
  systemResponse 1 [card accepted] { confirm the order }
}
```

## Changes

* Added the `useCaseRelationship` and `useCaseStepCondition` grammar rules

* Extended the `Requirement` metamodel with `includedUseCases` and `parentUseCases`, and `UseCaseStep` with an optional `condition`

* Updated the deep-copy constructor to preserve relationships and step conditions

* Updated requirement parsing in both class and trait parser implementations, moving step and relationship scanning into shared helpers alongside `parseReqRefTokens`

* Extended output rendering with `Specializes`, `Includes` and step conditions, in both the Plain Requirements Document generator and `Requirement.getBodyDisplayText`

* Preserved backward compatibility with plain, `userStory` and existing `useCase` requirements; rendered output for the existing tests is unchanged byte for byte

* Added a syntax note to the user manual in `build/reference/`

## Tests

Added coverage for:

* `include` with single and multiple references
* `isA` specialization
* conditional and unconditional steps in the same use case
* all extensions combined with `who/when/what/why`
* the lowercase `usecase` alias with extensions
* plain requirements keeping `include` and `isA` as free text
* metamodel storage and deep copy of relationships and step conditions
* Plain Requirements Document rendering of relationships and step conditions

## Notes

* `include` and `isA` references are stored but not validated; an unknown target is not reported. This is deliberate and left for a follow-up
* The full user manual page with worked examples is the remaining step of Issue #2378


Samoumine (commented):
Tried isA, include, and steps with conditions. Output looks right. Plain requirements and older use cases still render the same as before.
Small thing, not blocking: isA UC1, UC2; works but no test covers more than one parent.
Other than that, seems all good.

Ilyas-Ait-Ali:
Good catch. Added a test for it: `457_ReqUseCaseIsAMultipleParents.ump` with `isA UC1, UC2;`, asserting both parents are stored.

Samoumine (approved):

TimLethbridge:
Looks good. Will merge once CI finishes.

TimLethbridge (approved):
Looks good

Issue #2439, opened 2026-04-04
Class not fully clickable while in G mode while using light mode


kyle-labelle:
When in light mode, the SVG for the class has no background fill. As such it is only clickable on the text or on the border. This affects state machines as well.

PR #2519, opened 2026-09-16
2439 clickable class in light mode

kyle-labelle:
This PR fixes the bug for umpleonline in G mode where when using light mode theme, only the border of the box and the text was clickable when using the tools from the draw menu (such as association, generalization, delete). This PR allows you to click anywhere within the class box.


TimLethbridge (changes requested):
Can you please make the same fix for state machines (as mentioned in #2439).

Issue #2262, opened 2025-10-28
Add Drawing capability of Finite State Machines to UmpleOnline


eabase:
EDIT: For those tackling this issue, please read my comments further down for the specific details of this. See also #2224

I'm looking here:
* https://cruise.umple.org/umpleonline/

And the drawing tools for Class Diagrams, is really cool. 
But I need a drawing tool for the FSM as well. There is none.

Please consider also being able to draw a `Finite State Machine` to generate the code.  

<img width="1244" height="836" alt="Image" src="https://github.com/user-attachments/assets/662bc2f2-2331-4825-8770-5b5ca962164d" />

TimLethbridge:
For state machines, @eabase Click on the 'S' , Start by specifying a class, but then you can edit state machiines

We will make improvements to add more capabilities. We have some PRs we are preparing along this line

eabase:
@TimLethbridge 
Hi Tim, 

> Click on the 'S' , Start by specifying a class, but then you can edit state machiines

Sorry but I don't get it. Can you post some screen shots or video or something. 
I don't see how you can convert a class diagram to FSM item.

TimLethbridge:
@eabase There is a vast amount of documentation! 

Click on 'Manual' at the top and open the pages about State Machines. CLick on the examples (they will load in UmpleOnline).

Or in the middle of UmpleOnline where it says 'Class Diagrams', select that (it is a menu) and choose "State Machines'. Then pick an example from the menu below that.

We have had several million people figure this out, so I am sure you can!

But we do value hearing from someone who hasn't figured it out, as user experience is key. I assume you are using a laptop so you should be able to hover your cursor over the various items on the screen. If you are using a tablet or phone I do see how it would be harder.

You can't 'convert a class diagram to an FSM'. But you can create an FSM in a class. Follow the examples.

By the way, the 'S' is right there at the top left of UmpleOnline. There are three buttons E G S (E and G are different class diagram formats where you have been using E) mode.

eabase:
@TimLethbridge 

> Then pick an example from the menu below that.

I don't want to see an example. I want to draw my own and get the code generated. That doesn't work, because there are no items in the "draw" window to add/use.

TimLethbridge:
Umple should indeed have icons that appear in State Machine mode to

1) Add a new state (with click/click or  drag and drop ... if dropped within a state, then the new state would be a substate)
2) Make a transition (click from the origin state, and click on the destination state)
?? Others. 

[...]

PR #2522, opened 2026-09-22
[#2262] added S mode drawing

kyle-labelle:
Added ability to add states and transitions through the draw menu while in S mode. Works in the same way that adding classes and associations works for G mode class diagrams.


TimLethbridge:
Excellent! I tested this and it looks good. I will merge. We will have to also check if this sort of functionality exists in the experimental new UmpleOnline

Issue #2523, opened 2026-09-23
Draw use case diagrams from useCase requirements


Ilyas-Ait-Ali:
#2378 added the `useCase` requirement type. The metamodel now holds everything a UML use case diagram needs, so a generator could draw one. #2378 mentions this but its agreed breakdown did not cover it.

## What is already there

After #2378, `Requirement` carries:

- `who`  the actor
- `parentUseCases`  from `isA`, i.e. UML generalization
- `includedUseCases`  from `include`, i.e. the UML include relationship

No new syntax is needed, only a generator over the existing model.

## Proposal

A `GvUseCaseDiagram` generator, following `GvFeatureDiagram` as the precedent for rendering a non-class model through Graphviz:

| Model | Diagram |
|---|---|
| `req X useCase` | use case ellipse |
| `who { actor }` | actor, with an association to the use case |
| `include A, B;` | dashed arrow to A and B, labelled `«include»` |
| `isA P;` | generalization arrow to P |

Steps and conditions are not part of a UML use case diagram; they belong to the use case description, which PlainRequirementsDoc already renders. Drawing them would be a sequence or activity diagram, which is separate work.

TimLethbridge:
Yes, go ahead with this. It will need test cases, a user manual page with examples, and also menu items in old umpleonline and new umpleonline

PR #2524, opened 2026-09-23
Fixes #2378 document the useCase requirement type in the user manual

Ilyas-Ait-Ali:
## Changes

- Four examples in `manualexamples/`: plain, structured, partial structured, and one showing `isA`, `include` and conditional steps.
- `0215requirementsExamples.txt`: the four examples plus an updated intro.
- `0215requirementStatement.txt`: replaces the run-on `useCase` bullet from #2518 with a `useCase requirements` section.

## Verification

Full `bumple` passes. All four examples compile with no warnings.

## Notes

Use case diagrams, mentioned in #2378 but not in the agreed breakdown, are tracked separately in #2523 

Do not merge before #2518. This branch is based on it, so merging this first would pull the step 5 changes into master without that PR being approved.


TimLethbridge (approved):
Looks Excellent. I will merge.

Issue #2385, opened 2026-03-13
Update Umple diagram/output types so they are more consistent. Particularly state tables supporting state operations, ERDs supporting the same as class diagrams


TimLethbridge:
Some improvemnts to diagram types (especially as expressed on the right hand side):

Check these off when done

 - [ ] Enable the state table view to have the same edit menus attached as in state diagrams
 - [ ] Enable ERDs to select attributes and in general to select properly when items are clicked on
 - [ ] Ensure instance diagrams respect dark mode, and if one clicks on an instance its class should be selected in the text.
 - [ ] Ensure event sequence diagrams enabl their various elements to select the relevant code in state diagrams

PR #2525, opened 2026-09-23
[ #2385 ]: support source selection from ERD elements

Aliloraf:
## Summary

This PR fixes the ERD selection part of #2385.

Clicking ERD elements now selects the related Umple code more consistently.

## Changes

- Clicking an attribute selects the attribute declaration.
- Clicking an IsA node selects the child class.
- Clicking a relationship diamond selects the association.
- Clicking either relationship edge selects the association.
- ERD links now use the actual class/association information instead of Graphviz-generated node names.

## Testing

- Full build passes.
- 5,978 tests pass.
- ERD generator tests pass.
- Class diagram generator tests pass.
- Tested manually in UmpleOnline.

Part of #2385.


TimLethbridge:
This contributes to solution of #2385 ... state tables still need to be done

Issue #2504, opened 2026-08-18
genJson doesn't work with state machines


diamondq:
If you add a statemachine to a class, and activate the genJson option, the statement variables are not persisted (or loaded) from the JSON.

It appears to only store and load attributes.

```
suboption "genJson";
class Bank {
  String bankName;
  state {
     Initial {
        start -> Finish;
     }
     Finish {}
  }
}
```
generates:
```
public boolean toJsonHelper(StringBuilder toJsonOutput, HashSet<Object> visitedList, int nestLevel, boolean atConcreteClass){

          ...

          toJsonOutput.append(indent);
          toJsonOutput.append("  \"");
          toJsonOutput.append("bankName");
          toJsonOutput.append("\" : \"");
          String primValue_0=""+getBankName()+"";
          toJsonOutput.append(primValue_0.replace("\\","\\\\").replace("\"","\\\""));
          toJsonOutput.append("\"");
          haveOutputItem=true;
      //haveOutputItem = false;
      //haveOutputItem = false;
              
          toJsonOutput.append(indent+"  \n");
  
          ...
  }
```

PR #2529, opened 2026-09-23
Fixes #2504 genJson saves and restores state machine states

jatingartan:
With the genJson suboption, toJson now writes the current state of every state machine, and fromJson restores it. Before this change the state was never written, and a reloaded object came back with a null state, so its first event threw a NullPointerException.

- One pair per state machine enum field, nested regions included, placed after the attributes: `"status" : "On", "statusOn" : "Playing"`.
- fromJson assigns the fields directly, so no entry actions, do activities or timers run. A state missing from the json (for example json saved before this change) gets the value the constructor would give it. History fields get the constructor's default too.
- An unknown state name throws IllegalArgumentException.
- Java only, since genJson exists only there. The manual page gains a section on restore semantics and a new example.

Tests: a template test (`ToJsonTest_3`) and a testbed round trip (`JsonPlayerTest`). The testbed fixture lives in `testbed/src/runtime` because that directory is compiled one file at a time and genJson applies to the whole model. The fixture has no namespace and its test is in the default package, because fromJson rejects qualified class names, which is a separate pre-existing bug. The fixture follows the neighbouring `ToJsonTest_N` naming instead of an issue-number prefix.

Verified: bumple green (6770 tests, 0 failures, including all testbeds), the issue's sample round-trips on the command line, and the new manual example compiles and runs.


TimLethbridge:
Python tests are failing only in ant (Appveyor). This implies two things: 1) That the Ant build is doing some things that the Gradle build (Jenkins) is not doing ... hence the Gradle build needs fixing. 2) Changes made to code in this PR are not being properly translated to Python (which is not surprising since the Java-to-Python transpiler is not perfect).

